PR #2, opened 2016-11-15
WIP: Function interface specs

mauro3:
Concering comment https://github.com/JuliaDiffEq/Roadmap/issues/5#issuecomment-260681959:

I only saw the `jac_exists` function after I put this together.  However, neither `has_jac` nor `jac_exits` passes the test in above comment.

However, when using a trait-function it would be fully inferred.  Note however, that the traits would struggle with adding, say a jacobian, to the function.  It may would still dispatch to the wrong one.

### DIFF
--- a/src/problems/functions.jl
+++ b/src/problems/functions.jl
@@ -1,0 +1,96 @@
+# Functions for AbstractODEProblem and possibly other DEProblem types
+#
+# Implements passing in Jacobians (and possibly other functions) via
+# function overloading:
+#
+# - f(...) - objective function
+# - f(Val{:jac}, ...) - Jacobian of objective function
+# - the details of what `...` needs to be depends on the
+#   AbstractODEProblem subtype
+
+
+# Method_exists does not work:
+#
+# julia> f(::Val{:jac}, a, b, c) = 5
+# f (generic function with 1 method)
+#
+# julia> method_exists(f, Tuple{Val{:jac}, Vararg})
+# false
+#
+# Thus hand-code it:
+check_first_arg(f,T::Type) = check_first_arg(typeof(f),T)
+function check_first_arg{F}(::Type{F}, T::Type)
+    typ = Tuple{Any, T, Vararg}
+    for m in Base.MethodList(F.name.mt) # F.name.mt gets the method-table
+        m.sig<:typ && return true
+    end
+    return false
+end
+
+has_jac(f) = check_first_arg(f, Val{:jac})
+
+# Using SimpleTraits to dispatch on existence of jac-method
+using SimpleTraits
+@traitdef HasJac{F}
+@generated SimpleTraits.trait{F}(::Type{HasJac{F}}) = has_jac(F) ? :(HasJac{F}) : :(Not{HasJac{F}})
+# now a trait methods can dispatch on this:
+# @traitfn fn(g::::HasJac, ...) = ...
+# @traitfn fn(g::::(!HasJac), ...) = ...
+
+
+######################
+## Tests (TODO move to test/...)
+#####################
+f123(x,y) = 1
+f123(::Val{:jac}, x) = 2
+
+g123(x,y) = 1
+g123{T}(::Val{:jac}, x::T) = 2
+
+h123(x,y) = 1
+h123{T}(x::T) = 2
+
+immutable T123 end
+(::T123)(a) = 1
+(::T123)(::Val{:jac}, a) = 1
+
+immutable G123 end
+(::G123)(a) = 1
+(::G123)(b, a) = 1
+
+@assert has_jac(f123)
+@assert has_jac(g123)
+@assert !has_jac(h123)
+@assert has_jac(T123())
+@assert !has_jac(G123())
+# Arguably below could be ==false as T123 is a constructor
+# function. However, I'm not sure how to implement this.
+@assert has_jac(T123)
+@assert !has_jac(G123)
+
+@traitfn tfn(f::::HasJac) = true
+@traitfn tfn(f::::(!HasJac)) = false
+
+@assert tfn(f123)
+@assert tfn(g123)
+@assert !tfn(h123)
+@assert tfn(T123())
+@assert !tfn(G123())
+@assert !tfn(T123)  # NOTE this is inconsistent as has_jac(T123)==true
+@assert !tfn(G123)
+
+# ParameterizedFunction:
+type  LotkaVolterra <: ParameterizedFunction
+         a::Float64
+         b::Float64
+end
+(p::LotkaVolterra)(t,u,du) = begin
+         du[1] = p.a * u[1] - p.b * u[1]*u[2]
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
>Note however, that the traits would struggle with adding, say a jacobian, to the function. It may would still dispatch to the wrong one.

I think this is a fundamental issue, not an implementation issue. If having a trait can be inferred at compile time, then it must be constant (or at least, the `has_jac(f)` on a function has to be constant and essentially declared when `f` is declared). If `has_jac(f)` can notice that you later added a Jacobian function, then it has to be finding that out at runtime and thus not have the information at compile time. From this understanding (if it's correct), I don't think there's an easy way around that.

What we could do is have a different `has_jac(f)` dispatch for `ParameterizedFunction` which uses a trait and `has_jac(f)` for `Function` which uses the methods table. Then the easy way of defining a Jacobian will work, even if it's not necessarily good, and the more advanced way of using call-overloading with subtyping could then be documented after that (similar to how out-of-place functions are allowed, though in-place functions are preferred). With this, solvers can be written without care as to whether users took the easy or the hard way?

mauro3:
Actually, the way around it, is to call `@traitimpl HasJac{F}` after a jacobian is added.  Then (barring issues with Julia issue 265) it should work.

(Note the constant function used for trait-dispatch is `trait` [here](https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/2/files#diff-e91a1e70bb596306e7fc9ec07f2b7243R35).  Thus the use of a generated function, otherwise it would lead to dynamic dispatch in trait-functions.)

ChrisRackauckas:
>Actually, the way around it, is to call @traitimpl HasJac{F} after a jacobian is added. Then (barring issues with Julia issue 265) it should work.

That seems fine to me. It would just go in the same place of the documentation as defining the Jacobian (I want to make a full example for that first, and then below have the full interface. That way most users will likely define just the Jacobian which is fine: the others probably can't be done by hand).

The second probably needs a better title: https://juliadiffeq.github.io/DiffEqDocs.jl/latest/features/performance_overloads.html#Declaring-Explicit-Jacobians-1

src/problems/functions.jl
```diff
@@ -0,0 +1,96 @@
...
+check_first_arg(f,T::Type) = check_first_arg(typeof(f),T)
+function check_first_arg{F}(::Type{F}, T::Type)
+    typ = Tuple{Any, T, Vararg}
+    for m in Base.MethodList(F.name.mt) # F.name.mt gets the method-table
+        m.sig<:typ && return true
+    end
+    return false
+end
+
+has_jac(f) = check_first_arg(f, Val{:jac})
```
ChrisRackauckas: At least as stated before, the interface has a capital here:

https://juliadiffeq.github.io/DiffEqDocs.jl/latest/features/performance_overloads.html

>f(Val{:Jac},t,u,J) # Call the explicit Jacobian function

src/problems/functions.jl
```diff
@@ -0,0 +1,96 @@
...
+(::G123)(a) = 1
+(::G123)(b, a) = 1
+
+@assert has_jac(f123)
+@assert has_jac(g123)
+@assert !has_jac(h123)
+@assert has_jac(T123())
+@assert !has_jac(G123())
+# Arguably below could be ==false as T123 is a constructor
+# function. However, I'm not sure how to implement this.
```
ChrisRackauckas: That is weird but not a show-stopper

src/problems/functions.jl
mauro3: Actually, would you mind updating the readme with links to all relevant places in the docs? (I have no overview)

ChrisRackauckas: Sure thing.

ChrisRackauckas:
The tests should move to /test and I think the following test should be used to make sure the traits option is a compile-time check:

```julia
function testing(f)
  if has_jac(f)
    a = 2
  else
    a = 3.0
  end
  a
end
@inferred testing(f)
```

I think we should also just go forward with the SimpleTraits.jl dependency: it has a compelling enough use case. After this goes through I'll set ParameterizedFunctions.jl up with traits. 

Could you also expand the interface to have the other checks as well?

ChrisRackauckas:
The other way is that ParameterizedFunctions and the docs can switch to `:jac`. Is there a guide on how value types should be named? I know functions should be lower-case according to the guide, and types should be upper case, but I don't think they thought about a value-type for declaring a new function. It poses quite an interesting dilemma.

mauro3:
I think it's fine either way as long as we're consistent.  Actually, to be consistent with variable use, I'd think `:jac` is better (or `:J` but that is too short).

ChrisRackauckas:
I took it to JuliaPraxis, and after some discussion agree it should change to `:jac`.

ChrisRackauckas:
ParameterizedFunctions.jl was changed to have everything de-captialized (`:jac`). The docs were changed to match. I'll implement the traits on all of the ParameterizedFunctions in a bit.

ChrisRackauckas:
I looked into this. It seems that as I was saying before, any method which uses the methods list / table will not be able to compute this at compile time. We could go through the trouble of creating a workaround that is at runtime and etc. etc., but really, the user experience isn't bad at all:

```julia
f123(x,y) = 1
f123(::Val{:jac}, x) = 2
@traitimpl HasJac{typeof(f123)}
```

If we just have it documented to put that one line after the Jacobian declaration, then the three lines:

```julia
@traitdef HasJac{F}
@traitfn has_jac(x::::HasJac) = true
@traitfn has_jac(x::::(!HasJac)) = false
```

makes everything work at compile time. If someone is looking through the manual for how to declare Jacobians by hand (something which is already done automatically for many systems of ODEs), I think they can spare that one line of code. If we just accept that, then everything works perfectly. 

I am willing to take that option, and if you want to work on a macro to make it easier:

```julia
@add_jac f(t,u,J) = ... # Makes f(Val{:jac},t,u,J) and does @traitimpl
```

then I think that's a nice and easy solution for both developers and users.

ChrisRackauckas:
```julia
@traitfn testing(f::::HasJac)   =2
@traitfn testing(f::::!(HasJac))=3.0

@code_llvm testing(f123)
@code_llvm testing(g123)
@code_llvm testing(h123)
```

I am really dumb. After playing with it a bit, I see if you use the `@traitfn` correctly, everything is inferred. So this all works out. 

Since this all works, I'll pull this in and add the other traits. Looks like there's a merge conflict to handle though.

ChrisRackauckas:
I got it!

```julia
@traitdef HasJac{F}
__has_jac(f) = check_first_arg(f, Val{:jac})
@generated SimpleTraits.trait{F}(::Type{HasJac{F}}) = __has_jac(F) ? :(HasJac{F}) : :(Not{HasJac{F}})
has_jac{T}(f::T)   = istrait(HasJac{T})
```

Only check the methods table to generate the trait and you're good. Now it auto-finds the trait and `has_jac` is the right value. Thanks for the help!

PR #4, opened 2016-11-16
Exporting solve

mauro3:
Tests not passing, but that should be a unrelated malaise?  If that is the case, then please merge this.


ChrisRackauckas:
>Tests not passing, but that should be a unrelated malaise?

Yeah, the test is just saying that StochasticDiffEq isn't updated to the new interface. It is on master and needs a tag.

ChrisRackauckas:
Note that the test that fails there is just testing the solution interface, but uses OrdinaryDiffEq and StochasticDiffEq to get solutions to test it. That's why it's a problem.

PR #5, opened 2016-11-16
WIP: removing bits which should go into packages

mauro3:
This PR removes bits which I think should be in packages.  In particular the all the algorithms needen't and shouldn't be defined here.  If I understood correctly, this is my answer to https://github.com/JuliaDiffEq/Roadmap/issues/9.

Note that this of course also necessitates changes in downstream pkgs.


src/DiffEqBase.jl
```diff
@@ -16,84 +16,55 @@ module DiffEqBase
   abstract AbstractSDEProblem <: DEProblem
   abstract AbstractDAEProblem <: DEProblem
   abstract AbstractDDEProblem <: DEProblem
-  abstract AbstractPoissonProblem <: DEProblem
-  abstract AbstractHeatProblem <: DEProblem
-  "`PdeSolution`: Wrapper for the objects obtained from a solver"
+  abstract AbstractPDEProblem <: DEProblem
+  "`DSolution`: type to hold the objects obtained from a solver"
```
mauro3: In this package we should only define the abstract types of broad categories of equations, I think.  A heat-problem seems to specific.

src/DiffEqBase.jl
```diff
@@ -16,84 +16,55 @@ module DiffEqBase
...
   abstract DESolution
   abstract AbstractODESolution <: DESolution
   abstract AbstractSDESolution <: AbstractODESolution # Needed for plot recipes
   abstract AbstractDDESolution <: AbstractODESolution # Needed for plot recipes
   abstract AbstractFEMSolution <: DESolution
   abstract AbstractSensitivitySolution
-  "`Mesh`: An abstract type which holds a (node,elem) pair and other information for a mesh"
-  abstract Mesh
```
mauro3: A mesh seems like an implementation detail which should go elsewhere.

src/DiffEqBase.jl
```diff
@@ -16,84 +16,55 @@ module DiffEqBase
...
-  "`Tableau`: Holds the information for a Runge-Kutta Tableau"
+
+"`Tableau`: Holds the information for a Runge-Kutta Tableau"
   abstract Tableau
 
   "`ODERKTableau`: A Runge-Kutta Tableau for an ODE integrator"
   abstract ODERKTableau <: Tableau
 
-  "`DEIntegrator`: A DifferentialEquations Integrator type, used to initiate a solver."
-  abstract DEIntegrator
```
mauro3: Don't know what this does.

src/DiffEqBase.jl
```diff
@@ -16,84 +16,55 @@ module DiffEqBase
...
-
-  abstract ParameterizedFunction <: Function
-  abstract SensitivityFunction <: ParameterizedFunction
```
mauro3: If we follow PR #2 then this is not needed.

src/DiffEqBase.jl
```diff
@@ -16,84 +16,55 @@ module DiffEqBase
...
-    if default_set == true && !isempty(args)
-      error("The chosen algorithm, "*string(args[1])*", does not exist.
-        Please verify that the appropriate solver package has been installed.")
-    elseif default_set == true
-      error("No algorithm is chosen but default_set=true.")
-    end
-    alg,extra_kwargs = default_algorithm(prob;kwargs...)
-    solve(prob,alg,args...;default_set=true,kwargs...,extra_kwargs...)
-  end
+  # function solve(prob::DEProblem,args...;default_set=false,kwargs...)
```
mauro3: I'm not sure how to do a default solver best.  I think it has to be done package-wise: the first package providing Algs which gets loaded can set a default solver.  But not sure how to implement it.

src/algorithms/ode_algorithms.jl
```diff
@@ -1,158 +0,0 @@
-abstract AbstractODEAlgorithm
```
mauro3: Having these here definitely breaks encapsulation.  And there shouldn't be any problem defining them in their respective packages.

src/utils.jl
```diff
@@ -1,11 +1,3 @@
-macro def(name, definition)
```
mauro3: Not sure what this does.

src/DiffEqBase.jl
ChrisRackauckas: Makes sense.

src/DiffEqBase.jl
ChrisRackauckas: Then there should be a FEMMeshes.jl for holding meshing functionality common to PDEs.

src/DiffEqBase.jl
ChrisRackauckas: Nothing anymore. Good find.

src/DiffEqBase.jl
mauro3: Ok, let's leave it here.  (At some later stage, a PDE related interface package might make sense.)

src/DiffEqBase.jl
ChrisRackauckas: It's definitely not package-wise because the default solver for different cases is many times in different packages. For example, if you're trying to solve a stiff equation on Float64's, Sundials's CVODE_BDF and ODEInterface's `radau` tend to do best, while both of those package don't provide excellent methods for non-stiff equations, and don't work on odd number types.

Also, if it was package-wise it would clash.

src/algorithms/ode_algorithms.jl
ChrisRackauckas: Then how do you do defaults?

src/DiffEqBase.jl
ChrisRackauckas: The defaults code could just move to DifferentialEquations.jl if you think that's best.

ChrisRackauckas: Package-based default-choosing algorithms could still be available by the user passing in the package-wide abstract type?

src/DiffEqBase.jl
ChrisRackauckas: How does that PR handle Sensitivity functions?

ChrisRackauckas: Nevermind, I see. SensitivityFunction didn't really need to subtype ParameterizedFunction after all.

ChrisRackauckas:
I'm implementing these changes by hand to make sure everything (that is used) finds the right home and tests pass locally before merging.

ChrisRackauckas:
Note that I will be pushing to finish the SDE stuff today as well so that way those tests don't get in the way.

mauro3:
Sounds good.  I don't have time today and tomorrow I will be on a long haul flight. So I will not give much feedback.

ChrisRackauckas:
It's okay. The SDE stuff will intentionally look almost exactly like the ODE stuff (taking into account your changes) except the SDE problems will take two functions and a `NoiseProcess`. 

For the noise process I think it might be easiest to throw the draft to master and make sure everything is all working, and if you have any comments from there handle that. For now it's far too small to be its own thing.

ChrisRackauckas:
This should all be done by https://github.com/JuliaDiffEq/DiffEqBase.jl/commit/8f23efa8dc27bebb223952333cf8af9eee1437c3 . You might want to double check (it has the stochastic changes in there too. Again, nothing big except the noise process). I'll have to put in some minor tags all around for everything to reflect this (until then, tests fail almost everywhere because two package pretty much both export the same thing somewhere).

PR #6, opened 2016-11-16
WIP: algorithm registration and default alg

mauro3:
Here a possible implementation of how default solvers could work decentralized.  I'm not sure this is the way to go, as different solvers will be picked depending on what packages are "used".  But I like the registration of algorithms into the `ode_algs` dict.


ChrisRackauckas:
I'm not convinced this is a good way to handle this:

- Precision levels are not precise. What do you think "high precision means"? It probably means something different than when I tell you that the Feagin 14th order method tends to be a good method when tolerances are around 1e-40. 
- I don't think we should have a requirement to register methods as stiff or non-stiff. This leaves out switching methods like LSODE, and I am hoping to get some switching methods out soon. And I think that methods which switch behavior depending on stiffness will be a clear user-friendly (and sometimes performance increasing) option.
- If the grades can change anytime they might not correctly match with the benchmarks
- This can't easily be graphed. How do you chart what algorithm is chosen? [I'd like to chart the multi-algorithm like `\`](https://www.mathworks.com/help/matlab/ref/mldivide.html). 
- Other reasons why this needs the full problem + kwargs: it needs to know the types for compatibility, it should know more precisely the precision, whether or not Jacobians, Inverse Jacobians, Inverse W's exist, etc.

What you did get right here is that using Enum types for the choices and having them as separate kwargs makes a whole lot of sense, more sense than just the `alg_hints`. 

This choices are all pretty obvious and objective from theory, papers, implementation details, and most importantly, benchmarks. However, I am not seeing how an automated algorithm like this can "get it right" without running the benchmarks. 

Maybe I had this all wrong by trying to include this in DiffEqBase in the first place. Getting default algorithms in a large ecosystem of packages necessitates that the "batteries are included" to get it correct. Maybe as you said, this is better off living in DifferentialEquations.jl since it is the "batteries are included" package, and this is not a development/interface issue and rather a usage issue. That would then mean that packages which are used on their own won't have a defaults which I think is fine (if you're picking specific parts of the ecosystem, you're already doing so with some intention as to which algorithm you want to use). What do you think?

mauro3:
Yes, let DifferentialEquations.jl pick defaults.  

However, I think it is good to keep all currently available solvers in a central place (which gets updated as more packages are used).  I thought that classifying them a bit could be helpful, but maybe not; you're the expert on this.  Note that I am by no means saying that the current classification scheme is good.  (Maybe `LSODE` could be `both`).  How about classifying them as far as it is needed to run the benchmarks over them?  That way it could just hook into this and it would be each package's duty to get the classification right.  I think that means stiff vs nonstiff (or both), adaptive/fixed and Float-only vs anytype.

Concerning `@enum`s: I'm actually not so sure as they either pollute the namespace a lot or require prefixing with the module.  They are nice though as they clearly enumerate the options.  So maybe simple symbols are better (and pretty standard in Julia).

ChrisRackauckas:
>How about classifying them as far as it is needed to run the benchmarks over them? That way it could just hook into this and it would be each package's duty to get the classification right. I think that means stiff vs nonstiff (or both), adaptive/fixed and Float-only vs anytype.

That's an interesting idea, but for other reasons. I think having the dictionary of existing algorithms lets the multi-algorithm choose from the algorithms which exist. One case where this is nice is that because it could then use ODEInterface, which is the one package which is part of the interface but isn't included by default in DifferentialEquations.jl because so many people run into Fortran compilation issues. However, it could do an "if ODEInterface is installed and imported, then Radau is a good choice here" kind of thing.

I don't think registering stiff vs non-stiff will help much because we're at a point already where that would pull in too many algorithms to be useful. For the benchmarks, you need to declare which 8 or so algorithms you want to show in a given run because that already starts be hard to see. I think we have like over 20 non-stiff algorithms already, not including the 100 tableaus I implemented. So "plot all non-stiff algorithms" or anything like that is already far past its utility.

All of that said, registering the algorithm like this with adaptive vs fixed, and what types are compatible will be good for errors. If a method is registered as fixed timestep and doesn't give a dt, should an error be thrown? I think that a `compatible_types` part could be useful, though difficult. While `anytype` seems reasonable, and I think that a lot of people would believe that when properly coded their algorithm "should work for any type", the details are much more subtle. Maybe I should document this somewhere, but here's a list off the top of my head of details you may have overlooked if you haven't tested these cases:

- For Unitful numbers, you need to have every rate be a different type than the independent variables. So the `k`'s and all of those parts in the interpolations must be a different type than both `t` and `u` .
- For more general abstract arrays which are not contiguous, default constructors and shaped constructors do not exist. So every cache must be defined using `similar` (or a `zeros` call with `similar`)
- Julia's `norm` function is only defined for 1 and 2 dimensional tensors. General `N` dimensional tensors will error if you do `norm(error)` and the error is an error estimate from an equation on `N>3` dimensional tensors.
- ForwardDiff.jl, Calculus.jl, and FiniteDiff.jl are incompatible with complex numbers. Therefore any routine which requires a numerical Jacobian (thus not including the case where the Jacobian is passed in) needs its own routine for calculating said Jacobian since no package (that I know of) can handle this case.
- If you are using the `norm` function, you still can't do matrices of BigFloats because `norm(u)` is the largest singular value, and Julia doesn't provide a fallback for `svd` on BigFloats in Base. This can easily be fixed by adding a dependency to GenericSVD.jl.
- Compatibility with things like Rational and Int for timesteps requires that `eps` isn't used (and therefore it should be kept separate from compatibility with `uType`). And this still has a difference between adaptive and non-adaptive. I only handle the non-adaptive case, but I could see someone changing the adaptive algorithm when the number type is rational to just do halving/doubling to support that case.

And that's just off the top of my head. So likely no algorithm is actually compatible with any number types (I'm trying, but as you can see from this list, it's not easy). And the degree to which these issues show up can be very subtle (N>3 tensors, BigFloats in only some cases, etc). I don't know of a good way to characterize this at all though, it seems like a hidden treasure of knowledge that I don't necessarily know how to document/share without going into every detail, but is important for users to know (a good error message saying that a type isn't supported would be helpful).

mauro3:
Yes, that certainly looks tricky.  So I think it will be Float vs some-Others.  I still think that stiff vs non-stiff is useful, or maybe implicit vs explicit vs both (LSODE, IMEX).  Something flexible would be good. I'll polish this a bit, and remove the default-stuff.

ChrisRackauckas:
Shouldn't IMEX be a separate problem? I don't think it's possible to automatically split a function like that: the user needs to provide a stiff and a nonstiff part.

Stiff vs nonstiff is still better than explicit vs implicit, which can "mean the wrong thing" (Rosenbrock methods are explicit). stiff vs nonstiff vs switching vs semistiff (semistiff, like highly stable RK methods (ROCK2/4), stabilized high order RKs). But "semi" has two different meanings here (IMEX).

Also, is the dictionary needed? Couldn't this stuff be fields in the algorithm type?

mauro3:
To discover all the solvers we could also traverse the type-tree of `AbstractODEAlgorithm` but that would be harder.

> Also, is the dictionary needed? Couldn't this stuff be fields in the algorithm type?

No, not as the interface is now, as we don't specify how a alg-type can be instantiated. Also, these properties are the same for all instances of a solver type.  In OO these would be called class-variables but in Julia those don't exist.  There are many ways to store that info with the type, one of those is a dict.  (If it hasn't changed that is what e.g. the julia help-system does under the hood)

ChrisRackauckas:
I see

ChrisRackauckas:
>To discover all the solvers we could also traverse the type-tree of AbstractODEAlgorithm but that would be harder.

When would this ever need to be done?

ChrisRackauckas:
I don't think we're going this direction.

Issue #10, opened 2016-11-25
Make `isinplace` a trait


mauro3:
I think `isinplace` should be a trait and not a type parameter.  Thoughts?

ChrisRackauckas:
Probably. That seems to make sense.

The `inplace` story should likely get more complex. Currently I have it so that way you can give inplace functions and non-inplace functions, which is useful since equations on immutables (numbers) cannot be inplace. And I always divided it by `Number` and `AbstractArray`. A non-inplace function on an `AbstractArray` would auto-convert to an in-place function so the `AbstractArray` methods would work.

There are two things here. The first is generalizing these ideas, and secondly it's making them more compatible. The ideas should be generalized by making it `Indexable` (`AbstractArray`) and `NotIndexable` (not `AbstractArray`). This would expand `Number` to include ApproxFun `Fun` types directly, and fully support their changing length nature.

However, if you check the OrdinaryDiffEq integrators, to support these two types I have two dispatches for each method. There has to be a better way:

```julia
du = f(t,u,du)
```

I wonder if that works correctly and gets the right performance...

ChrisRackauckas:
It's hard to do because `isinplace` has a different meaning (different number of parameters) depending on the type of problem. This would mess with promotion, so it has to be placed during the construction of the type. This would require some kind of parametric trait to do correctly, so I'll close unless something comes up.

PR #11, opened 2016-11-25
WIP: ODE->DAE promotion

### LINKED ISSUE #1: Create DAE types and a promotion structure
As discussed in https://github.com/JuliaDiffEq/Roadmap/issues/5, we have set out for the DAE types something like `DAEProblem` for ODEs defined by an implicit function `f(t,u,du,resid)` where `resid` is the residual (does anyone have a better common name for that arg? Or is that good?), and `MassMatrixODEProblem` for problems with a mass matrix (extends the ODE problem to have a spot for the mass matrix). What should be name that field? `mass_matrix`? We would then implement a promotion structure:

- ODEProblem -> MassMatrixODEProblem by making the `mass_matrix = I` the uniform scaling operator (allows one to do a type check as well to turn off extra calculations, and acts nicely in the common ways mass matrices are used).
-  MassMatrixODEProblem -> DAEProblem by defining the residual problem `out = M*du - f(t,u,du)` (as real code, of course this won't work)? In this promotion we also need to provide `du0` and that can be determined by an evaluation of `f`. 
- Is there an intermediate common case we are missing?

Ideally you'd then just write your algorithm at the top level with enough promotion information exposed so that way you don't lose performance. So say for my Rosenbrock algorithms it can solve a `MassMatrixODEProblem`, so what I write those in a manner such that, if someone does

```julia
solve(prob,Rosenbrock23;kwargs)
```

and prob is an `ODEProblem`, it will be promoted to a `MassMatrixODEProblem` with `mass_matrix = I` and given to my Rosenbrock routine.

I don't know how to do promotions really at all, so could someone explain this @mauro3?

Note that some starter code is in AlgebraicDiffEq.jl, but this is probably easier to just re-write given what we have from `ODEProblem` in DiffEqBase since a lot has changed. For each of these we should also have an appropriate `TestProblem` and `build_..._problem` function for testing. 

I can do this over the weekend if needed, but maybe someone else who's more interested in DAEs should take a stab at this to get to know the infrastructure well and identify potential problems. @pwl or @mauro3?

Once this is together, I'll make a common interface dispatch for Sundials' `idasol` to move that code out of AlgebraicDiffEq.jl. One that's done, AlgebraicDiffEq.jl will be a shell package, which is fine because it will be ready for people to add methods if they so please (I'll be leaving my Rosenbrock functions and creating new ones in OrdinaryDiffEq.jl since I tend to consider mass matrix problems more ODEs. That can be a discussion though).

mauro3:
Fixes #1

It wasn't as easy as I thought it would be...

Note, for the tests to pass [this line](https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/e0f46ffb37d0e908083df4df5ce15815d050dad0/src/sde/sde_solve.jl#L1) needs to be changed to `function solve{uType,tType,isinplace,NoiseClass,F,F2,F3,algType<:AbstractSDEAlgorithm}(`.

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -5,30 +5,44 @@ module DiffEqBase
   using RecipesBase, Parameters, RecursiveArrayTools, SimpleTraits
   using Ranges # For plot recipes with units
   import Base: length, size, getindex, endof, show, print,
-               next, start, done, eltype, eachindex
+               next, start, done, eltype, eachindex, convert, promote
 
-  # Problems
+  ## Problems
   "`DEProblem`: Defines differential equation problems via its internal functions"
   abstract DEProblem
   abstract DEElement
   abstract DESensitivity
-  abstract AbstractODEProblem{uType,tType,isinplace,F} <: DEProblem
+
+  # Initial Value Problems (IVP)
+  abstract AbstractIVPProblem{uType,tType,isinplace,F} <: DEProblem
+  "Checks whether uses an in-place function or not"
+  isinplace{uType,tType,I,F}(::Type{AbstractIVPProblem{uType,tType,I,F}}) = I
+  isinplace{A<:AbstractIVPProblem}(::Type{A}) = isinplace(supertype(A))
+  isinplace{A<:AbstractIVPProblem}(::A) = isinplace(A)
+  @traitdef IsInplace{P}
+  @traitimpl IsInplace{P} <- isinplace(P)
+
+  # IVP subtypes
+  abstract AbstractODEProblem{uType,tType,isinplace,F} <: AbstractIVPProblem{uType,tType,isinplace,F}
   abstract AbstractODETestProblem{uType,tType,isinplace,F} <: AbstractODEProblem{uType,tType,isinplace,F}
-  abstract AbstractSDEProblem{uType,tType,isinplace,NoiseClass,F,F2,F3}  <: DEProblem
+  abstract AbstractSDEProblem{uType,tType,isinplace,NoiseClass,F,F2,F3}  <: AbstractIVPProblem{uType,tType,isinplace,F}
   abstract AbstractSDETestProblem{uType,tType,isinplace,NoiseClass,F,F2,F3}  <: AbstractSDEProblem{uType,tType,isinplace,NoiseClass,F,F2,F3}
-  abstract AbstractDAEProblem{uType,duType,tType,isinplace,F} <: DEProblem
+  abstract AbstractDAEProblem{uType,duType,tType,isinplace,F} <: AbstractIVPProblem{uType,tType,isinplace,F}
   abstract AbstractDAETestProblem{uType,duType,tType,isinplace,F} <: AbstractDAEProblem{uType,duType,tType,isinplace,F}
+
+  # Misc other problems
   abstract AbstractDDEProblem <: DEProblem
   abstract AbstractPoissonProblem <: DEProblem
   abstract AbstractHeatProblem <: DEProblem
 
-  # Algorithms
+  ## Algorithms
   abstract DEAlgorithm
-  abstract AbstractODEAlgorithm <: DEAlgorithm
-  abstract AbstractSDEAlgorithm <: DEAlgorithm
-  abstract AbstractDAEAlgorithm <: DEAlgorithm
+  abstract IVPAlgorithm <: DEAlgorithm
+  abstract AbstractODEAlgorithm <: IVPAlgorithm
+  abstract AbstractSDEAlgorithm <: IVPAlgorithm
+  abstract AbstractDAEAlgorithm <: IVPAlgorithm
 
-  # Solutions
+  ## Solutions
   abstract DESolution
   abstract AbstractODESolution <: DESolution
   abstract AbstractSDESolution <: AbstractODESolution # Needed for plot recipes
@@ -37,7 +51,7 @@ module DiffEqBase
   abstract AbstractFEMSolution <: DESolution
   abstract AbstractSensitivitySolution
 
-  # Misc
+  ## Misc
   "`Mesh`: An abstract type which holds a (node,elem) pair and other information for a mesh"
   abstract Mesh
   "`Tableau`: Holds the information for a Runge-Kutta Tableau"
@@ -57,7 +71,12 @@ module DiffEqBase
   include("problems/sde_problems.jl")
   include("problems/dae_problems.jl")
 
-  function solve end
+  solve(p::DEProblem, a::DEAlgorithm; kwargs...) = solve(promote(p, a), a; kwargs...)
+  promote(p::DEProblem, a::DEAlgorithm) =
+      error("Algorithm (solver) of type `$(typeof(a))` is not applicable to problem of type `$(typeof(p))`")
+  convert{P1<:DEProblem}(::Type{P1}, p::DEProblem) =
+      error("Cannot convert problem type $(typeof(p)) into problem type $P1")
+  convert{P1<:DEProblem}(::Type{P1}, p::P1) = p
 
   export DEProblem, DESolution, DEParameters, AbstractDAEProblem, AbstractDDEProblem,
          AbstractODEProblem, AbstractSDEProblem, DAESolution, DEIntegrator, Mesh,
--- a/src/problems/dae_problems.jl
+++ b/src/problems/dae_problems.jl
@@ -1,3 +1,22 @@
+"""
+Defines a differential algebraic equation (DAE) (or implicitly defined
+ODE if there are no purely algebraic equations) of form: `f(t,u,du/dt) = 0`
+
+Input:
+- `f(t,u,du,out)` or `f(t,u,du)`
+- `u0` initial conditions (IC) on `u`
+- `du0` IC on `du`
+- `tspan` tuple of start and end-time `(t0,tend)`
+
+Extra function, such as the Jacobian can be provided by function overloading:
+- `f(::Val{:jac}, t, u, du, alpha, out)` where `jac = df/du + alpha* df/d(du)`
+
+Notes:
+- the types of inputs determines the types used in the numerical
+  solver.  For instance, using `Int`s in `tspan` will not work.
+- ideally the IC are consistent, i.e. `f(tspan[1], u0, du0)==0`.  Some
+  solvers will solve for consistent IC but this may fail.
+"""
 type DAEProblem{uType,duType,tType,isinplace,F} <: AbstractDAEProblem{uType,duType,tType,isinplace,F}
   f::F
   u0::uType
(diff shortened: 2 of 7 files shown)


src/DiffEqBase.jl
```diff
@@ -5,30 +5,44 @@ module DiffEqBase
...
-  # Problems
+  ## Problems
   "`DEProblem`: Defines differential equation problems via its internal functions"
   abstract DEProblem
   abstract DEElement
   abstract DESensitivity
-  abstract AbstractODEProblem{uType,tType,isinplace,F} <: DEProblem
+
+  # Initial Value Problems (IVP)
+  abstract AbstractIVPProblem{uType,tType,isinplace,F} <: DEProblem
```
mauro3: Adding this supertype was needed for a general inplace function.

src/DiffEqBase.jl
```diff
@@ -5,30 +5,44 @@ module DiffEqBase
...
+  "Checks whether uses an in-place function or not"
+  isinplace{uType,tType,I,F}(::Type{AbstractIVPProblem{uType,tType,I,F}}) = I
+  isinplace{A<:AbstractIVPProblem}(::Type{A}) = isinplace(supertype(A))
+  isinplace{A<:AbstractIVPProblem}(::A) = isinplace(A)
+  @traitdef IsInplace{P}
+  @traitimpl IsInplace{P} <- isinplace(P)
```
mauro3: As mentioned over in #10, maybe ditch the inplace type parameter completely in favor of this trait.

src/problems/dae_problems.jl
```diff
@@ -1,3 +1,22 @@
...
+ODE if there are no purely algebraic equations) of form: `f(t,u,du/dt) = 0`
+
+Input:
+- `f(t,u,du,out)` or `f(t,u,du)`
+- `u0` initial conditions (IC) on `u`
+- `du0` IC on `du`
+- `tspan` tuple of start and end-time `(t0,tend)`
+
+Extra function, such as the Jacobian can be provided by function overloading:
+- `f(::Val{:jac}, t, u, du, alpha, out)` where `jac = df/du + alpha* df/d(du)`
```
mauro3: This is what PETSc and DASSL.jl use for jacobians.  Is that how you do it too?

src/problems/dae_problems.jl
```diff
@@ -37,3 +56,78 @@ function show{uType,tType,isinplace,F}(io::IO,prob::AbstractODEProblem{uType,tTy
...
+Input:
+- T: type of new AbstractIVPProblem
+- prob: old problem instance
+- Val{:sometag}: convert `:sometag` method
+- `f_new`: the generic function to which the converted method is added
+
+Out:
+- nothing
+"""
+function convert_overloaded!{T<:AbstractIVPProblem}(::Type{T}, prob::AbstractIVPProblem, ::Val, f_new)
```
mauro3: For each overloaded-function and <:AbstractIVPProblem, a conversion needs to be defined.

test/promotion.jl
```diff
@@ -0,0 +1,16 @@
+# Checks that promotion to more general DEProblem types works.
+using DiffEqProblemLibrary, Sundials
+
+# prob = prob_ode_2Dlinear this one does not work with IDA
+for prob in [prob_ode_linear, prob_ode_vanderpol]
```
mauro3: Do you have a test problem with a analytic Jocobian?  Also note that `prob_ode_2Dlinear` doesn't seem to work with IDA, as the residual which gets passed into `f` is a vector instead of a matrix.

src/problems/dae_problems.jl
ChrisRackauckas: Ahh. This is where a lot of work will be. I think just handling Jacobians for now is fine.

src/DiffEqBase.jl
ChrisRackauckas: 👍. Yes. It should have the same treatment as the extended functions so that way uses like in the Sundials functions will be easy to do:

https://github.com/JuliaDiffEq/Sundials.jl/blob/master/src/common.jl#L189

src/DiffEqBase.jl
ChrisRackauckas: 👍 . It's probably more generally useful anyways. Not sure about all of the type parameters though.

src/problems/dae_problems.jl
ChrisRackauckas: I just had `jac = df/du` since I haven't dealt with fully implicit equations yet. I guess the other part is useful for DAEs? Check the Sundials docs for what it needs. It's probably similar.

src/problems/ode_problems.jl
```diff
@@ -1,3 +1,19 @@
...
+Defines a ordinary differential equation (ODE) in explicit form
+du/dt = f(t,u).
+
+Input:
+- `f(t,u,out)` or `f(t,u)`
+- `u0` initial conditions (IC) on `u`
+- `tspan` tuple of start and end-time `(t0,tend)`
+
+Extra function, such as the Jacobian can be provided by function overloading:
+- `f(::Val{:jac}, t, u, out)` where `jac = df/du + alpha* df/d(du)`
```
ChrisRackauckas: Yes, with `jac = df/du`

src/problems/ode_problems.jl
```diff
@@ -1,3 +1,19 @@
...
+
+Notes:
+- the types of inputs determines the types used in the numerical
+  solver.  For instance, using `Int`s in `tspan` will not work.
```
ChrisRackauckas: Be more specific here. Using `Int` in `tspan` will define a Problem which requires the each timestep to be an `Int` (this should work with most solvers when adaptivity is turned off)

src/problems/dae_problems.jl
```diff
@@ -37,3 +56,78 @@ function show{uType,tType,isinplace,F}(io::IO,prob::AbstractODEProblem{uType,tTy
...
+    warn("No conversion to $T for overloaded function method $val.  Not converting...")
+end
+@traitfn function convert_overloaded!{T<:AbstractIVPProblem, P2<:AbstractODEProblem,Fnew}(::Type{T}, prob::P2::IsInplace, ::Val{:jac}, f_new::Fnew)
+    f = prob.f
+    function (::Fnew)(::Val{:jac}, t, u, du, alpha, out)
+        f(Val{:jac}(), t, u, out)
+        out .= .+(out, alpha*I) # alpha*df/ddu == diag(alpha)
```
ChrisRackauckas: Mark as change in 0.6

src/problems/dae_problems.jl
```diff
@@ -37,3 +56,78 @@ function show{uType,tType,isinplace,F}(io::IO,prob::AbstractODEProblem{uType,tTy
...
+###
+# Promotion & conversion
+promote(prob::ODEProblem, ::AbstractDAEAlgorithm) = convert(DAEProblem, prob)
+promote(prob::ODETestProblem, ::AbstractDAEAlgorithm) = convert(DAETestProblem, prob)
+function convert{P<:AbstractDAEProblem}(::Type{P}, prob::AbstractODEProblem)
+    @unpack f,u0,tspan = prob
+    f_new = convert_obj_fn(P,prob)
+    # IC
+    if isinplace(prob)
+        du0 = similar(u0) # TODO: this needs to be Unit-stable (Chris, help!)
```
ChrisRackauckas: `du0 = similar(u0/tspan[1])` should be fine. Probably should have a comment for why that division is there.

test/promotion.jl
ChrisRackauckas: IDA will automatically reshape it though. 

https://github.com/JuliaDiffEq/Sundials.jl/blob/master/src/common.jl#L195

So whatever is going on is a little bit more subtle. What's the error message?

ChrisRackauckas: >Do you have a test problem with a analytic Jocobian?

The ones defined using the ParameterizedFunctions.jl macros should have Jacobians. You can also double check that the form is correct by looking at `f.Jex` to see the expression it used, or `f.symjac` for the matrix of symbolic derivatives.

mauro3: The fix can either be changing [Sundials.jl line](https://github.com/JuliaDiffEq/Sundials.jl/blob/c7881996cbddf74e8e22bb290233fda2d5e9b061/src/common.jl#L196) to `        f! = (t,u,du,out) -> (prob.f(t,reshape(u,sizeu),reshape(du,sizedu),reshape(out,sizeu));` or make the `out` in this [PR line](https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/11/files#diff-1e8a894e987831097b0af6710596d269R94) always a vector.  Either is fine, but probably should pick the more logical one: is the residual the same shape as `u` or always a vector?

ChrisRackauckas: Not necessarily either: `out` can be larger than `u` or `du` (constrained ODE), but then I don't know if that means we restrict it to a vector. But we don't necessarily even know the size of `out` until after function calculation.

src/problems/dae_problems.jl
ChrisRackauckas: That seems general enough. I'll take that as the DAE Jacobian. Need a good way to document it.

test/promotion.jl
ChrisRackauckas: >Why couldn't we just have size(u)==size(du)? 

Yes, this one should hold.

>For constraint problems, we'd maybe would need to introduce additional DOFs to have length(u)==length(out)

This one won't, and I don't think more DOFs is the right solution. How would you suppose adding more DOFs to `u`? Just add a bunch of spots with 0's? That will mess with plotting, error estimators, give more work to interpolations. What would be helped by making `u` larger? This will be a problem for any problem with a singular mass matrix (in a reduced form). 

Maybe we operate on the assumption that `size(u)==size(du)==size(out)`, but allow for users to provide (via keyword arg) a template for `out`? It would have to be a "template", i.e. something to `copy` and `similar` since `size(out)` misses lots of information for the problem (index choices, relevant type information, etc). Algorithms which are restrictive on the types of inputs (like IDA, DASSL, etc.) could then just use `size` on the template and only support `Array{T,N}`, but it allows this to grow towards more comprehensive support.

src/problems/dae_problems.jl
```diff
@@ -37,3 +56,78 @@ function show{uType,tType,isinplace,F}(io::IO,prob::AbstractODEProblem{uType,tTy
...
+        error("Cannot convert non-test problem into test-problem.")
+    end
+end
+
+#"Convert the objective function `f` or prob"
+@traitfn function convert_obj_fn{T<:AbstractDAEProblem, P2<:AbstractODEProblem}(::Type{T}, prob::P2::IsInplace)
+    f = prob.f
+    return function (t,u,du,out)
+        f(t,u,out)
+        out .= .-(out, du) # change in 0.6
```
ChrisRackauckas: (-).(...)

ChrisRackauckas:
I thought this would be much easier to do and support. But thanks for trying this out, it makes me realize that this is really really hard to do in general. Also I'm not convinced it's very useful anymore. The original idea was so that one can "just write a DAE BDF" like DASSL.jl and have that work on ODEs. Well, DAE versions have extra behaviors which make them not as optimized for ODEs anyways, so you really wouldn't want to do this unless the solver was actually aware you were doing this. But that would mean building in the support at the solver level and having it `if typeof(prob)<:AbstractODEProblem ...`.

The newer approach to allowing easy interface between problem types is through overloading. `DynamicalODEProblem`s and other split ODEs employ this to great effect to "act like" ODE functions when they need to. Linear operators do this at the type level as well. In the end, this can be done in very specific cases like that much easier.

Thank you @mauro3 for your work because it really solidifies what this choice would look like.

PR #15, opened 2016-12-19
Add a `vars` parameter to plot phase diagrams

Armavica:
The most precise way to specify which variables to plot is to pass a
list of tuples `(x, y)` where `x` and `y` are positive integers to
indicate the corresponding dependent variable, or 0 to refer to time
(the independent variable).  Example:

    vars = [(0,1), (1,3), (4,5)]

will plot on the same graph `sol[1]` as a function of `t`, `sol[3]` as a
function of `sol[1]`, and `sol[5]` as a function of `sol[4]` (whatever
that could mean).

Everything could be done with that, but for convenience several
shortcuts are allowed:

* Everywhere in a tuple position where we only find an integer, this
variable is plotted as a function of time.  For example, the list above
is equivalent to:

        vars = [1, (1,3), (4,5)]

    and

        vars = [1, 3, 4]

    is the most concise way to plot the variables 1, 3, and 4 as a function
of time.

* It is possible to omit the list if only one plot is wanted: `(2,3)`
and `4` are respectively equivalent to `[(2,3)]` and `[(0,4)]`.

* Finally, a tuple containing one or several lists will be expanded by
associating corresponding elements of the lists with each other:

        vars = ([1,2,3], [4,5,6])

    is equivalent to

        vars = [(1,4), (2,5), (3,6)]

    and

        vars = (1, [2,3,4])

    is equivalent to

        vars = [(1,2), (1,3), (1,4)]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ os:
   - osx
 julia:
   - 0.5
-  - nightly
+matrix:
+  allow_failures:
+    - julia: nightly
 notifications:
   email: false
 # uncomment the following lines to override the default test script
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,9 @@
 environment:
   matrix:
-  - JULIAVERSION: "julialang/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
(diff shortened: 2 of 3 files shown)


ChrisRackauckas:
I put this threw a bunch of tests indirectly over the last few days and it never failed me, so I am pulling this in now. Great work!

PR #20, opened 2017-02-13
Use new macros to simplify code and eliminate deprecation warnings in v0.6

ScottPJones:
This change eliminates all of the deprecation warnings about `abstract` types in DiffEqBase, when building under v0.6.

### DIFF
--- a/src/typemacros.jl
+++ b/src/typemacros.jl
@@ -1,0 +1,92 @@
+__precompile__()
+module TypeMacros
+export @abstract_type, @struct_type, @public_abstract_type, @public_abstract_subs
+
+macro abstract_type(exp) ; _at(exp) ; end
+macro struct_type(exp)   ; _st(exp) ; end
+macro public_abstract_type(exp); _pat(exp); end
+macro public_abstract_subs(exp); _pas(exp); end
+
+@static if VERSION < v"0.6-"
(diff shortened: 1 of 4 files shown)


src/typemacros.jl
```diff
@@ -0,0 +1,92 @@
+__precompile__()
+module TypeMacros
+export @abstract_type, @struct_type, @public_abstract_type, @public_abstract_subs
+
+macro abstract_type(exp) ; _at(exp) ; end
+macro struct_type(exp)   ; _st(exp) ; end
+macro public_abstract_type(exp); _pat(exp); end
+macro public_abstract_subs(exp); _pas(exp); end
+
+@static if VERSION < v"0.6-"
```
tkelman: `v"0.6.0-dev.2746"` otherwise the package will be needlessly broken on earlier prereleases

tkelman:
Compat already handles abstract type and immutable is not deprecated or used here, so this is an ill advised overcomplicated was to solve a problem where at-compat would have done fine

ChrisRackauckas:
Yeah, I agree. I'll revert the change to an `@compat` solution when I have time. Or @ScottPJones can. Instead of resorting to macros to handle the exports, I think the exports should just be better organized.

PR #24, opened 2017-03-25
Imex problem

thesharpshooter:
Waiting for review

### DIFF
--- a/src/problems/imex_problem.jl
+++ b/src/problems/imex_problem.jl
@@ -1,0 +1,39 @@
+type IMEXProblem{uType,tType,isinplace,F,G} <: AbstractODEProblem{uType,tType,isinplace,F,G}
+  f::F
+  g::G
+  u0::uType
+  tspan::Tuple{tType,tType}
+end
+
+type IMEXTestProblem{uType,AType,tType,isinplace,F,G} <: AbstractODETestProblem{uType,tType,isinplace,F,G}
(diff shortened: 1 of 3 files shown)


src/problems/imex_problem.jl
```diff
@@ -1,11 +1,11 @@
-type IMEXProblem{uType,tType,isinplace,F,G} <: AbstractODEProblem{uType,tType,isinplace,F,G}
+type IMEXProblem{uType,tType,isinplace,F,G} <: AbstractIMEXProblem{uType,tType,isinplace,F,G}
```
ChrisRackauckas: You need to make that abstract type or this will error.

ChrisRackauckas:
Use this as a reference for defining an abstract type:

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/v0.10.0/src/DiffEqBase.jl

You need to define it and export it. It probably should `<: AbstractODEProblem{ ...} `. So something like:

```julia
@compat abstract type AbstractIMEXProblem{uType,tType,isinplace,F,G} <: AbstractODEProblem{uType,tType,isinplace,F}
export AbstractIMEXProblem, IMEXProblem, IMEXTestProblem
```

You should test it to make sure it doesn't error too. Just do it locally, as DiffEqBase doesn't have adequate tests for problem building (it should get some)

Issue #30, opened 2017-04-26
`ODEProblem`: Check whether `tType` is `Any`


ranocha:
I've encountered some strange error messages if `tType == Any` in `ODEProblem`:

```
julia> using DifferentialEquations
julia> f(t,u) = u
julia> prob1 = ODEProblem(f, u0, (0.,1))
julia> solve(prob1)
ERROR: MethodError: no method matching OrdinaryDiffEq.DEOptions{uEltype,uEltypeNoUnits,tTypeNoUnits,tType,F2,F3,F4,F5,F6,tstopsType,ECType,SType}(::Int64, ::Int64, ::Bool, ::Bool, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Rational{Int64}, ::OrdinaryDiffEq.#ODE_DEFAULT_NORM, ::Void, ::DataStructures.BinaryHeap{Any,DataStructures.LessThan}, ::DataStructures.BinaryHeap{Any,DataStructures.LessThan}, ::DataStructures.BinaryHeap{Any,DataStructures.LessThan}, ::Void, ::Bool, ::Int64, ::String, ::OrdinaryDiffEq.#ODE_DEFAULT_PROG_MESSAGE, ::Bool, ::Bool, ::Float64, ::Float64, ::Float64, ::Bool, ::Bool, ::DiffEqBase.CallbackSet{Tuple{},Tuple{}}, ::OrdinaryDiffEq.#ODE_DEFAULT_ISOUTOFDOMAIN, ::OrdinaryDiffEq.#ODE_DEFAULT_UNSTABLE_CHECK, ::Bool, ::Bool, ::Bool, ::Bool, ::Bool)
Closest candidates are: ...
```

```
julia> using DifferentialEquations
julia> f(t,u) = u
julia> prob2 = ODEProblem(f, u0, (0,1.))
julia> solve(prob2)
ERROR: InexactError()
 in convert at ./rational.jl:68 [inlined]
 in Int64(::Rational{Int64}) at ./sysimg.jl:53
[...]

PR #31, opened 2017-04-26
Use 'promote_type(map(typeof,tspan)...)' instead of 'eltype(tspan)'

ranocha:
As discussed in #30.


ChrisRackauckas:
I'll look at this after RecursiveArrayTools gets tagged and DiffEqNoiseProcess gets registered. Then tests should pass and I'll merge this.

PR #33, opened 2017-05-09
Don't use Ranges in v0.6

ScottPJones:
Per Tim Holy's comments, following the code in Unitful.jl

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -3,7 +3,12 @@ __precompile__()
 module DiffEqBase
 
 using RecipesBase, SimpleTraits, RecursiveArrayTools, Compat, Iterators, Juno
-using Ranges # For plot recipes with units
+@static if VERSION < v"0.6.0-dev.2390"
+    using Ranges # For plot recipes with units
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
You have to address:

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/5d7410b264b037260f5c6b57eda89783341aaf43/src/integrator_interface.jl#L75
https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/68d46263143bbe2a7b25f27d1925da6a0ee22010/src/solutions/solution_interface.jl#L61

Tests will pass because the tests can't generate plots... (that should get fixed sometime)

ScottPJones:
Darn!  Those just be linspace instead of Ranges.linspace, correct?

ChrisRackauckas:
>Darn! Those just become linspace instead of Ranges.linspace, correct?

According to @timholy, yes. Those need to run a special test though. Run this test file:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/units_tests.jl

and plot the results. You will get a unit error if there's a problem.

PR #39, opened 2017-05-10
Update solution_interface.jl

aytekinar:
Revert the previous change. `state >= length(sol)` is the correct one, since `next` increments the state before saving it! My bad...


ChrisRackauckas:
This is why we need a unit test here. 👍

aytekinar:
yeah! sorry for the confusion and the problem :(

PR #42, opened 2017-05-11
Add function to check keyword args on different solvers to fix #166

### LINKED ISSUE #166: Callback mutability is no thread-safe
@YingboMa I'm not sure if there's a good solution to this.

ScottPJones:


src/DiffEqBase.jl
```diff
@@ -16,6 +16,22 @@ import Base: resize!, deleteat!
 
 import RecursiveArrayTools: recursivecopy!, tuples
 
+# support function
+export check_keywords
+function check_keywords(alg, kwargs, warnlist)
```
ChrisRackauckas: should go in utils.jl

ScottPJones: OK!

PR #43, opened 2017-05-12
Improve support functions for warning output

ScottPJones:


ChrisRackauckas:
So the other setups work perfectly fine, this is all just cleanup?

ScottPJones:
This is needed for the new versions of the PR to work correctly (they call a one-line function to display the final compatibility help warning, and need check_keywords to return a bool now)

PR #44, opened 2017-05-13
Added reduction and related functionality for Monte Carlo simulations

elevien:
This is for [DiffEqMonteCarlo #10](https://github.com/JuliaDiffEq/DiffEqMonteCarlo.jl/pull/10). I decided to put `reduction` and `u_init` in the problem type. I've also added a field `converged` indicating whether reduction returned `true`.

Let me know if this is ok and I can finish up in DiffEqMonteCarlo.

### DIFF
--- a/src/problems/monte_problems.jl
+++ b/src/problems/monte_problems.jl
@@ -1,10 +1,12 @@
-type MonteCarloProblem{T,T2,T3} <: AbstractMonteCarloProblem
+type MonteCarloProblem{T,T2,T3,T4,T5} <: AbstractMonteCarloProblem
   prob::T
   prob_func::T2
   output_func::T3
+  reduction::T4
+  u_init::T5
 end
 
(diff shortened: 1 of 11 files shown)


src/problems/monte_problems.jl
```diff
@@ -1,10 +1,12 @@
...
   output_func::T3
+  reduction::T4
+  u_init::T5
 end
 
 MonteCarloProblem(prob::DEProblem;
                   output_func = (sol,i)-> sol,
-                  prob_func= (prob,i)->prob) =
-                  MonteCarloProblem(prob,prob_func,output_func)
+                  prob_func= (prob,i)->prob,reduction = (u,data,I)->(append!(u,data),false),u_init = Vector{Any}()) =
```
ChrisRackauckas: Just write that as `u_init = []`.

PR #45, opened 2017-05-13
MonteCarloSolution always takes converged

elevien:
I don't think there is any reason `MonteCarloSolution` constructor should work without taking `converged` as an argument. I also forgot about `MonteCarloTestSolution` in the last PR. I guess I should have run tests on both DiffEqBase and my changes to DiffEqMonteCarlo.  I don't think there is anything else I should need to change in base for the basic reduction functionality to work.

### DIFF
--- a/src/solutions/monte_solutions.jl
+++ b/src/solutions/monte_solutions.jl
@@ -6,7 +6,7 @@ type MonteCarloTestSolution{T,N,S} <: AbstractMonteCarloSolution{T,N}
   elapsedTime::Float64
   converged::Bool
 end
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
>I guess I should have run tests on both DiffEqBase and my changes to DiffEqMonteCarlo.

That's the bad part about having bad test coverage here... it's just a boring issue that needs to be solved.

ChrisRackauckas:
>I don't think there is any reason MonteCarloSolution constructor should work without taking converged as an argument.

I agree

PR #64, opened 2017-08-06
Indexing and broadcasting changes for DEDataArray

devmotion:
I changed the default implementation of `size` and removed the multidimensional indexing operations. Since DEDataArray is a subtype of AbstractVector it should act like a one-dimensional vector in my opinion, and not propagate size and indices of `x`.

This is shown by the following example on the current master branch:
```julia
julia> using DiffEqBase
julia> type MatrixType{T,S} <: DEDataArray{T}
           x::Matrix{T}
           f::S
       end
julia> b = MatrixType{Int,Float64}([1  2; 4  3], 3.0)
2×2 MatrixType{Int64,Float64}:
 1
 4
julia> b.x
2×2 Array{Int64,2}:
 1  2
 4  3
julia> b.^2
2×2 Array{Int64,2}:
  1   1
 16  16
julia> b .= b .+ 2
2×2 MatrixType{Int64,Float64}:
 5
 8
julia> b.x
2×2 Array{Int64,2}:
 5  2
 8  3
julia> b.+b.x
2×2 Array{Int64,2}:
 10   7
 16  11
```
so broadcasting works as expected but only along the first dimension; I assume this happens since `MatrixType` is an `AbstractVector`, and hence only the size of the first dimension is respected which equals 2.

Correcting the indices and size of DEDataArray gives:
```julia
julia> using DiffEqBase
julia> type MatrixType{T,S} <: DEDataArray{T}
           x::Matrix{T}
           f::S
       end
julia> b = MatrixType{Int,Float64}([1  2; 4  3], 3.0)
4-element MatrixType{Int64,Float64}:
 1
 4
 2
 3
julia> b.x
2×2 Array{Int64,2}:
 1  2
 4  3
julia> b.^2
4-element Array{Int64,1}:
  1
 16
  4
  9
julia> b .= b .+ 2
4-element MatrixType{Int64,Float64}:
 3
 6
 4
 5
julia> b.x
2×2 Array{Int64,2}:
 3  4
 6  5
julia> b.+b.x
ERROR: DimensionMismatch("arrays could not be broadcast to a common size")
Stacktrace:
 [1] _bcs1(::Base.OneTo{Int64}, ::Base.OneTo{Int64}) at ./broadcast.jl:70
 [2] _bcs at ./broadcast.jl:63 [inlined]
 [3] broadcast_shape at ./broadcast.jl:57 [inlined] (repeats 2 times)
 [4] broadcast_indices at ./broadcast.jl:53 [inlined]
 [5] broadcast_c at ./broadcast.jl:311 [inlined]
 [6] broadcast(::Function, ::MatrixType{Int64,Float64}, ::Array{Int64,2}) at ./broadcast.jl:434
```
So everything works as expected, but only for vectors even though the underlying array `x` is a matrix. The question is whether this is desirable or not.

If this behaviour is intended, the following changes can be removed again, and everything works as expected. Otherwise, if we would rather broadcast on `x` instead of the vector wrapper defined by DEDataArray I added some simple implementations of `_containertype`, `_promote_containertype`, `broadcast_c`, and `broadcast_c!`. I'm not sure whether this is the correct way to implement broadcast methods, I just followed the advice given at https://discourse.julialang.org/t/broadcast-for-custom-type/3497/3.

This then leads to the following behaviour:
```julia
julia> using DiffEqBase
julia> type MatrixType{T,S} <: DEDataArray{T}
           x::Matrix{T}
           f::S
       end
julia> b = MatrixType{Int,Float64}([1  2; 4  3], 3.0)
4-element MatrixType{Int64,Float64}:
 1
 4
 2
 3
julia> b.x
2×2 Array{Int64,2}:
 1  2
 4  3
julia> b.^2
2×2 Array{Int64,2}:
  1  4
 16  9
julia> b .= b .+ 2
4-element MatrixType{Int64,Float64}:
 3
 6
 4
 5
julia> b.x
2×2 Array{Int64,2}:
 3  4
 6  5
julia> b.+b.x
2×2 Array{Int64,2}:
  6   8
 12  10
```

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -4,8 +4,10 @@ module DiffEqBase
 
 using RecipesBase, SimpleTraits, RecursiveArrayTools, Compat, Juno, LinearMaps
 
-import Base: length, ndims, size, getindex, setindex!, endof, show, print,
-             next, start, done, eltype, eachindex, similar
+import Base: length, ndims, size, getindex, setindex!, show, print,
+             next, start, done, similar, IndexStyle
+
+import Base.Broadcast: _containertype, promote_containertype, broadcast_c, broadcast_c!
 
 import Base: resize!, deleteat!
 
--- a/src/data_array.jl
+++ b/src/data_array.jl
@@ -1,3 +1,20 @@
+# define AbstractArray interface of DEDataArray
+
+# iteration
+start(A::DEDataArray)  = start(A.x)
+next(A::DEDataArray, i) = next(A.x, i)
+done(A::DEDataArray, i) = done(A.x, i)
+length(A::DEDataArray) = length(A.x)
+
+# size
+size(A::DEDataArray) = (length(A.x),)
+ndims(A::DEDataArray) = 1
+
+# indexing
+getindex(A::DEDataArray, i::Int) = (A.x[i])
+setindex!(A::DEDataArray, x, i::Int) = (A.x[i] = x)
+Base.IndexStyle(::Type{<:DEDataArray}) = Base.IndexLinear()
+
 similar(A::DEDataArray) = deepcopy(A)
 
 @generated function similar(A::DEDataArray,dims::Tuple)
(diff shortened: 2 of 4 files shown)


ChrisRackauckas:
What about making `abstract type DEDataArray{T,N} <: AbstractArray{T,N} end` ? I am not sure why that was restricted to `AbstractVector` in the first place. But seems this could be useful?

devmotion:
First, I just wanted to propagate everything from `A.x` to `A` but this collided with `A` being an AbstractVector. I would prefer defining it as AbstractArray instead. Then we can also get rid of the extra broadcast methods.

devmotion:
Maybe we then should also add abstract types `DEDataVector` and `DEDataMatrix`, similar to AbstractVector and AbstractMatrix?

ChrisRackauckas:
>Maybe we then should also add abstract types DEDataVector and DEDataMatrix, similar to AbstractVector and AbstractMatrix?

Yeah, make some aliases. They're cheap and nice to have.

devmotion:
Closed in favour of https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/65

PR #65, opened 2017-08-07
Change DEDataArray implementation

devmotion:
These are some changes to `DEDataArray` we discussed in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/64. Since the implementation here is quite different to the other PR and does not contain any specific broadcast methods, I thought it might be cleaner to open a separate PR and close the old one.

`DEDataArray{T,N}` is now a subtype of `AbstractArray{T,N}`, which seems more natural. Of course, if the wrapped array `x` is not of type `AbstractArray{T,N}` the desired behaviour and propagation of `x` won't work but I guess that's a problem the user has to deal with.

Moreover I defined the convenient aliases `DEDataVector` and `DEDataMatrix`. I removed unneeded methods of the AbstractArray interface which will just work because of their defaults in Base, so e.g. only one implementation of `similar` is needed and also indexing can be defined in a more general way which would also work for nonlinear indices. The default behaviour of `similar(A::DEDataArray)` is changed, now instead of a deep copy of `A` it returns a DEDataArray that wraps `similar(A.x)` with a deep copy of all other fields. The old implementations of `similar` with multiple arguments (that are removed) already showed the same behaviour - so there was an inconsistency between the simple and more advanced cases.

The name of `copy_non_array_fields(!)` was misleading since it copies arrays, only array `x` is excluded. So I renamed it to `copy_fields(!)` but I'd like to hear your suggestions since also this name does not seem to fit perfectly. Moreover, `x` was always assumed to be the first argument of the constructor; this is changed, now creation of a new `DEDataArray` with an array `arr` and a `DEDataArray` template calls the constructor with `arr` always at its correct position. Another minor change is that `recursivecopy!` and `copy_non_array_fields(!)` now copy also fields that are not arrays (before they were just set to the value of the source).

Oh, and the example in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/64 gives:
```julia
julia> using DiffEqBase
julia> type MatrixType{T,S} <: DEDataMatrix{T}
                  x::Matrix{T}
                  f::S
              end
julia> b = MatrixType{Int,Float64}([1  2; 4  3], 3.0)
2×2 MatrixType{Int64,Float64}:
 1  2
 4  3
julia> b.x
2×2 Array{Int64,2}:
 1  2
 4  3
julia> b.^2
2×2 Array{Int64,2}:
  1  4
 16  9
julia> b .= b .+ 2
2×2 MatrixType{Int64,Float64}:
 3  4
 6  5
julia> b.x
2×2 Array{Int64,2}:
 3  4
 6  5
julia> b.+b.x
2×2 Array{Int64,2}:
  6   8
 12  10
```

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -4,8 +4,8 @@ module DiffEqBase
 
 using RecipesBase, SimpleTraits, RecursiveArrayTools, Compat, Juno, LinearMaps
 
-import Base: length, ndims, size, getindex, setindex!, endof, show, print,
-             next, start, done, eltype, eachindex, similar
+import Base: length, size, getindex, setindex!, show, print,
+             next, start, done, similar, indices
 
 import Base: resize!, deleteat!
 
@@ -85,9 +85,13 @@ abstract type AbstractDiffEqInterpolation <: Function end
 abstract type AbstractDEOptions end
 abstract type DECache end
 abstract type DECallback end
-abstract type DEDataArray{T} <: AbstractVector{T} end
 
-export AbstractDiffEqInterpolation, AbstractDEOptions, DECache, DECallback, DEDataArray
+abstract type DEDataArray{T,N} <: AbstractArray{T,N} end
+const DEDataVector{T} = DEDataArray{T,1}
+const DEDataMatrix{T} = DEDataArray{T,2}
+
+export AbstractDiffEqInterpolation, AbstractDEOptions, DECache, DECallback, DEDataArray,
+       DEDataVector, DEDataMatrix
 
 # Integrators
 abstract type DEIntegrator end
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
This looks great. One extra check though: does it pass the data_array_tests.jl in OrdinaryDiffEq.jl?

ChrisRackauckas:
Tests pass and implementation looks good. Will merge when there's confirmation the data array tests have a simple change to `DEDataVector` to pass.

devmotion:
Just saw that I forgot to remove the old exports of `copy_non_array_field`...

ChrisRackauckas:
I wish "bad exports" would actually error, instead of being silent.

ChrisRackauckas:
@aytekinar you might want to make note of this. Your codes should all still work by changing `DEDataArray` to `DEDataVector`. Sorry about that!

aytekinar:
Thank you for informing us, @ChrisRackauckas. We should not have any trouble --- I will look into that 👍

PR #68, opened 2017-08-21
Implemented size method for AffineDiffEqOperators

jagot:
`size` is necessary for F*u. The sizes of the constituent operators
have to agree as well, for the sum to be defined.

NB. To not depend on OrdinaryDiffEq.jl for testing, I implemented a dummy operator consisting only of size information.


ChrisRackauckas:
I'm kind of wondering whether the array operator should be in Base.

jagot:
As in DiffEqBase.jl or Base?

ChrisRackauckas:
DiffEqBase. I think it would be nice to auto-apply it to arrays that are passed into problems. But maybe that's just Base bloat.

PR #77, opened 2017-11-13
Convenience interface for BVProblems

BeastyBlacksmith:


ChrisRackauckas:
Thanks!

BeastyBlacksmith:
you are welcome, I just wondered, if there is a possibility to guarantee that those are not used when doing Shooting

ChrisRackauckas:
It could be structured like the `ODEProblem` to have a trait for whether it's a collocation discretization or not.

PR #82, opened 2018-01-28
Add update warning

Datseris:
(requires new tag release to take effect)

This message is displayed as an info when a user updates `DiffEqBase` (so, *any* package of the ecosystem).

It is displayed only once, during the update, because the script here triggers the "Pkg.build" after the update.

*I am not sure whether this is triggered multiple times if one updates multiple times.* I am sure however that this does not trigger on `using ...`, only updates.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,0 +1,33 @@
+infomsg = """
+~~~ DifferentialEquations.jl *BREAKING* changes  ~~~
+
+We have changed the front-end API on how
+users may define equations of motion and 
+problems, for all problem types that can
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Good stuff! Will merge when tests pass and I see what it looks like.

Datseris:
As a side-remark: `Pkg.clone()` does not trigger the info (because they told me in slack it doesn't trigger Build). So I don't know if travis of DiffEq will show it.

ChrisRackauckas:
That shows really nicely. Thanks! I'll leave it up for probably 3-6 months.

PR #83, opened 2018-02-04
WIP Add TableTraits.jl interface for solutions

davidanthoff:
This moves the iterable tables integration from IterableTables.jl into this repo here.

Tests don't pass right now, I think mainly because I don't understand DiffEqBase well enough and the way these problems are setup doesn't work anymore with the new DiffEqBase release. My hope is that @ChrisRackauckas can point out how to change these in negative time.

A bigger questions is probably whether these tests should depend on ParameterizedFunctions.jl or not. Right now the last test does. Not sure what the best choice is there.

The corresponding PR on the IterableTables.jl side is https://github.com/davidanthoff/IterableTables.jl/pull/72. Ideally we should tag new versions of both packages roughly at the same time with these two PRs merged.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,3 +6,6 @@ Compat 0.19.0
 Requires
 LinearMaps
 FunctionWrappers
+IteratorInterfaceExtensions 0.0.2
+TableTraits 0.0.3
+NamedTuples 4.0.0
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
Yeah, the DiffEqBase tests are a little odd since they need to avoid dependent libraries in tests in order to make bounding possible. I fixed the tests and it works great. Thanks! I'm tagging right now.

davidanthoff:
Great, thanks!

Here is the corresponding IterableTables.jl METADATA PR: https://github.com/JuliaLang/METADATA.jl/pull/13243

PR #84, opened 2018-02-13
Fix set_u0 and add a test

tkf:


ChrisRackauckas:
IMO we should just do it right (it's still undocumented, so this can be changed).

https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/230

Instead of doing a quick fix, let's try to get something along those lines? Either multiple functions with a good naming scheme, or a single function with kwargs?

tkf:
I thought I'd suggest/ask for a more comprehensive solution after this quick fix, so yes, that sounds great.

PR #85, opened 2018-02-15
add `step!(integ, Δt)`

Datseris:


ChrisRackauckas:
@tkf I like this extension better than a separate name, thoughts?

ChrisRackauckas:
Just make that `dt` instead of `Δt` to match the naming convention. It doesn't actually effect usage though.

Datseris:
Are the individual `step!` for each package defined within the package? Like e.g. `OrdinaryDiffEq`? I was expecting to see a simple `function step! end` in DiffEqBase. 

We have to be careful whether we need to add `import DiffEqBase:  step!`

ChrisRackauckas:
Yes, the `step!` implementations are in the packages, but this doesn't need to know about them since it will just be calling the overloads internally. So this will generically work even though it's defined before the actual implementations.

Datseris:
I like docstrings. I added one! Is it correct?

ChrisRackauckas:
Put it on the generic definition at the top.

ChrisRackauckas:
Seems great to me.

Though right after merging it, maybe there should be an arg `stop_at_t = false` to add a `tstop` at `t+dt` (`add_tstop!(integ,t+dt)`), and it should check `>=` instead of `>` to be more robust. @Datseris would you like to PR that to finish it up?

Datseris:
Hi, thanks for merging it. 

I won't have time to do the change you suggested today, but if it is a small 4 liner, maybe you can just go ahead and do it?

PR #86, opened 2018-02-17
Tweak and fix step!, add set_t!

tkf:
closes https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/253

This PR bundles some related changes:

* Tweak `step!(integrator, dt)` to return the actual advanced step.
* Add `set_t!(integrator, t)`
* A few tests for integrator interface.
* Fixed a typo/bug in `step!`

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -229,7 +229,7 @@ export resize!,deleteat!,addat!,get_tmp_cache,full_cache,user_cache,u_cache,du_c
        terminate!,
        add_tstop!,add_saveat!,set_abstol!,
        set_reltol!,get_du,get_dt,get_proposed_dt,set_proposed_dt!,u_modified!,
-       savevalues!,reinit!, auto_dt_reset!, set_t!
+       savevalues!,reinit!, auto_dt_reset!, set_t!, set_u!
 
 export numargs, @def
 
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -229,7 +229,7 @@ export resize!,deleteat!,addat!,get_tmp_cache,full_cache,user_cache,u_cache,du_c
        terminate!,
        add_tstop!,add_saveat!,set_abstol!,
        set_reltol!,get_du,get_dt,get_proposed_dt,set_proposed_dt!,u_modified!,
-       savevalues!,reinit!, auto_dt_reset!
+       savevalues!,reinit!, auto_dt_reset!, set_t!
 
 export numargs, @def
 
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -44,6 +44,14 @@ Set current time point of the `integrator` to `t`.
 set_t!(integrator::DEIntegrator, t::Real) =
     error("set_t!: method has not been implemented for the integrator")
 
+"""
+    set_u!(integrator::DEIntegrator, u)
+
+Set current state of the `integrator` to `u`.
+"""
+set_u!(integrator::DEIntegrator, u) =
+    error("set_u!: method has not been implemented for the integrator")
+
 ### Addat isn't a real thing. Let's make it a real thing Gretchen
 
 function addat!(a::AbstractArray,idxs,val=zeros(length(idxs)))
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -3,7 +3,8 @@
 Perform one (successful) step on the integrator.
 
 Alternative, if a `dt` is given, then `step!` the integrator until
-there is a temporal difference `≥ dt` in `integ.t`.
+there is a temporal difference `≥ dt` in `integ.t`.  It returns the
+actual temporal difference advanced by the integrator.
 """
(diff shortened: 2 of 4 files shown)


tkf:
I also wonder maybe the method has to step exactly `dt` by default, since `step!(integrator, dt)` advancing more than `dt` is kind of surprising.  To be more precise, how about something like:

>     step!(integrator::DEIntegrator [, dt [, exact = true]])
> Perform one (successful) step on the `integrator`.
>
> When the second argument `dt` is given, advance the `integrator` by
> `dt`.  If the third argument `exact` is `false`, the `integrator` may
> advance more than `dt`.  Actual advanced temporal difference is
> returned.

tkf:
Also, should `step!` check if `dt * tdir > 0` and otherwise error out?  Or should we assert that `dt > 0` and advance `dt * tdir`?

ChrisRackauckas:
>I also wonder if the method has to step exactly dt by default, since step!(integrator, dt) advancing more than dt is kind of surprising.

But not all methods can do that though (but the vast majority can). But sure, I can see it making more sense to assume it as exact.

>Also, should step! check if dt * tdir > 0 and otherwise error out? Or should we assert that dt > 0 and advance dt * tdir?

We do need to take direction into account. For `dt` choice I always allow that to be absolute value since requiring it to be negative for negative direction is easy to get wrong. But yes, currently the check is incorrect if we allow negative tdir.

tkf:
I see.  If not all methods can do it, I guess current semantics is fine, since `step!` is a low-level API anyway.  But at the same time, I also think forcing users to be aware of the limitation and let them write the third argument explicitly makes sense.  I don't have a very strong opinion on this though.

Re: tdir, do all integraters have tdir?  If so, I can just add a one-linear check.

tkf:
I saw `integrator.tdir` in the `@recipe` so I just assume it's always defined.

ChrisRackauckas:
>I saw integrator.tdir in the @recipe so I just assume it's always defined.

Alright, let's formalize it then. Maybe that should get a function call? Either way, it should get added to the docs as part of the interface.

tkf:
Why don't we open an issue about it and pull this patch?  I mean, many fields of abstract type `DEIntegrator` (including .tdir) are touched in integrator_interface.jl already so I don't think it is urgent problem.  I thought it was fine because those are interface internal to JuliaDiffEq ecosystem.  I suggested `set_t!` since I thought it was a popular attribute you might want to change, even from packages outside JuliaDiffEq (i.e., I want to use it :wink:).  Also, having `set_t!` means that you can change `t` without knowing how it is implemented.

But maybe we should just document `integrator.t = new value` is fine?  Now that `setproperty!` is coming, we can inject error check for the integrators in which this operation is unsafe, right?

ChrisRackauckas:
>But maybe we should just document integrator.t = new value is fine?

Nope, `set_t!`, but I'm not sure we should give a default implementation since that's not always correct. Methods that extrapolate for the initial value of the Newton iterations (`Trapezoid()`) and multistep methods (`CVODE_BDF()`) need to take extra care (`Trapezoid()` just needs to flip an argument to tell it to not extrapolate the next step, `CVODE_BDF()` needs to `reinit!`). So let's add that without a default implementation and then set up some packages with it. We can just error on the hard cases for now (in OrdinaryDiffEq, that would just be if `alg_extrapolates(alg) || !alg_dtchangable(alg) == true`)

ChrisRackauckas:
Let me clarify something. In general setting `t` and `u` will break things. That's mentioned in the docs. 

http://docs.juliadiffeq.org/latest/basics/integrator.html#Note-about-mutating-1

Callbacks make sure this is done safely though by doing the appropriate handling. In the integrator interface, the user would have to know the right way to handle it. Or we can have `set_t!` and `set_u!` be safe ways of doing it, which sounds like a good option (maybe `set_ut!` too?)

ChrisRackauckas:
Make the `set_t!` be a placeholder (or a separate PR) and this is good to go.

tkf:
Thanks for the clarification.  I remembered that `u` shouldn't be changed but not `t`.

OK. Rebased against master, turned `set_t!` to the not-implemented error, and also added `set_u!`.

tkf:
BTW, not sure if I understand why you want `set_ut!`.  If both `set_t!` and `set_u!` are inlined, calling both of them would be equivalent to `set_ut!`, right?  Or you want to "fuse" the error check or other kind of processing?

ChrisRackauckas:
The `set` functions may need to call `reinit`. For example, this would be required for multistep methods like Sundials. In that case, I'm not sure that LLVM will know that it only needs to do it once. `reinit` is cheap, but I wouldn't want to double up too much.

tkf:
I see.  Thanks.  It sounds very hard for LLVM/Julia to optimize that out.

So the PR is good to go?  Looks like Travis is OK with it.

ChrisRackauckas:
Thanks! The new `set` stuff should get added to the docs (and implementations of course).

tkf:
Thank you!

Datseris:
nice

PR #87, opened 2018-02-17
Add remake() for reconstructing immutable structs

tkf:
closes https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/230

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -67,3 +67,99 @@ function check_keywords(alg, kwargs, warnlist)
 end
 warn_compat() =
     warn("Please see http://docs.juliadiffeq.org/latest/basics/compatibility_chart.html")
+
+
+"""
+    @add_kwonly function_definition
+
+Define keyword-only version of the `function_definition`.
+
+    @add_kwonly function f(a, b; c=1, d=2)
+        ...
+    end
+
+expands to:
+
+    function f(x; y=1)
+        ...
+    end
+    function f(; x = error("No argument x"), y=1)
+        ...
+    end
+"""
+macro add_kwonly(ex)
+  esc(add_kwonly(ex))
+end
+
+add_kwonly(ex::Expr) = add_kwonly(Val{ex.head}, ex)
+
+function add_kwonly(::Type{<: Val}, ex)
+  error("add_only does not work with expression $(ex.head)")
+end
+
+function add_kwonly(::Union{Type{Val{:function}},
+                            Type{Val{:(=)}}}, ex::Expr)
+  body = ex.args[2:end]  # function body
+  default_call = ex.args[1]  # e.g., :(f(a, b=2; c=3))
+  kwonly_call = add_kwonly(default_call)
+  if kwonly_call === nothing
+    return ex
+  end
+
+  return quote
+    begin
+      $ex
+      $(Expr(ex.head, kwonly_call, body...))
+    end
+  end
+end
+
+function add_kwonly(::Type{Val{:where}}, ex::Expr)
+  default_call = ex.args[1]
+  rest = ex.args[2:end]
+  kwonly_call = add_kwonly(default_call)
+  if kwonly_call === nothing
+    return nothing
+  end
+  return Expr(:where, kwonly_call, rest...)
+end
+
+function add_kwonly(::Type{Val{:call}}, default_call::Expr)
+  # default_call is, e.g., :(f(a, b=2; c=3))
+  funcname = default_call.args[1]  # e.g., :f
+  required = []  # required positional arguments; e.g., [:a]
+  optional = []  # optional positional arguments; e.g., [:(b=2)]
+  default_kwargs = []
+  for arg in default_call.args[2:end]
+    if isa(arg, Symbol)
+      push!(required, arg)
+    elseif arg.head == :(::)
+      push!(required, arg)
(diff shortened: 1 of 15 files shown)


src/utils.jl
```diff
@@ -67,3 +67,119 @@ function check_keywords(alg, kwargs, warnlist)
 end
 warn_compat() =
     warn("Please see http://docs.juliadiffeq.org/latest/basics/compatibility_chart.html")
+
+
+"""
+    @add_kwonly function_definition
```
ChrisRackauckas: I feel like this could even get its own package. That's quite a handy macro.

src/utils.jl
```diff
@@ -67,3 +67,119 @@ function check_keywords(alg, kwargs, warnlist)
...
+
+struct_as_dict(st) = [(n => getfield(st, n)) for n in fieldnames(typeof(st))]
+
+"""
+    remake(thing; <keyword arguments>)
+
+Re-construct `thing` with new field values specified by the keyword
+arguments.
+"""
+function remake(thing; kwargs...)
```
ChrisRackauckas: Does this infer well?

src/utils.jl
```diff
@@ -163,3 +163,23 @@ function add_kwonly(::Type{Val{:call}}, default_call::Expr)
...
+  T = parameterless_type(thing)
+  constructor = if method_exists(T, ())
+    # This path is required for, e.g., NoiseProblem
+    T
+  else
+    # Assume that T wants isinplace
+    T{isinplace(thing)}
```
tkf: I don't like that it assumes the constructor has the `iip` type parameter but it should be fine as long as `remake` is used for types from DiffEqBase.  Note that I can't dispatch on `DEProblem` to avoid this since using it means that `remake` won't work with `SplitFunction` etc.

test/remake_tests.jl
```diff
@@ -0,0 +1,37 @@
...
+# Create a ODEProblem and test set_u0:
+prob1 = SplitODEProblem(f,f,u0,tspan,Dict(); mass_matrix=eye(length(u0)))
+prob2 = remake(prob1; u0 = prob1.u0 + 1)
+@test prob1.f === prob2.f
+@test prob1.p === prob2.p
+@test prob1.u0 + 1 ≈ prob2.u0
+@test prob1.tspan == prob2.tspan
+@test prob1.jac_prototype === prob2.jac_prototype
+@test prob1.callback === prob2.callback
+@test prob1.mass_matrix === prob2.mass_matrix
```
ChrisRackauckas: Probably want to add some inference tests. Some might fail, in which case just `@test_broken` them so we can keep track of it for v0.7.

ChrisRackauckas:
That's a very clean solution. You should consider making `Kwonly.jl` for that macro because it's quite nice and I could see it being used elsewhere.

src/utils.jl
tkf: Oops. I just comment it on https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/87#pullrequestreview-97369157

tkf:
Thanks :)  Well, it needs a bit of help from `remake` to "inject" isinplace type parameter so it's not quite a general solution (yet?).  But I agree that making a package that does this makes sense.  Until then, it can live here, I suppose.

ChrisRackauckas:
Sounds good. Then just some inference checks and it'll be done. If inference is really bad then it may need to be an `@generated` function?

ChrisRackauckas:
>BTW, I think current set_u0 is not working. 

Yes, remove it. It's undocumented and so it's fine. Make sure we ping whoever added it. I think it was @BeastyBlacksmith ? Check blame.

tkf:
It was @Datseris.

@Datseris We are removing `set_u0` Use `remake` after this PR is merged.

Datseris:
thanks for letting me know. Is there any documentation string I can read? I am super stressed at the moment to read all git diff.

tkf:
`set_u0(prob, u)` is just `remake(prob; u0=u)` after this PR.

Datseris:
okay great

tkf:
@ChrisRackauckas So I wrote the generated version (see #89) but `@inferred` raises an error.  It looks like the constructor itself cannot be `@inferred`ed prior to this PR (see #88).  Am I using `@inferred` correctly?

tkf:
It looks like this is it: https://github.com/JuliaLang/julia/issues/25918

Example:

```julia
type Spam{T}
    Spam(x::T) where T = new{T}()
    Spam(; x::T = nothing) where T = new{T}()
end

using Base.Test

macro test_nothrow(ex)
    quote
        @test begin
            $(esc(ex))
            true
        end
    end
end

@testset "Spam" begin
    @test_nothrow @inferred Spam(nothing)
    @test_nothrow @inferred Spam()  # it fails
    # @test_nothrow @inferred Spam(x="x")
end
```

ChrisRackauckas:
Ugh, keyword arguments on v0.6.2 don't even infer the default usage? They used to. Okay then whatever. Mark as `@test_broken` and we'll see what happens with v0.7.

ChrisRackauckas:
Alright, then let's do this for now and leave fixing the inference for v0.7 when kwargs are fixed.

tkf:
Thank you for the merge!

PR #88, opened 2018-02-17
Test problem constructors against @inferred

tkf:
This is just to demonstrate that the tests fail at the moment.

### DIFF
--- a/test/problem_creation_tests.jl
+++ b/test/problem_creation_tests.jl
@@ -15,6 +15,9 @@ prob = ODEProblem{true}(f,u0,tspan)
 prob = ODEProblem{false}(f,u0,tspan)
 @test isinplace(prob) == false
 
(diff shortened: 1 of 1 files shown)


test/problem_creation_tests.jl
```diff
@@ -15,6 +15,9 @@ prob = ODEProblem{true}(f,u0,tspan)
 prob = ODEProblem{false}(f,u0,tspan)
 @test isinplace(prob) == false
 
+@test_nothrow @inferred ODEProblem{true}(f,u0,tspan)
```
ChrisRackauckas: Just use `@test_broken`?

tkf: Done in the rebased version.

ChrisRackauckas:
Thanks!

PR #89, opened 2018-02-17
@generated remake

tkf:
see #87


tkf:
Now that #87 is merged, I'm closing it. We can't know if it is required until tests in #88 pass.

PR #93, opened 2018-02-22
Use u_modified! in INITIALIZE_DEFAULT

tkf:
https://github.com/JuliaDiffEq/DiffEqDocs.jl/issues/81#issuecomment-367146902

I should say I didn't test anything :)


ChrisRackauckas:
Right now DiffEqCallbacks fails because of a change in NLsolve.jl which merged and I think needs a tag.

PR #97, opened 2018-03-28
Add detection in recipe for log xscale and space t appropriately

### LINKED ISSUE #95: Use logspace instead of linspace in plotting recipe when appropriate
If `xscale` is set to e.g. `:log10`, `logspace` should be used instead of
`linspace`. This is necessary for visualizing the example from the DAE tutorial
well (https://github.com/JuliaDiffEq/DiffEqDocs.jl/pull/113).

non-Jedi:
Closes #95

### DIFF
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -72,15 +72,23 @@ DEFAULT_PLOT_FUNC(x,y,z) = (x,y,z) # For v0.5.2 bug
     end_idx = findlast(x -> x<=tspan[end],sol.t)
   end
 
+  # determine type of spacing for plott
+  tscale = get(plotattributes, :xscale, :identity)
+  densetspacer = if tscale in [:ln, :log10, :log2]
+    (start, stop, n) -> logspace(log10(start), log10(stop), n)
+  else
+    linspace
+  end
+
   if denseplot
     # Generate the points from the plot from dense function
     if tspan == nothing && !(typeof(sol) <: AbstractAnalyticalSolution)
-      plott = collect(linspace(sol.t[start_idx],sol.t[end_idx],plotdensity))
+      plott = collect(densetspacer(sol.t[start_idx],sol.t[end_idx],plotdensity))
     elseif typeof(sol) <: AbstractAnalyticalSolution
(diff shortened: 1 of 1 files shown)


src/solutions/solution_interface.jl
```diff
@@ -72,15 +72,23 @@ DEFAULT_PLOT_FUNC(x,y,z) = (x,y,z) # For v0.5.2 bug
     end_idx = findlast(x -> x<=tspan[end],sol.t)
   end
 
+  # determine type of spacing for plott
+  tscale = get(plotattributes, :xscale, :identity)
+  densetspacer = if tscale in [:ln, :log10, :log2]
+    (start, stop, n) -> logspace(log10(start), log10(stop), n)
```
ChrisRackauckas: always `log10` here?

non-Jedi: Since logspace is doing 10^x for evenly spaced x. I've got it in my head that this gives equivalent output to if logspace was doing e^x and we used `log` instead, but I haven't actually proven that on paper. Hold on.

non-Jedi: Don't have pen and paper right now but:

```
julia> bounds = (1,10)
(1, 10)

julia> bounds10 = map(log10, bounds)
(0.0, 1.0)

julia> boundse = map(log, bounds)
(0.0, 2.302585092994046)

julia> t10 = linspace(bounds10..., 10)
0.0:0.1111111111111111:1.0

julia> te = linspace(boundse..., 10)
0.0:0.2558427881104495:2.302585092994046

julia> log10space = map(x -> 10^x, t10)
10-element Array{Float64,1}:
  1.0    
  1.29155
  1.6681 
  2.15443
  2.78256
  3.59381
  4.64159
  5.99484
  7.74264
 10.0    

julia> logespace = map(exp, te)
10-element Array{Float64,1}:
  1.0    
  1.29155
  1.6681 
  2.15443
  2.78256
  3.59381
  4.64159
  5.99484
  7.74264
 10.0
```

ChrisRackauckas: Sweet

PR #98, opened 2018-04-06
Initiate Julia 1.0 support

tkf:
https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/271

Incidentally, I was playing with DiffEqBase to make it work with nightly :)

It turned out just a few tweaks to the code makes the unit test happy.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -40,12 +40,7 @@ end
 
 using Compat.TypeUtils: typename
 
-if :wrapper in fieldnames(TypeName)
-    parameterless_type(T::Type) = typename(T).wrapper
-else
-    parameterless_type(T::Type) = typename(T).primary
-end
-
+parameterless_type(T::Type) = typename(T).wrapper
 parameterless_type(x) = parameterless_type(typeof(x))
 
 # support functions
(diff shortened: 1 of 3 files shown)


src/utils.jl
```diff
@@ -40,12 +40,7 @@ end
 
 using Compat.TypeUtils: typename
 
-if :wrapper in fieldnames(TypeName)
```
tkf: It turned out `Base.TypeName` is gone (although `Core.TypeName` is still there). I assumed that `typename(T).primary` is for Julia < 0.6; if so, I think it's simplest if we remove the `else` block. Is it correct? Otherwise, I suppose adding `using Core: TypeName` would work.

test/tabletraits_tests.jl
```diff
@@ -8,15 +8,15 @@ sol =solve(prob,InternalEuler.FwdEulerAlg();dt=1//2^(4))
 array = collect(getiterator(sol))
 
 @test length(array) == 17
-@test fieldnames(array[1]) == [:timestamp, :value]
```
tkf: fieldnames returns a tuple in 0.7 so I just wrap it with collect

test/runtests.jl
```diff
@@ -5,7 +5,9 @@ tic()
 @time @testset "Number of Parameters Calculation" begin include("numargs_test.jl") end
 @time @testset "Data Arrays" begin include("data_array_tests.jl") end
 @time @testset "Extended Functions" begin include("extended_function_tests.jl") end
-@time @testset "DiffEqFunctions" begin include("diffeqfunction_tests.jl") end
+if VERSION < v"0.7-"
+  @time @testset "DiffEqFunctions" begin include("diffeqfunction_tests.jl") end
```
tkf: https://github.com/yuyichao/FunctionWrappers.jl/issues/6

PR #99, opened 2018-04-13
Add functions to check return code

tkf:
https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/276


src/integrator_interface.jl
```diff
@@ -88,6 +88,79 @@ function Base.show(io::IO, m::MIME"text/plain", A::DEIntegrator)
...
 
+### Error check (retcode)
+
+"""
+    check_error(integrator)
+
+Check state of `integrator` and return one of the
+[Return Codes](http://docs.juliadiffeq.org/latest/basics/solution.html#Return-Codes-(RetCodes)-1)
+"""
+function check_error(integrator::DEIntegrator)
```
tkf: Should it be `::Union{AbstractODEIntegrator, AbstractSDEIntegrator}`?  `AbstractRODEIntegrator` too?

src/integrator_interface.jl
```diff
@@ -88,6 +88,79 @@ function Base.show(io::IO, m::MIME"text/plain", A::DEIntegrator)
...
+end
+
+@inline function next(integrator::DEIntegrator,state)
+  state += 1
+  step!(integrator) # Iter updated in the step! header
+  # Next is callbacks -> iterator  -> top
+  integrator,state
+end
+
+@inline done(integrator::DEIntegrator, _) = check_error!(integrator) != :Success
```
ChrisRackauckas: I think this will loop forever since it won't know when it hits the last tstop?

src/integrator_interface.jl
ChrisRackauckas: That's an interesting question. What should Sundials et al. do? They do their own checks right now.

src/integrator_interface.jl
tkf: Thanks. Fixed.

tkf:
I'm locally running tests with https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/316
I'll let you know if it passes.

src/integrator_interface.jl
```diff
@@ -88,6 +88,94 @@ function Base.show(io::IO, m::MIME"text/plain", A::DEIntegrator)
...
+  if integrator.last_stepfail && !integrator.opts.adaptive
+    if integrator.opts.verbose
+      warn("Newton steps could not converge and algorithm is not adaptive. Use a lower dt.")
+    end
+    return :ConvergenceFailure
+  end
+  return :Success
+end
+
+function postamble! end
```
tkf: FYI I need to turn `postamble!` into an internal interface function to make `check_error!` and iterator work.

ChrisRackauckas: yeah I saw that. That's fine.

src/integrator_interface.jl
tkf: Do they have their own integrator type? If so, isn't it fine, since they are going to override the default?

tkf:
So, originally, `@ode_exit_conditions` had checks for

* C1 `integrator.iter > integrator.opts.maxiters`
* C2 `!integrator.opts.force_dtmin && integrator.opts.adaptive && abs(integrator.dt) <= abs(integrator.opts.dtmin)`
* C3 `integrator.opts.unstable_check(integrator.dt,integrator.u,integrator.p,integrator.t)`
* C4 `integrator.last_stepfail && !integrator.opts.adaptive`

and `done(integrator::ODEIntegrator,state)` for

* I1 `integrator.iter > integrator.opts.maxiters`
* I2 `integrator.opts.unstable_check(integrator.dt,integrator.u,integrator.p,integrator.t)`
* I3 `!integrator.opts.force_dtmin && integrator.opts.adaptive && abs(integrator.dt) <= abs(integrator.opts.dtmin)`
* I4 `isempty(integrator.opts.tstops)`
* I5 `integrator.just_hit_tstop`

The correspondence is that:

* I1 = C1
* I2 = **C3**
* I3 = **C2**
* (new) = C4
* I4, I5 (still in `done` so it's fine)

Is this OK change for `done`?

src/integrator_interface.jl
tkf: So they have

```julia
function interpret_sundials_retcode(flag)
  flag >= 0 && return :Success
  flag == -1 && return :maxiterss
  (flag == -2 || flag == -3) && return :Unstable
  flag == -4 && return :ConvergenceFailure
  return :Failure
end
```

https://github.com/JuliaDiffEq/Sundials.jl/blob/c7e216b488f0028ee06c5a1db72b2023b0ebd1f6/src/common_interface/solve.jl#L767

and do `solution_new_retcode(integrator.sol,interpret_sundials_retcode(integrator.flag))` at the end of `solve!`

tkf: A dumb way to solve it would be to define something like `check_error_default` in DiffEqBase and do `check_error(integrator::ODEIntegrator) = check_error_default(integrator::ODEIntegrator)` in OrdinaryDiffEq.jl. I can't think of any elegant solution....

> What should Sundials et al. do?

Something like

```julia
check_error(integrator::AbstractSundialsIntegrator) =
  interpret_sundials_retcode(integrator.flag)
```

PR #100, opened 2018-04-14
Check sol.retcode in done

tkf:
Once https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/319 and https://github.com/JuliaDiffEq/StochasticDiffEq.jl/pull/74 are merged, there is no need to call `check_error!` in `done` since it is already called in `step!` (via `next`).

If `check_error!` (or in particular `postamble!`) is idempotent then the old code should be OK.  If so, maybe it's better to not pull this PR for a while, until all DiffEq downstreams call `check_error!` in `step!`.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -159,7 +159,7 @@ end
 end
 
 @inline function done(integrator::DEIntegrator, _)
-  if check_error!(integrator) != :Success
+  if ! (integrator.sol.retcode in (:Default, :Success))
     return true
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
>Once JuliaDiffEq/OrdinaryDiffEq.jl#319 and JuliaDiffEq/StochasticDiffEq.jl#74 are merged, there is no need to call check_error! in done since it is already called in step! (via next).

This is missing Sundials.jl though.

PR #101, opened 2018-04-14
Check sol.retcode in step!(integrator, dt)

tkf:
In `step!(integrator, dt)`, `integrator.sol.retcode` has to be checked since otherwise it could enter into an infinite loop.

See: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/319 and https://github.com/JuliaDiffEq/StochasticDiffEq.jl/pull/74

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -315,6 +315,7 @@ function step!(integ::DEIntegrator, dt::Real, stop_at_tdt = false)
     stop_at_tdt && add_tstop!(integ,next_t)
     while integ.t < next_t
         step!(integ)
+        integ.sol.retcode in (:Default, :Success) || break
     end
     return integ.t - t
 end
--- a/test/integrator_tests.jl
+++ b/test/integrator_tests.jl
@@ -1,10 +1,15 @@
+mutable struct DummySolution
+  retcode
+end
+
 mutable struct DummyIntegrator <: DEIntegrator
   t
   dt
   tdir
   tstops
+  sol
 
-  DummyIntegrator() = new(0,1,1,[])
+  DummyIntegrator() = new(0,1,1,[],DummySolution(:Default))
 end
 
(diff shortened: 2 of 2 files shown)


tkf:
Since callers of `step!` also have to check the return code, I wonder if it should return something like Haskell's Either.  (I'm sure I saw a Julia package to do it on GitHub and a JuliaCon talk on YouTube, but I don't remember where it is...)

Or maybe we just should return a tuple `(dt, retcode)`.

ChrisRackauckas:
Whatever it is, it should be consistent. I kind of think `nothing` is fine, since it's all just updating the integrator.

tkf:
By consistent, you mean `step!(integrator)` and `step!(integrator, dt)` have to return the same type?  I think if input type/arity of a function is different it is OK to return different type, since the returned type can be statically inferred.

But I'm not against for `step!(integrator, dt)` to return `nothing`.  Caller can calculate the actual advanced step by comparing `integrator.t` before and after.  As it is a bit cumbersome, I sent a PR to return the advanced step in #86.

Can Julia eliminate unused returned object?  If so, I think it would be better if `step!` always return a named tuple with `retcode` and `dt` (and some more handy information, if any).

ChrisRackauckas:
>Can Julia eliminate unused returned object? If so, I think it would be better if step! always return a named tuple with retcode and dt (and some more handy information, if any).

I think so? But I think I've seen a case where it hurt? Time it.

tkf:
Do you want an API change in this PR?  It is supposed to be a simple bug fix PR.  I suggest to open a new issue for this.

If we are to introduce a new API (return a info struct/tuple from `step(integrator, dt)` *and* `step(integrator)`), it will require changes across downstreams.

ChrisRackauckas:
>it will require changes across downstreams.

Is it actually used downstream? I don't think it is.

Yeah, new PR.

PR #103, opened 2018-04-14
Return nothing from step!(integrator, dt)

tkf:
closes https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/280


ChrisRackauckas:
MWTP

PR #105, opened 2018-04-28
Remove undefined exports

tkf:
Here is a way to fix UndefVarError quirks (see also https://github.com/JuliaDiffEq/diffeqpy/pull/24#issuecomment-384885291).  I was half unsure if it's just end up being a yet another nitpick but I found that `include("alg_traits.jl")` is missing while making this PR.  I'm not sure if it is intentionally removed.  But if the intention was to add `include("alg_traits.jl")`, I think it actually make sense to include such test, at least in `DiffEqBase`, to make sure that the interface functions are correctly exported.

If we are to apply this to other packages, each package "just" have to add a one-linear test. For example, for `OrdinaryDiffEq`:

```julia
@test DiffEqBase.undefined_exports(OrdinaryDiffEq) == []
```

Fixing a package (say `OrdinaryDiffEq`) is less cumbersome, e.g., if you use `sed`:

```sh
sed -E -i "$(julia -e 'using DiffEqBase, OrdinaryDiffEq; println("s/\\b(", join(DiffEqBase.undefined_exports(OrdinaryDiffEq), "|"), ")\\b(, *)?//g")')" src/**/*.jl
```

Then just do `git diff` to see what's removed and fix some syntax errors.

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -208,6 +208,7 @@ include("data_array.jl")
 include("internal_euler.jl")
 include("juno_rendering.jl")
 include("tabletraits.jl")
+include("alg_traits.jl")
 
 export NSODEFunction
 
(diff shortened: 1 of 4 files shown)


tkf:
Note that after this PR is merged, `using OrdinaryDiffEq; using DiffEqBase` would emit a warning since `isautodifferentiable` is shadowed.  We'd need `import DiffEqBase: isautodifferentiable` there.

PR #108, opened 2018-06-12
Update arguments style for `DiffEqFunction`

MSeeker1340:
Updates the argument style from the old `(t,u[,du])` to `([du],u,p,t)`.

The section for `NSODEFunction` still seems problematic for me. First of all why create a dummy type `NSODEFunction` when in fact we're just returning a `DiffEqFunction`? The types for `FunctionWrapper` also don't take into account units (the type of `du` might be different from `u`).

Is it really necessary to use `FunctionWrapper` after all?

### DIFF
--- a/src/diffeqfunction.jl
+++ b/src/diffeqfunction.jl
@@ -39,12 +39,12 @@ end
 
 struct NSODEFunction{iip} <: AbstractDiffEqFunction{iip} end
 
-function NSODEFunction(f,t,u;kwargs...)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
>Is it really necessary to use FunctionWrapper after all?

I'm not sure, but we should probably have an easy documented way to make it all compilation-free:

https://stackoverflow.com/questions/47501844/julia-differentialequations-jl-speed/47507933#47507933

PR #110, opened 2018-06-23
Bump REQUIRE to 0.7-alpha

YingboMa:


src/solutions/solution_interface.jl
```diff
@@ -398,8 +398,7 @@ function solplot_vecs_and_labels(dims,vars,plot_timeseries,plott,sol,plot_analyt
 
     f = x[1]
 
-    if VERSION < v"0.6-" && length(tmp) == 3
-      # Work around bug
+    if       # Work around bug
```
YingboMa: bad bot

src/solutions/solution_interface.jl
```diff
@@ -413,7 +412,7 @@ function solplot_vecs_and_labels(dims,vars,plot_timeseries,plott,sol,plot_analyt
       end
       tmp = res
     else
-      tmp = f.(tmp...)
+      tmp = f.(tmp...,)
```
ChrisRackauckas: Is this correct? @keno

Keno: No, it's not. I'll make a note of it, but in the future ideally file issues. Otherwise these things get lost.

src/solutions/solution_interface.jl
Keno: Indeed. Though the `bad bot` only works with "Changes requested", which I now realize doesn't work if you're the PR author, so I'm gonna have to figure out what to do about that. In the mean time, manual issues on Deprecations.jl also work.

PR #114, opened 2018-07-04
WIP: Implement a high level init/solve for ecosystem wide warnings and input functions

ChrisRackauckas:
This uses @tkf's suggestion in https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/251 to easily allow for function inputs of u0 and tspan. Additionally other tspan inputs are allowed, and tspan from kwargs is an accepted method (@Datseris). This will also be used to solve https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/275 

For this to work well, `remake` needs to be fast and inferred and I don't think that's true right now.

### DIFF
--- a/src/problems/analytical_problems.jl
+++ b/src/problems/analytical_problems.jl
@@ -1,7 +1,7 @@
 struct AnalyticalProblem{uType,tType,isinplace,P,F,C} <: AbstractAnalyticalProblem{uType,tType,isinplace}
   f::F
   u0::uType
-  tspan::Tuple{tType,tType}
+  tspan::tType
   p::P
   callback::C
   @add_kwonly function AnalyticalProblem{iip}(f,u0,tspan,p=nothing;
--- a/src/problems/bvp_problems.jl
+++ b/src/problems/bvp_problems.jl
@@ -4,7 +4,7 @@ struct BVProblem{uType,tType,isinplace,P,J,F,bF,PT,CB,MM} <: AbstractBVProblem{u
     f::F
     bc::bF
     u0::uType
-    tspan::Tuple{tType,tType}
+    tspan::tType
     p::P
     jac_prototype::J
     problem_type::PT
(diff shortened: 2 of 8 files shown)


Datseris:
Very nice ! To summarize for anyone else, this PR (to the best of my knowledge) implements the ideas almost exactly from [this post](https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/251#issuecomment-366519924)\

@ChrisRackauckas do you want me to help review or something? 

I saw this:
```
function get_concrete_problem(prob,kwargs)
+  if typeof(prob.tspan) <: Function
+    _tspan = prob.u0(prob.p,prob.tspan[1])
```
Doesn't `prob.u0` give the state? maybe you wanted to type `prob.tspan(p)` instead?

ChrisRackauckas:
This is WIP

ChrisRackauckas:
This has an annoying breaking change that `tType` in the problem needs to be the full tuple instead of the element type. But this only effects the solvers and not anything else.

ChrisRackauckas:
>I would make it a functor and have the counter as an extra field, (and also store all vectors as fields).

Can you make that functor a type that is `<:Function`? Wrap it in an anonymous function? There are lots of workarounds here. I agree it's less than ideal, but I think this is the best way to do it since it's not obvious what the other inputs really are (they don't have to be numbers or `AbstractArrays`, just things with arithmetic defined on them and possibly broadcast).

Everything is just too generic :P. I am going to merge and then when updating the solvers I'll set this all up with each one, and likely will have to keep tweaking it. What isn't handled though is Distribution input. That will need some fancy Requires.jl tricks.

Datseris:
> Can you make that functor a type that is `<:Function`? Wrap it in an anonymous function? There are lots of workarounds here. 

Yeap this is fine, I just didn't think of it, that it is possible. You should definitely mention both approaches in the documentation. I am an intermediate level user so I can imagine beginners may not immediatelly think these approaches.

The current setup is a good plan, I was just being the devil's advocate  :D

PR #115, opened 2018-07-08
0.7 compatibility

YingboMa:
This PR aims to fix a test failure in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/392, i.e.

```julia
  prob = DiscreteProblem(0.5,(0.0,1.0))
  sol = solve(prob,FunctionMap())
```

It also fixs some depwarns by `LinearAlgebra` and `Statistics`.

### DIFF
--- a/src/solutions/ode_solutions.jl
+++ b/src/solutions/ode_solutions.jl
@@ -37,7 +37,7 @@ function build_solution(
     f = prob.f
   end
 
-  if !(typeof(prob)<:DiscreteProblem) && has_analytic(f)
+  if has_analytic(f)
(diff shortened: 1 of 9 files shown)


src/solutions/ode_solutions.jl
```diff
@@ -37,7 +37,7 @@ function build_solution(
     f = prob.f
   end
 
-  if has_analytic(f)
+  if !(typeof(prob)<:DiscreteProblem) && has_analytic(f)
```
ChrisRackauckas: this seems like a cop out way of doing it instead of setting it up so they can have an analytical solution.

PR #116, opened 2018-07-10
Refactor `SplitFunction`

YingboMa:


ChrisRackauckas:
I changed it to forward `f1` instead of `f2` since IMEX use the Jacobian only of `f1`.

PR #119, opened 2018-07-13
Refactor `DynamicalODEProblem`, `SplitFunction` and `SecondOrderODEProblem`

YingboMa:


src/solutions/ode_solutions.jl
```diff
@@ -62,15 +62,11 @@ end
 
 function calculate_solution_errors!(sol::AbstractODESolution;fill_uanalytic=true,timeseries_errors=true,dense_errors=true)
 
-  if typeof(sol.prob.f) <: Tuple
```
ChrisRackauckas: oh man, these hacks are finally gone. Thank god.

src/diffeqfunction.jl
```diff
@@ -114,18 +143,25 @@ has_invW_t(f::ODEFunction) = f.invW_t != nothing
...
+has_tgrad(f::DynamicalODEFunction) = f.f1.tgrad != nothing
+has_invW(f::DynamicalODEFunction) = f.f1.invW != nothing
+has_invW_t(f::DynamicalODEFunction) = f.f1.invW_t != nothing
+has_paramjac(f::DynamicalODEFunction) = f.f1.paramjac != nothing
+
 has_analytic(f::DiscreteFunction) = f.analytic != nothing
 
 ######### Compatibility Constructor from Tratis
 
-ODEFunction(f::T) where T = return T<:ODEFunction ? f : convert(ODEFunction,f)
```
ChrisRackauckas: you don't think having the constructor is good?

YingboMa: Its signature is the same as https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/diffeqfunction.jl#L99. I introduced the method collision in my previous PR.

PR #122, opened 2018-07-17
Fix method redefinition warnings

YingboMa:


ChrisRackauckas:
I don't know how this'll work with `remake`, but we'll see.

ChrisRackauckas:
This didn't work and was reverted.

PR #127, opened 2018-07-24
Add p kwarg to MonteCarloProblem constructor

Vaibhavdixit02:


src/problems/monte_problems.jl
```diff
@@ -12,3 +12,10 @@ end
     reduction = (u,data,I)->(append!(u,data),false),
     u_init = []) =
     MonteCarloProblem(prob,prob_func,output_func,reduction,u_init)
+
+@add_kwonly MonteCarloProblem(;prob,
```
ChrisRackauckas: no macro needed. Also take it off the above.

src/problems/monte_problems.jl
```diff
@@ -12,3 +12,10 @@ end
...
+    output_func = (sol,i)-> (sol,false),
+    prob_func= (prob,i,repeat)->prob,
+    reduction = (u,data,I)->(append!(u,data),false),
+    u_init = [], p = []) =
```
ChrisRackauckas: `p = nothing`

PR #136, opened 2018-08-11
Fix plot recipe for `sol` (backport to 0.7)

tkf:
Backport @YingboMa's #134 to 0.7 branch.


tkf:
Actually, this reduces warning so it makes sense to pull.  Reopening it.  (I was confused by something else when I closed it)

PR #137, opened 2018-08-11
Fix plot(sol, tspan=...)

tkf:
Otherwise you get error like `MethodError(keys, (getfield(DiffEqBase, Symbol("##78#84"))(Core.Box((70, 90))),), 0x0000000000006c9b)`.  I'm sending a PR to 07-fix branch since Plots.jl is not ready yet for 1.0.

It fixes some semantics of the previous code.  Reading the code history, I _guess_ this is the correct implementation.  But rest of the plotting function is so large and I'm not 100% sure about this...

### DIFF
--- a/src/solutions/solution_interface.jl
+++ b/src/solutions/solution_interface.jl
@@ -78,8 +78,8 @@ DEFAULT_PLOT_FUNC(x,y,z) = (x,y,z) # For v0.5.2 bug
     end
     start_idx = 1
   else
-    start_idx = something(findfirst(isequal(sol.t), x -> x>=tspan[end]), 0)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Yup this looks good. And yes, this plotting function is nefarious yet I am not sure how to simplify it better... that's definitely something I would like to get done.

tkf:
The plot function seems to be doing a lot of thing so I guess it's ok to big.  But it would be nice to have at least some kind of somke tests (which could have prevent from this kind of regression).  I need to look at Plots.jl to see if it's easy to implement a "no-op" backend which just runs the recipe code (but plot nothing).

PR #139, opened 2018-08-14
Fix initialisations in interpolation method

Libbum:
Have been migrating a few simple projects through the 0.7-1.0 process and hit a depreciation that might have been overlooked:

> ┌ Warning: `Array{T, 1}(m::Int) where T` is deprecated, use `Array{T, 1}(undef, m)` instead.
> │   caller = interpolation(::Array{Float64,1}, ::Function, ::Nothing, ::Type, ::Array{Float64,1}) at interpolation.jl:44
> └ @ DiffEqBase ~/.julia/packages/DiffEqBase/mKOqQ/src/interpolation.jl:44

There are three initialisation calls here that need `undef` included now&mdash;a simple enough fix.


Libbum:
Ahh, sorry - I see you have a dedicated 07-fix branch. I just updated the PR branch.

ChrisRackauckas:
Can you rebase this for master? Are these not fixed?

ChrisRackauckas:
Thanks!

PR #141, opened 2018-08-18
remove remake branch

ChrisRackauckas:
@tkf what about removing the branch by just adding a method to `NoiseProblem`?


ChrisRackauckas:
Good enough for now and can keep improving.

PR #143, opened 2018-08-19
Fixing remake

tkf:
PR to #141

### DIFF
--- a/src/problems/ode_problems.jl
+++ b/src/problems/ode_problems.jl
@@ -9,9 +9,10 @@ struct ODEProblem{uType,tType,isinplace,P,F,C,PT} <:
   p::P
   callback::C
   problem_type::PT
-  @add_kwonly function ODEProblem(f::AbstractODEFunction,u0,tspan,p=nothing,
-                      problem_type=StandardODEProblem();
-                      callback=nothing)
+  @add_kwonly function ODEProblem{iip}(f::AbstractODEFunction{iip},
+                                       u0,tspan,p=nothing,
+                                       problem_type=StandardODEProblem();
+                                       callback=nothing) where {iip}
     _tspan = promote_tspan(tspan)
     new{typeof(u0),typeof(_tspan),
        isinplace(f),typeof(p),typeof(f),
@@ -23,6 +24,10 @@ struct ODEProblem{uType,tType,isinplace,P,F,C,PT} <:
   function ODEProblem{iip}(f,u0,tspan,p=nothing;kwargs...) where {iip}
     ODEProblem(convert(ODEFunction{iip},f),u0,tspan,p;kwargs...)
   end
+end
+
+function ODEProblem(f::AbstractODEFunction,u0,tspan,args...;kwargs...)
+  ODEProblem{isinplace(f)}(f,u0,tspan,args...;kwargs...)
 end
 
 function ODEProblem(f,u0,tspan,p=nothing;kwargs...)
--- a/src/remake.jl
+++ b/src/remake.jl
@@ -3,6 +3,16 @@
   Expr(:tuple, A...)
 end
 
+remaker_of(prob::T) where {T} = parameterless_type(T){isinplace(prob)}
+
+# Define `remaker_of` for the types that does not (make sense to)
+# implement `isinplace` trait:
+for T in [
+    NoiseProblem,
+    ]
+  @eval remaker_of(::$T) = $T
+end
+
 """
     remake(thing; <keyword arguments>)
 
@@ -10,6 +20,6 @@ Re-construct `thing` with new field values specified by the keyword
 arguments.
 """
 function remake(thing; kwargs...)
-  T = parameterless_type(typeof(thing))
-  T{isinplace(thing)}(; struct_as_namedtuple(thing)...,kwargs...)
+  T = remaker_of(thing)
+  T(; struct_as_namedtuple(thing)...,kwargs...)
 end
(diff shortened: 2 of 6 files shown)


test/remake_tests.jl
```diff
@@ -28,7 +28,7 @@ tspan = (0,1.0)
 
 # Create a ODEProblem and test remake:
 prob1 = SplitODEProblem(f,f,u0,tspan,Dict())
-prob2 = remake(prob1; u0 = prob1.u0 .+ 1)
+prob2 = @inferred remake(prob1; u0 = prob1.u0 .+ 1)
```
tkf: It works! (in my laptop at least) :tada:

src/remake.jl
```diff
@@ -0,0 +1,27 @@
+@generated function struct_as_namedtuple(st)
+  A = ( Expr(:(=), n, :(st.$n)) for n in fieldnames(st))
+  Expr(:tuple, A...)
+end
+
+remaker_of(prob::T) where {T} = parameterless_type(T){isinplace(prob)}
```
tkf: We can also probably `@pure`ify it :)

src/remake.jl
```diff
@@ -0,0 +1,27 @@
...
+
+# Define `remaker_of` for the types that does not (make sense to)
+# implement `isinplace` trait:
+for T in [
+    NoiseProblem,
```
tkf: I moved `remake` related stuff to a separate file since I have to loop over the structs.

tkf:
This PR (at least at this point) is not complete since there are many problem types without with-iip keyword-only constructors, e.g.:

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/916abeb0ac4b5815a830e1e0b51fc089bd562380/src/problems/dde_problems.jl#L12

src/problems/ode_problems.jl
```diff
@@ -25,6 +26,10 @@ struct ODEProblem{uType,tType,isinplace,P,F,C,PT} <:
   end
 end
 
+function ODEProblem(f::AbstractODEFunction,u0,tspan,args...;kwargs...)
+  ODEProblem{isinplace(f)}(f,u0,tspan,args...;kwargs...)
```
tkf: Implementing this for all problem types is a bit annoying since we already have:

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/916abeb0ac4b5815a830e1e0b51fc089bd562380/src/problems/ode_problems.jl#L28-L30

How about implementing `convert(::Type{AbstractODEFunction}, _)` instead of `convert(::Type{ODEFunction}, _)`?  There are, e.g., `convert(::Type{AbstractArray{T}}, a::AbstractArray)` in `Base` so I think it's fine to define `convert` to abstract types?

tkf: So we need changes like this if we were to go to `convert(::Type{AbstractODEFunction}, _)` direction:

```diff
diff --git a/src/diffeqfunction.jl b/src/diffeqfunction.jl
index 1fdf2b1..fba8833 100644
--- a/src/diffeqfunction.jl
+++ b/src/diffeqfunction.jl
@@ -487,7 +487,9 @@ has_paramjac(f::DynamicalODEFunction) = f.f1.paramjac != nothing
 ######### Compatibility Constructor from Tratis

 ODEFunction{iip}(f::T) where {iip,T} = return T<:ODEFunction ? f : convert(ODEFunction{iip},f)
-function Base.convert(::Type{ODEFunction},f)
+
+Base.convert(::Type{AbstractODEFunction}, f::AbstractODEFunction) = f
+function Base.convert(::Type{AbstractODEFunction},f)
   if __has_analytic(f)
     analytic = (args...) -> f(Val{:analytic},args...)
   else
diff --git a/src/problems/ode_problems.jl b/src/problems/ode_problems.jl
index 74fd2a6..4b38f89 100644
--- a/src/problems/ode_problems.jl
+++ b/src/problems/ode_problems.jl
@@ -26,12 +26,9 @@ struct ODEProblem{uType,tType,isinplace,P,F,C,PT} <:
   end
 end

-function ODEProblem(f::AbstractODEFunction,u0,tspan,args...;kwargs...)
-  ODEProblem{isinplace(f)}(f,u0,tspan,args...;kwargs...)
-end
-
-function ODEProblem(f,u0,tspan,p=nothing;kwargs...)
-  ODEProblem(convert(ODEFunction,f),u0,tspan,p;kwargs...)
+function ODEProblem(f,u0,tspan,args...;kwargs...)
+  f_ = convert(AbstractODEFunction,f)
+  ODEProblem{isinplace(f_)}(f_,u0,tspan,args...;kwargs...)
 end

 abstract type AbstractDynamicalODEProblem end
```

ChrisRackauckas:
>This PR (at least at this point) is not complete since there are many problem types without with-iip keyword-only constructors, e.g.:

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/916abeb0ac4b5815a830e1e0b51fc089bd562380/src/problems/dde_problems.jl#L30 ?

src/remake.jl
ChrisRackauckas: yes I don't see an issue there.

src/problems/ode_problems.jl
ChrisRackauckas: I'm not immediately seeing how that helps

src/remake.jl
ChrisRackauckas: why is NoiseProblem here? Dropping the overload?

ChrisRackauckas:
Looks like this is good and inference tests pass?

src/remake.jl
tkf: I thought it's much easier to add one line in the list above than implementing `T{iip}` and `isinplace(::T)` to remake `T`.  In some case, it may not make sense for `T` to have `isinplace` trait.

PR #147, opened 2018-08-20
Implement T{iip}(; kwargs...) constructors for remake

tkf:
fix #145 
fix #146

PR #148, opened 2018-08-20
Use Appveyor.jl

tkf:
Copied from https://github.com/JuliaCI/Appveyor.jl and then tweaked `environment.matrix`.


tkf:
Both CIs failed because tests *are* broken. Once #147 is merged, I'll rebase it.

PR #149, opened 2018-08-23
Improve recursivecopy! and copy_fields! considering the types when copying

ronisbr:
Hi!

With this PR, the functions `recursivecopy!` and `copy_fields!` are improved by considering the field type when copying.

This partially fixes the bug JuliaDiffEq/DifferentialEquations.jl#336, because it is now possible to use `SArray` as a field type in `DEDataVector`.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,5 +4,6 @@ RecursiveArrayTools 0.12.3
 Compat 0.19.0
 Requires
 IteratorInterfaceExtensions 0.1.0
+StaticArrays 0.8.3
 TableTraits 0.3.0
 TreeViews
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Thanks! How come it only partially fixes the bug?

ronisbr:
NP!

Now, I can now have a field of type `SArray` but not the vector `x`. The problem seems to be in the `similar` function, but I was not able yet to figure out why.

ChrisRackauckas:
You cannot `similar` an `SArray` since `similar` returns a mutable type (for `SArray` it returns an `MArray`)

PR #150, opened 2018-08-29
Fix BVProblem constructor

AttilaKovacs:
The first example (pendulum) in the [docs](http://docs.juliadiffeq.org/latest/tutorials/bvp_example.html) threw the following error:
```julia
ERROR: MethodError: no method matching has_analytic(::typeof(simplependulum!))
Closest candidates are:
  has_analytic(::DiffEqBase.AbstractDiffEqFunction) at C:\Users\Attila\.julia\packages\DiffEqBase\YSk7M\src\diffeqfunction.jl:466
```
I checked the commits and it seems the constructor for the `BVProblem` was not updated in PR #147, so the input function is not converted to the `ODEFunction`.
With these changes the example now works properly and it seems I did not break anything, however feel free to comment if this messes up anything else.

### DIFF
--- a/src/problems/bvp_problems.jl
+++ b/src/problems/bvp_problems.jl
@@ -8,7 +8,7 @@ struct BVProblem{uType,tType,isinplace,P,F,bF,PT,CB} <: AbstractBVProblem{uType,
     p::P
     problem_type::PT
     callback::CB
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
That looks good, thanks!

PR #152, opened 2018-08-30
Fix #151, using Measurements and DiffEqBase results in error.

tkoolen:

PR #153, opened 2018-08-30
Fix for method ambiguity errors with dual_number_warn and measurements_warn

tkoolen:
Ref https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/152\#issuecomment-417401398.

I guess `dual_number_warn` is also broken in the same way, right? 

I don't know how I feel about this whole setup by the way. What makes these two special?


ChrisRackauckas:
These just seem to be FAQs, so I thought having Requires.jl integration for non-mandatory warning could be a good thing now that we can do it. Now that we could, the question is whether we should, and I'm on the fence.

PR #154, opened 2018-09-06
Change interpolation API for #203

YingboMa:
Ref: https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/203


ChrisRackauckas:
Yes DAEs too. Also there's https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/interpolation.jl which is used by the other packages.

PR #158, opened 2018-10-01
Change `f.f1` to be always an `ODEFunction`

YingboMa:


ChrisRackauckas:
Looks like this broke OrdinaryDiffEq.jl tests?

PR #160, opened 2018-10-02
logspace replace by base.^range(..)

tungli:
`logspace` does not work anymore in Julia 1.0. I replaced it with `10.0^range(...)` because of an error when using `plot` from Plots.jl


ChrisRackauckas:
Thanks!

PR #162, opened 2018-10-09
Add matrix free interface for Jacobian

YingboMa:


ChrisRackauckas:
Can't this be done with just a Jacobian type that defines `mul!`?

YingboMa:
Maybe we can add a `DiffEqOperators.MatrixFreeOperator`?

ChrisRackauckas:
Yes that would be a good idea, one that makes it easy to define one from a multiplication function.

YingboMa:
On another thought, maybe a much easier way to do it is to differentiate `jac(u, p, t)` and `jac(v, u, p, t)` for the out-of-place version and similarly for the in-place ones.

ChrisRackauckas:
Handled alternatively.

PR #164, opened 2018-10-14
Add callbacks handling

YingboMa:
Ref: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/507

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -115,12 +115,12 @@ Base.isempty(cb::AbstractDiscreteCallback) = false
 # Use Recursion to find the first callback for type-stability
 
 # Base Case: Only one callback
-function find_first_continuous_callback(integrator, callback::DiffEqBase.AbstractContinuousCallback)
+function find_first_continuous_callback(integrator, callback::AbstractContinuousCallback)
   (find_callback_time(integrator,callback,1)...,1,1)
 end
 
 # Starting Case: Compute on the first callback
-function find_first_continuous_callback(integrator, callback::DiffEqBase.AbstractContinuousCallback, args...)
+function find_first_continuous_callback(integrator, callback::AbstractContinuousCallback, args...)
   find_first_continuous_callback(integrator,find_callback_time(integrator,callback,1)...,1,1,args...)
 end
 
@@ -144,7 +144,7 @@ end
 @inline function determine_event_occurance(integrator,callback,counter)
   event_occurred = false
   if callback.interp_points!=0
-    DiffEqBase.addsteps!(integrator)
+    addsteps!(integrator)
   end
   Θs = range(typeof(integrator.t)(0), stop=typeof(integrator.t)(1), length=callback.interp_points)
   interp_index = 0
@@ -169,10 +169,10 @@ end
     # Since near even we use direction instead of location to reset
 
     if callback.interp_points==0
-      DiffEqBase.addsteps!(integrator)
+      addsteps!(integrator)
     end
 
-    if DiffEqBase.ismutablecache(integrator.cache)
+    if ismutablecache(integrator.cache)
       if typeof(callback.idxs) <: Nothing
         tmp = integrator.cache.tmp
       else !(typeof(callback.idxs) <: Number)
@@ -181,7 +181,7 @@ end
     end
 
     abst = integrator.tprev+integrator.dt*100*eps(integrator.tprev)
-    if DiffEqBase.ismutablecache(integrator.cache) && !(typeof(callback.idxs) <: Number)
+    if ismutablecache(integrator.cache) && !(typeof(callback.idxs) <: Number)
       integrator(tmp,abst,Val{0},idxs=callback.idxs)
     else
       tmp = integrator(abst,Val{0},idxs=callback.idxs)
@@ -208,8 +208,8 @@ end
   if ((prev_sign<0 && !(typeof(callback.affect!)<:Nothing)) || (prev_sign>0 && !(typeof(callback.affect_neg!)<:Nothing))) && prev_sign*next_sign<=0
     event_occurred = true
     interp_index = callback.interp_points
-  elseif callback.interp_points!=0  && !DiffEqBase.isdiscretealg(integrator.alg) # Use the interpolants for safety checking
-    if DiffEqBase.ismutablecache(integrator.cache)
+  elseif callback.interp_points!=0  && !isdiscretealg(integrator.alg) # Use the interpolants for safety checking
+    if ismutablecache(integrator.cache)
       if typeof(callback.idxs) <: Nothing
         tmp = integrator.cache.tmp
       else !(typeof(callback.idxs) <: Number)
@@ -218,7 +218,7 @@ end
     end
     for i in 2:length(Θs)
       abst = integrator.tprev+integrator.dt*Θs[i]
-      if DiffEqBase.ismutablecache(integrator.cache) && !(typeof(callback.idxs) <: Number)
+      if ismutablecache(integrator.cache) && !(typeof(callback.idxs) <: Number)
         integrator(tmp,abst,Val{0},idxs=callback.idxs)
       else
         tmp = integrator(abst,Val{0},idxs=callback.idxs)
@@ -250,8 +250,8 @@ function find_callback_time(integrator,callback,counter)
         top_Θ = typeof(integrator.t)(1)
         bottom_θ = typeof(integrator.t)(0)
       end
-      if callback.rootfind && !DiffEqBase.isdiscretealg(integrator.alg)
-        if DiffEqBase.ismutablecache(integrator.cache)
+      if callback.rootfind && !isdiscretealg(integrator.alg)
+        if ismutablecache(integrator.cache)
           _cache = first(get_tmp_cache(integrator))
           if typeof(callback.idxs) <: Nothing
             tmp = _cache
@@ -261,7 +261,7 @@ function find_callback_time(integrator,callback,counter)
         end
         zero_func = (Θ) -> begin
           abst = integrator.tprev+integrator.dt*Θ
-          if DiffEqBase.ismutablecache(integrator.cache) && !(typeof(callback.idxs) <: Number)
+          if ismutablecache(integrator.cache) && !(typeof(callback.idxs) <: Number)
             integrator(tmp,abst,Val{0},idxs=callback.idxs)
           else
             tmp = integrator(abst,Val{0},idxs=callback.idxs)
@@ -298,7 +298,7 @@ function find_callback_time(integrator,callback,counter)
         # a float which is slightly after, making it out of the domain, causing
         # havoc.
         new_t = integrator.dt*Θ
-      elseif interp_index != callback.interp_points && !DiffEqBase.isdiscretealg(integrator.alg)
+      elseif interp_index != callback.interp_points && !isdiscretealg(integrator.alg)
         new_t = integrator.dt*Θs[interp_index]
       else
         # If no solve and no interpolants, just use endpoint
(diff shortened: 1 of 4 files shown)


src/integrator_interface.jl
```diff
@@ -34,6 +34,11 @@ set_reltol!(i::DEIntegrator,t) = error("set_reltol!: method has not been impleme
 reinit!(integrator::DEIntegrator,args...; kwargs...) =
        error("reinit!: method has not been implemented for the integrator")
 auto_dt_reset!(integrator::DEIntegrator) = error("auto_dt_reset!: method has not been implemented for the integrator")
+change_t_via_interpolation!(i::DEIntegrator,args...) = error("change_t_via_interpolation!: method has not been implemented for the integrator")
+ismutablecache(cache::DECache) = error("ismutablecache method has not been implemented for the cache")
+addsteps!(i::DEIntegrator,args...) = nothing
+isdiscretealg(alg::DEAlgorithm) = false
```
ChrisRackauckas: This is an algorithm trait.

src/integrator_interface.jl
```diff
@@ -34,6 +34,11 @@ set_reltol!(i::DEIntegrator,t) = error("set_reltol!: method has not been impleme
 reinit!(integrator::DEIntegrator,args...; kwargs...) =
        error("reinit!: method has not been implemented for the integrator")
 auto_dt_reset!(integrator::DEIntegrator) = error("auto_dt_reset!: method has not been implemented for the integrator")
+change_t_via_interpolation!(i::DEIntegrator,args...) = error("change_t_via_interpolation!: method has not been implemented for the integrator")
+ismutablecache(cache::DECache) = error("ismutablecache method has not been implemented for the cache")
```
ChrisRackauckas: This is a cache trait

src/integrator_interface.jl
```diff
@@ -34,6 +34,11 @@ set_reltol!(i::DEIntegrator,t) = error("set_reltol!: method has not been impleme
...
+addsteps!(i::DEIntegrator,args...) = nothing
+isdiscretealg(alg::DEAlgorithm) = false
+reeval_internals_due_to_modification!(integrator::DEIntegrator) = nothing
```
ChrisRackauckas: we need this here?

ChrisRackauckas:
Has this been tested with Sundials, StochasticDiffEq, and ODEInterfaceDiffEq to make sure it's all on the same page? I think when we merge to change to this we'll need to be careful about further interface changes to it, so we should make sure it's compatible first.

src/integrator_interface.jl
YingboMa: Yes, we need it for OrdinaryDiffEq.jl

YingboMa:
`Sundials.jl`, `ODEInterfaceDiffEq.jl`, `StochasticDiffEq.jl` and `OrdinaryDiffEq.jl` tests passed locally. :tada: @ChrisRackauckas

src/callbacks.jl
```diff
@@ -107,3 +107,344 @@ end
 Base.isempty(cb::CallbackSet) = isempty(cb.continuous_callbacks) && isempty(cb.discrete_callbacks)
 Base.isempty(cb::AbstractContinuousCallback) = false
 Base.isempty(cb::AbstractDiscreteCallback) = false
+
+#======================================================#
+# Callback handling
+#======================================================#
+
+function get_tmp(integrator::DEIntegrator, callback)
```
ChrisRackauckas: Why is this needed? Why is this not just `get_tmp_cache`?

src/callbacks.jl
```diff
@@ -107,3 +107,344 @@ end
...
+  _tmp = get_tmp_cache(integrator)
+  _tmp === nothing && return nothing
+  _cache = first(_tmp)
+  if callback.idxs isa Nothing
```
ChrisRackauckas: oh I see, just a convenience.

src/callbacks.jl
```diff
@@ -107,3 +107,344 @@ end
...
+  elseif typeof(callback.idxs) <: Number
+    next_sign = sign(callback.condition(integrator.u[callback.idxs],integrator.t,integrator))
+  else
+    next_sign = sign(callback.condition(@view(integrator.u[callback.idxs]),integrator.t,integrator))
+  end
+
+  if ((prev_sign<0 && !(typeof(callback.affect!)<:Nothing)) || (prev_sign>0 && !(typeof(callback.affect_neg!)<:Nothing))) && prev_sign*next_sign<=0
+    event_occurred = true
+    interp_index = callback.interp_points
+  elseif callback.interp_points!=0 && !isnative(integrator) || !isdiscrete(integrator.alg) # Use the interpolants for safety checking
```
ChrisRackauckas: why `!isnative(integrator)`?

YingboMa: Because if it is not a native integrator, then `integrator.alg` might not exist.

YingboMa: It is missing in `ODEInterfaceDiffEq.jl`.
https://github.com/JuliaDiffEq/ODEInterfaceDiffEq.jl/blob/master/src/integrator_types.jl#L8-L20

ChrisRackauckas: Lets add it.

ChrisRackauckas:
What about adding a `getproperty` overload for `dt`, and adding it to the integrator interface?

ChrisRackauckas:
I don't think there's a reason for it. I think that's just a natural error when copying over the same strategy across 4 repos haha

ChrisRackauckas:
That whole strategy right there is subject to change though. I'm not sure if scaling it by dt is a good idea anyways.

ChrisRackauckas:
IMO, having an extra non-standard environment which only Travis (by default) runs to test a few downstream things like this would be really nice.

src/callbacks.jl
ChrisRackauckas: so is `isnative` still necessary here?

src/callbacks.jl
```diff
@@ -107,3 +107,283 @@ end
...
+        #Θ = prevfloat(...)
+        # prevfloat guerentees that the new time is either 1 floating point
+        # numbers just before the event or directly at zero, but not after.
+        # If there's a barrier which is never supposed to be crossed,
+        # then this will ensure that
+        # The item never leaves the domain. Otherwise Roots.jl can return
+        # a float which is slightly after, making it out of the domain, causing
+        # havoc.
+        new_t = integrator.dt*Θ
+      elseif interp_index != callback.interp_points && !isnative(integrator) || !isdiscrete(integrator.alg)
```
ChrisRackauckas: remove !isnative here?

src/callbacks.jl
```diff
@@ -107,3 +107,283 @@ end
...
+
+    # Only due this if the discontinuity did not move it far away from an event
+    # Since near even we use direction instead of location to reset
+
+    if callback.interp_points==0
+      addsteps!(integrator)
+    end
+
+    tmp = get_tmp(integrator, callback)
+    ismutable = !(tmp === nothing)
```
ChrisRackauckas: why is this here?

src/callbacks.jl
```diff
@@ -107,3 +107,283 @@ end
...
+  else
+    next_sign = sign(callback.condition(@view(integrator.u[callback.idxs]),integrator.t,integrator))
+  end
+
+  if ((prev_sign<0 && !(typeof(callback.affect!)<:Nothing)) || (prev_sign>0 && !(typeof(callback.affect_neg!)<:Nothing))) && prev_sign*next_sign<=0
+    event_occurred = true
+    interp_index = callback.interp_points
+  elseif callback.interp_points!=0 && !isnative(integrator) || !isdiscrete(integrator.alg) # Use the interpolants for safety checking
+    tmp = get_tmp(integrator, callback)
+    ismutable = !(tmp === nothing)
```
ChrisRackauckas: is this needed?

src/callbacks.jl
```diff
@@ -107,3 +107,283 @@ end
...
+      if callback.interp_points!=0
+        top_Θ = Θs[interp_index] # Top at the smallest
+        bottom_θ = Θs[prev_sign_index]
+      else
+        top_Θ = typeof(integrator.t)(1)
+        bottom_θ = typeof(integrator.t)(0)
+      end
+      if callback.rootfind && !isnative(integrator) || !isdiscrete(integrator.alg)
+        tmp = get_tmp(integrator, callback)
+        ismutable = !(tmp === nothing)
```
ChrisRackauckas: this isn't needed too

src/callbacks.jl
```diff
@@ -107,3 +107,283 @@ end
...
+      new_t = zero(typeof(integrator.t))
+    else
+      if callback.interp_points!=0
+        top_Θ = Θs[interp_index] # Top at the smallest
+        bottom_θ = Θs[prev_sign_index]
+      else
+        top_Θ = typeof(integrator.t)(1)
+        bottom_θ = typeof(integrator.t)(0)
+      end
+      if callback.rootfind && !isnative(integrator) || !isdiscrete(integrator.alg)
```
ChrisRackauckas: `isnative` isn't needed?

src/callbacks.jl
```diff
@@ -107,3 +107,283 @@ end
...
+  if isnative(integrator) # if we have integrator.cache
+    ismutable && !(typeof(callback.idxs) isa Number) ? integrator(tmp,abst,Val{0},idxs=callback.idxs) :
+                                                       tmp = integrator(abst,Val{0},idxs=callback.idxs)
+    return callback.condition(tmp,evalt,integrator)
+  else # Sundials or ODEInterfaceDiffEq
+    if !(typeof(callback.idxs) <: Number)
+      ismutable ? integrator(tmp,abst) : tmp = integrator(abst)
+      callback.idxs == nothing ? _tmp = tmp : _tmp = @view tmp[callback.idxs]
+    else
+      _tmp = integrator(abst)[callback.idxs]
```
ChrisRackauckas: after this PR, we should try making this `idxs = callback.idxs`.

src/callbacks.jl
YingboMa: Ah, I forgot to delete it.

src/callbacks.jl
```diff
@@ -107,3 +107,283 @@ end
...
+  else
+    tmp = nothing
+  end
+  return tmp
+end
+
+function get_condition(integrator::DEIntegrator, callback, abst, evalt=abst)
+  tmp = get_tmp(integrator, callback)
+  ismutable = !(tmp === nothing)
+  if isnative(integrator) # if we have integrator.cache
```
ChrisRackauckas: >if we have integrator.cache

Why does that matter? That doesn't seem to show up here?

src/callbacks.jl
```diff
@@ -129,7 +129,7 @@ end
 function get_condition(integrator::DEIntegrator, callback, abst, evalt=abst)
   tmp = get_tmp(integrator, callback)
   ismutable = !(tmp === nothing)
-  if isnative(integrator) # if we have integrator.cache
+  if isnative(integrator)
```
ChrisRackauckas: why is isnative still necessary?

YingboMa: Sundials doesn't support `idx`, I tried to add it, however, I got segmentation faults. So, I kept this condition.

YingboMa:
Just addressed review comments. Tests still pass.

src/callbacks.jl
```diff
@@ -183,10 +173,8 @@ end
   # Check if the event occured
   if typeof(callback.idxs) <: Nothing
     previous_condition = callback.condition(integrator.uprev,integrator.tprev,integrator)
-  elseif typeof(callback.idxs) <: Number
-    previous_condition = callback.condition(integrator.uprev[callback.idxs],integrator.tprev,integrator)
   else
-    previous_condition = callback.condition(@view(integrator.uprev[callback.idxs]),integrator.tprev,integrator)
+    @views previous_condition = callback.condition(integrator.uprev[callback.idxs],integrator.tprev,integrator)
```
ChrisRackauckas: Does this handle the number case?

YingboMa: Yes, `@views` expands to `maybeview` which handles number.

ChrisRackauckas: beautiful. That makes the code much nicer.

PR #165, opened 2018-10-16
Pass p of ODEProblem to SteadyStateProblem

tkf:
Is it OK to assume `prob.p` exists for any `AbstractODEProblem`?


ChrisRackauckas:
Yes, that's a fine assumption.

PR #170, opened 2018-10-21
fix saved_in_cb

ChrisRackauckas:
@MSeeker1340 I noticed that `saved_in_cb` wasn't being set even when the callbacks were saving. This fixes a lot of tests in DiffEqCallbacks.jl. Testing to see what happens to other packages before merging.


ChrisRackauckas:
There doesn't seem to be a way to express "only save after the callback is applied, or if not applied". For example, with the PositiveDomain callback, we only want to save positive values, which is either true or happens after the effect. but the first `savevalues!` will cause a negative value to save before the callback is applied. There should be a better answer to this. I would assume this effects PuMaS?

PR #180, opened 2018-12-17
add type parameters to DEIntegrator

Datseris:


ChrisRackauckas:
Why would something without time have a time parameter?...

PR #190, opened 2019-02-21
Add DEStat

YingboMa:
Ref: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/667


src/diffeqfunction.jl
```diff
@@ -186,7 +202,7 @@ function ODEFunction{iip,true}(f;
                  typeof(jac),typeof(jac_prototype),typeof(invW),typeof(invW_t),
                  typeof(paramjac),typeof(syms)}(
                  f,mass_matrix,analytic,tgrad,jac,jac_prototype,invW,invW_t,
-                 paramjac,syms)
+                 paramjac,syms,Ref(Int(1)))
```
ChrisRackauckas: why 1?

YingboMa: Oops, that is a mistake.

src/destats.jl
```diff
@@ -0,0 +1,20 @@
+mutable struct DEStat{E}
```
ChrisRackauckas: DEStats

src/DiffEqBase.jl
```diff
@@ -117,6 +117,7 @@ include("solutions/rode_solutions.jl")
 include("solutions/dae_solutions.jl")
 include("solutions/monte_solutions.jl")
 include("solutions/solution_interface.jl")
+include("DEStats.jl")
```
eljungsk: Hi,

This line unfortunately breaks this package on Linux due to case sensitivity.

PR #192, opened 2019-02-24
Make RecursiveFactorization.lu the default factorization

YingboMa:


ChrisRackauckas:
What is the downside of doing this? Does OrdinaryDiffEq.jl's test pass with this?

PR #195, opened 2019-03-01
Fix type instability in build_solution

devmotion:
Currently the type of `sol(ode_prob, alg)` can not be inferred for algorithms in ODEInterfaceDiffEq and LSODA. Here's a (slightly complicated) example:
```julia
using ParameterizedFunctions, DiffEqBase
rober = @ode_def Rober begin
  dy₁ = -k₁*y₁+k₃*y₂*y₃
  dy₂ =  k₁*y₁-k₂*y₂^2-k₃*y₂*y₃
  dy₃ =  k₂*y₂^2
end k₁ k₂ k₃
prob = ODEProblem(rober,[1.0;0.0;0.0],(0.0,1e11),(0.04,3e7,1e4))

using ODEInterfaceDiffEq
@code_warntype solve(prob,radau())

using LSODA
@code_warntype solve(prob,lsoda())
```

### DIFF
--- a/src/solutions/ode_solutions.jl
+++ b/src/solutions/ode_solutions.jl
@@ -21,8 +21,7 @@ function build_solution(
         dense=false,dense_errors=dense,
         calculate_error = true,
         k=[],
-        du=[],
-        interp = !isempty(du) ? HermiteInterpolation(t,u,du) : LinearInterpolation(t,u),
+        interp = LinearInterpolation(t,u),
         retcode = :Default, destats=DEStats(), kwargs...)
 
   T = eltype(eltype(u))


src/solutions/ode_solutions.jl
```diff
@@ -21,8 +21,7 @@ function build_solution(
         dense=false,dense_errors=dense,
         calculate_error = true,
         k=[],
-        du=[],
-        interp = !isempty(du) ? HermiteInterpolation(t,u,du) : LinearInterpolation(t,u),
+        interp = LinearInterpolation(t,u),
```
ChrisRackauckas: does this keep Sundials as Hermite?

devmotion: The interpolation is passed explicitly in https://github.com/JuliaDiffEq/Sundials.jl/blob/master/src/common_interface/solve.jl#L230-L231, https://github.com/JuliaDiffEq/Sundials.jl/blob/master/src/common_interface/solve.jl#L502-L503, and https://github.com/JuliaDiffEq/Sundials.jl/blob/master/src/common_interface/solve.jl#L773-L774, so that should be no problem. However, lines https://github.com/JuliaDiffEq/Sundials.jl/blob/master/src/common_interface/solve.jl#L229, https://github.com/JuliaDiffEq/Sundials.jl/blob/master/src/common_interface/solve.jl#L501, and https://github.com/JuliaDiffEq/Sundials.jl/blob/master/src/common_interface/solve.jl#L772 have to be removed.

ChrisRackauckas: okay cool. But this is then a change that will be breaking? So we'll need to update Sundials immediately afterwards?

devmotion: No, since the interpolation object is already passed explicitly and `du` is never used in `build_solution` apart from creating the interpolation object, it should be a non-breaking change in Sundials.

devmotion: Yes, this PR breaks Sundials but I try to verify that removing these three lines fixes it.

devmotion: Oh no, wait it's actually non-breaking - I missed that `build_solution` has a `kwargs...` argument. So everything is fine, but still those lines should be removed in Sundials since they are not needed anymore (and never were, actually).

PR #199, opened 2019-03-15
API Documentation

jlumpe:
This is a work in progress of documentation for the package's API, which isn't really covered in depth in DiffEqDocs.jl. I'm basically just adding docstrings to the important functions and types and then organizing references to them in some Documenter.jl files. So far I've covered most of the problem types and added stubs for the `AbstractDiffEqFunction` subtypes.

My goal here is to get enough of the documentation structure set up so that myself and others can continue filling in the docstrings in smaller increments.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -8,3 +8,4 @@ StaticArrays 0.8.3
 TableTraits 0.3.0
 TreeViews
 Roots
+DocStringExtensions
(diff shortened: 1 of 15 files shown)


ChrisRackauckas:
Great stuff? Would you by chance happen to be interested in? We could put in an application for Julia. https://developers.google.com/season-of-docs/docs/timeline

jlumpe:
Awesome, glad this is useful. Unfortunately I too much going on right now to commit to anything with deadlines, but I'm planning on submitting more PRs soon.

PR #204, opened 2019-04-12
A even faster broadcast that assumes no aliasing

YingboMa:


src/diffeqfastbc.jl
```diff
@@ -50,11 +50,6 @@ macro ..(x)
     if expr.head == :(.=)
```
ChrisRackauckas: does .+= already expand to .= by here?

PR #212, opened 2019-04-25
Add MonteCarloMeasurements ODE_DEFAULT_NORM

baggepinnen:
This is still WIP since I can not get it to work, the code below causes a call to 
`convert(Float64, ::Particles{Float64,500})` which is not supported unless all particles are the same. Measurements.jl handles this convert the same, i.e., only allows it if the error is zero.

After having added the code in this PR, I have
```julia
julia> g
(500 Particles: 9.79 ± 0.02)

julia> DiffEqBase.ODE_DEFAULT_NORM(g,1.)
9.789991179929432
```
which indicates that the method is there and is producing a float as expected.

Any help appreciated :)

```julia
using MonteCarloMeasurements, OrdinaryDiffEq, Plots
using MonteCarloMeasurements: ±
g = 9.79 ± 0.02; # Gravitational constants
L = 1.00 ± 0.01; # Length of the pendulum

#Initial Conditions
u₀ = [0 ± 0, π / 60 ± 0.01] # Initial speed and initial angle
tspan = (0.0, 6.3)

#Define the problem
function simplependulum(du,u,p,t)
    θ  = u[1]
    dθ = u[2]
    du[1] = dθ
    du[2] = -(g/L)*θ
end

#Pass to solvers
prob = ODEProblem(simplependulum, u₀, tspan)
sol = solve(prob, Tsit5(), reltol = 1e-6)
```

### DIFF
--- a/src/init.jl
+++ b/src/init.jl
@@ -45,6 +45,20 @@ function __init__()
     end
     @inline ODE_DEFAULT_NORM(u::Measurements.Measurement,t) = abs(Measurements.value(u))
   end
+  
+  @require MonteCarloMeasurements="0987c9cc-fe09-11e8-30f0-b96dd679fdca" begin
+
+    value(x::MonteCarloMeasurements.AbstractParticles) = mean(x)
+
+    # Support adaptive steps should be errorless
+    @inline function ODE_DEFAULT_NORM(u::AbstractArray{<:MonteCarloMeasurements.AbstractParticles,N},t) where {N}
+      sqrt(mean(x->ODE_DEFAULT_NORM(x[1],x[2]),zip((value(x) for x in u),Iterators.repeated(t))))
+    end
+    @inline function ODE_DEFAULT_NORM(u::Array{<:MonteCarloMeasurements.AbstractParticles,N},t) where {N}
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
The issue is that for numbers, `eltype(p) == typeof(p)`. This assumption is checked when recursing the vectors to find the element type at the bottom, which is used to extract the unitless element type for unit support. Here, the unitless element type comes up as `Float64`, since it goes one layer too deep. Either this needs to be special cased in RecrusiveArrayTools.jl or it should follow the conventions of a `Number` in MonteCarloMeasurements.

baggepinnen:
Thanks for clarifying! I wasn't aware of the convention for number, but it makes sense and I've updated to follow it. It now works to run the entire example notebook for Measurements.jl with MonteCarloMeasurements.jl and this PR is no longer WIP.
Thank!
https://github.com/JuliaRegistries/General/pull/400

ChrisRackauckas:
Should it not remove the swarm when time is a swarm? I'm not sure how to think about that case here.

baggepinnen:
I think you're right, see https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/215

PR #217, opened 2019-05-08
flip size constraint

dkarrasch:
RecursiveFactorization.jl states that it is faster for sizes above 500, not below 500


ChrisRackauckas:
https://github.com/YingboMa/RecursiveFactorization.jl/tree/master#performance

ChrisRackauckas:
Closing this because I think you read it wrong, but @YingboMa double check please.

dkarrasch:
Yes, seems like. I found the y-axis annotation confusing.

YingboMa:
> Closing this because I think you read it wrong, but @YingboMa double check please.

Yes, the original version is correct.

> Yes, seems like. I found the y-axis annotation confusing.

Using FLOPS is very standard in dense linear algebra.

PR #221, opened 2019-05-20
[WIP] VectorContinuousCallbacks

ChrisRackauckas:
An implementation of what was discussed in https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/331 and various Discourse posts. The idea is for the implementation to not be separate from the one used on ContinuousCallback, since that would promote code reuse and would allow ContinuousCallback to output things like a StaticVector and just work, which would be nice. However, dealing with the caches performantly seems to be an issue. Current issues are:

1. Breaking. Needs to add `previous_condition` and `callback_cache` to the integrator. This should be in the integrator and not in the callback in order to make it thread-safe. These should be allocated by the integrator to get the right element type, but the length needs to be given by the user when the callback is generated.
2.  Right now it has some allocations of `prev_sign` and `next_sign`. These should be integrator caches as well.
3. Right now for catching crossings I'm using `findfirst`. In reality, if there are multiple crossings happening in a given interval, you need to `findall` and then in the rootfinding to a vector rootfind to `findfirst`. However, `finall(true) = [true]`, making it inefficient on the scalar code. So something trickier needs to be done to that part of the code.
4. Since we cannot guarantee a single index in the rootfind, we still need a vector rootfinding method. At least a bisection.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -250,6 +250,14 @@ end
   end
   integrator.sol.destats.ncondition += 1
   ivec = integrator.vector_event_last_time
+  if callback isa VectorContinuousCallback
+    prev_sign = @view(integrator.callback_prev_sign[1:callback.len])
+    next_sign = @view(integrator.callback_next_sign[1:callback.len])
+  else
+    prev_sign = 0.0
+    next_sign = 0.0
+  end
+
   if integrator.event_last_time == counter && minimum(ODE_DEFAULT_NORM(previous_condition[ivec],integrator.t)) < 100ODE_DEFAULT_NORM(integrator.last_event_error,integrator.t)
 
     # If there was a previous event, utilize the derivative at the start to
@@ -273,7 +281,7 @@ end
     # This is necessary to ensure proper non-detection of a root
     # == is for exact floating point equality!
     if callback isa VectorContinuousCallback
-      prev_sign = @. sign(previous_condition)
+      @. prev_sign = sign(previous_condition)
       prev_sign[ivec] = tmp_condition[ivec] > previous_condition[ivec] ? 1.0 :
                     (tmp_condition[ivec] == previous_condition[ivec] ?
                     (prev_sign[ivec] = sign(previous_condition[ivec])) : -1.0)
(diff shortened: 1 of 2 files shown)


tomaklutfu:
For point 3, I think `findall` should be used because there may be several sign change in interval, some of them may actually happened before the others and it is not guaranteed to be the first one in the vector.

kanav99:
Some serious doubts:
`VectorContinuousCallback` is basically a vector of continuous callbacks and we allow the user to write their own logic when a particular set of callbacks' event occurs right? For example if first 3rd and 4th callback fires we do this, if 1st and 4th we do something else and so on.
If what I understand is correct then I think we should have `findall` here https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/221/files#diff-f2ec50e45e2c0d098a69b4e7c72f86f3R287 instead. And for `interp_points` case, all the callbacks can have different interpolation points where they fire, so `interp_points` should be an array too.

ChrisRackauckas:
>How should we deduce the element type? For these are continous callbacks, should we assume AbstractFloat or Float?

Assume it matches the element type of `uType`.

>VectorContinuousCallback is basically a vector of continuous callbacks and we allow the user to write their own logic when a particular set of callbacks' event occurs right? For example if first 3rd and 4th callback fires we do this, if 1st and 4th we do something else and so on.

Just return the index of the first one that fires as part of `affect!(integrator,i)`.

>If what I understand is correct then I think we should have findall here https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/221/files#diff-f2ec50e45e2c0d098a69b4e7c72f86f3R287 instead. And for interp_points case, all the callbacks can have different interpolation points where they fire, so interp_points should be an array too.

It should be `findall`, but we are really looking just for the first.

kanav99:
> Just return the index of the first one that fires as part of affect!(integrator,i).

But if that's the case then what is the difference between a `CallbackSet` of all `ContinuousCallbacks` and `VectorContinuousCallback`? I quite don't understand.

ChrisRackauckas:
>EDIT: Okay so we want to know which callback fired?

Yes, the first event is the `i` which fired.

>But if that's the case then what is the difference between a CallbackSet of all ContinuousCallbacks and VectorContinuousCallback? I quite don't understand.

If you have 1000 ContinuousCallbacks then it's really slow. It's not something that scales well to a large number.

src/callbacks.jl
```diff
@@ -281,18 +281,18 @@ end
 
   prev_sign_index = 1
   abst = integrator.t
-  next_sign = sign(get_condition(integrator, callback, abst))
+  next_sign = sign.(get_condition(integrator, callback, abst))
 
   integrator.sol.destats.ncondition += 1
-  event_idx = findfirst(x-> ((prev_sign<0 && !(typeof(callback.affect!)<:Nothing)) || (prev_sign>0 && !(typeof(callback.affect_neg!)<:Nothing))) && prev_sign*x<=0, next_sign)
+  event_idx = findfirst(x-> ((x[1]<0 && !(typeof(callback.affect!)<:Nothing)) || (x[1]>0 && !(typeof(callback.affect_neg!)<:Nothing))) && x[1]*x[2]<=0, collect(zip(prev_sign,next_sign)))
```
ChrisRackauckas: you don't need to collect if you can iterate it.

src/callbacks.jl
```diff
@@ -304,7 +304,8 @@ end
     end
   end
 
-  event_occurred,interp_index,Θs,prev_sign,prev_sign_index,event_idx
+  prev_sign_of_event_idx = event_idx != nothing ? prev_sign[event_idx] : nothing
```
ChrisRackauckas: there can be more than 1 callback identified as having an event in the first interval.

kanav99:
Current Status: Throws error at random times:
```
ERROR: LoadError: ArgumentError: The interval [a,b] is not a bracketing interval.
You need f(a) and f(b) to have different signs (f(a) * f(b) < 0).
Consider a different bracket or try fzero(f, c) with an initial guess c.
```
Some weird stuff going on, logic seems correct. Took a quiet a lot of time on this because I initially wrote a very different but useless code for this

kanav99:
@ChrisRackauckas I believe I have understood the floating point problem well and made a fix for the case of VectorContinuousCallback. Please have a good look on the last commit before approving as I may have understood the intution wrong. Tested for a identity function and works well for that case.

kanav99:
Now allocations have been removed. But `determine_event_occurance` looks ugly due to many branches on callback being VectorContinuousCallback or not. Should I make a specialized version for both of them?

src/callbacks.jl
```diff
@@ -126,13 +177,22 @@ function get_tmp(integrator::DEIntegrator, callback)
   return tmp
 end
 
-function get_condition(integrator::DEIntegrator, callback, abst)
+function get_condition(integrator::DEIntegrator, callback, abst, out = nothing)
```
ChrisRackauckas: ?

kanav99: oops

src/callbacks.jl
```diff
@@ -172,14 +232,33 @@ end
...
+  if callback isa VectorContinuousCallback
+    prev_sign = @view(integrator.callback_prev_sign[1:callback.len])
+    next_sign = @view(integrator.callback_next_sign[1:callback.len])
+  else
+    prev_sign = 0.0
+    next_sign = 0.0
+  end
 
-  if integrator.event_last_time == counter && ODE_DEFAULT_NORM(previous_condition,integrator.t) < 100ODE_DEFAULT_NORM(integrator.last_event_error,integrator.t)
+  if integrator.event_last_time == counter && minimum(ODE_DEFAULT_NORM(previous_condition[ivec],integrator.t)) < 100ODE_DEFAULT_NORM(integrator.last_event_error,integrator.t)
```
ChrisRackauckas: view?

kanav99: ivec is an integer so views is not useful here

kanav99:
So I get a strange result:
```
using OrdinaryDiffEq, BenchmarkTools

function f(u,p,t)
	u
end

u0 = 1.0
tspan = (0.0,1.0)

prob = ODEProblem(f,u0,tspan)

function condition(idxs,u,t,integrator)
	idxs[1] = u - 1.5
	idxs[2] = u - 2.1
end

function condition2(u,t,integrator)
	u - 1.2
end

affect2!(integrator) = println("hello")

function affect!(integrator,idx)
	println(idx)
end


cb = DiffEqBase.VectorContinuousCallback(condition,affect!,2)
cb2 = ContinuousCallback(condition2,affect2!)
cbs = CallbackSet(cb,cb2)
@btime sol = solve(prob,Tsit5();callback=cbs)
```
Using keys -    170.947 μs (1158 allocations: 34.48 KiB)
Using collect - 145.917 μs (819 allocations: 34.19 KiB)

Why is the allocations on using keys (latest commit) more?

src/callbacks.jl
ChrisRackauckas: I see

ChrisRackauckas:
That is odd, I'm not sure. But using an iterator of sorts should be better there for sure.

kanav99:
Original ContinousCallback is hit right now. This fixes it upto some extent, but I will try to fix it in the coming hours.

kanav99:
Performance of previous callbacks match now, plus a logical error had also been fixed which caused extra allocations. Can be merged now. I will make a seperate PR for performance fixes. Once this is merged, OrdinaryDiffEq tests need to be rerun on the PR.

ChrisRackauckas:
Can we get a comment documenting what exactly is expected of an integrator for callbacks?

PR #229, opened 2019-06-04
Adds CallbackCache

kanav99:
To make all the callback's temporary registers come in one place, this PR introduces a new structure `CallbackCache`. Changes required in OrdinaryDiffEq are also made in the PR JuliaDiffEq/OrdinaryDiffEq.jl#754. I am also trying to convert `next_sign` and `prev_sign` to `BitArray` but it seems that handling Floating Point error is tricky with this. Also we need to handle the case when condition at time steps is zero differently.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -207,8 +207,8 @@ function get_condition(integrator::DEIntegrator, callback, abst)
   end
   integrator.sol.destats.ncondition += 1
   if callback isa VectorContinuousCallback
-    callback.condition(@view(integrator.callback_cache[1:callback.len]),tmp,abst,integrator)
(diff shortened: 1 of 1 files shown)


src/callbacks.jl
```diff
@@ -628,4 +628,19 @@ function max_vector_callback_length(cs::CallbackSet)
...
+end
+
+mutable struct CallbackCache{conditionType,signType}
+  tmp_condition::conditionType
+  previous_condition::conditionType
+  next_sign::signType
+  prev_sign::signType
+end
+
+function CallbackCache(max_len,conditionType::Type,signType::Type)
```
ChrisRackauckas: It may need to be Type{conditionType}, Type{signType) where {conditionType,signType} to specialize

PR #232, opened 2019-06-05
Fix Newton-Krylov performance

YingboMa:
Needs https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/764


src/nlsolve/newton.jl
```diff
@@ -149,7 +154,11 @@ end
       mul!(vecdz,W,vecztmp) # Here W is actually invW
       @.. vecdz = -vecdz
     else
-      cache.linsolve(vecdz,W,vecztmp,iter == 1 && new_W)
+      if W isa AbstractDiffEqLinearOperator
+        @.. u = uprev+z
+        update_coefficients!(W,u,p,tstep)
+      end
+      cache.linsolve(vecdz,W,vecztmp,iter == 1 && new_W; Pl=ScaleVector(weight, true), Pr=ScaleVector(weight, false), tol=lintol)
```
ChrisRackauckas: Keyword arguments always allowed, and mentioned what's passed down for docs.

src/nlsolve/newton.jl
```diff
@@ -112,9 +113,13 @@ end
 
 @muladd function nlsolve!(nlsolver::NLSolver, nlcache::NLNewtonCache, integrator)
   @unpack t,dt,uprev,u,p,cache = integrator
-  @unpack z,dz,tmp,ztmp,k,κ,c,γ,max_iter = nlsolver
+  @unpack z,dz,tmp,ztmp,k,κ,c,γ,max_iter,weight = nlsolver
   @unpack W, new_W, W_dt = nlcache
   cache = unwrap_cache(integrator, true)
+  if W isa AbstractDiffEqLinearOperator
+    calculate_residuals!(weight, fill!(weight, one(eltype(u))), uprev, u, integrator.opts.abstol, integrator.opts.reltol, integrator.opts.internalnorm, t)
```
ChrisRackauckas: always compute, just in case. The real fix is to put a trait on the linsolve.

PR #233, opened 2019-06-05
ODEFunction accepts both in- and out-of-place styles in function call

jlumpe:
Updated `(f::ODEFunction)()` to work with both in-place and out-of-place arguments, regardless of whether the wrapped function is in place or not. I'm not sure if calling out-of-place functions with the in-place style is a common use case, but I think the reverse definitely is (especially in interactive use). You can now use both types interchangeably without having to know the type parameter (although of course most applications that need optimal performance will still have to specialize).

You could probably easily extend this to the rest of the `DEFunction` subtypes by writing a macro for it.

### DIFF
--- a/src/diffeqfunction.jl
+++ b/src/diffeqfunction.jl
@@ -185,7 +185,30 @@ end
 
 ######### Backwards Compatibility Overloads
 
-(f::ODEFunction)(args...) = f.f(args...)
+# Out-of-place function, out-of-place signature
+function (f::ODEFunction{false})(u, p, t)
+  f.f(u, p, t)
+end
+
+# Out-of-place function, in-place signature
+function (f::ODEFunction{false})(du, u, p, t)
+  du[:] = f.f(u, p, t)
+  nothing
+end
+
+# In-place function, in-place signature
+function (f::ODEFunction{true})(u, p, t)
+  du = similar(u)
+  f.f(du, u, p, t)
+  return du
+end
+
+# In-place function, out-of-place signature
+function (f::ODEFunction{true})(du, u, p, t)
+  f.f(du, u, p, t)
+  nothing
+end
+
 (f::ODEFunction)(::Type{Val{:analytic}},args...) = f.analytic(args...)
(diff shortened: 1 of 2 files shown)


jlumpe:
Forgot to say, the one thing I'm unsure of is what the `eltype` of the returned array should be when calling the in-place function with out-of-place arguments. Currently it's just allocating an uninitialized copy of the `u` argument, but that wouldn't play nicely with ForwardDiff if `eltype(u) == Float64` and `eltype(p) == ForwardDiff.dual`. It might be better to use `promote_type(eltype(u), eltype(p), typeof(t))` or something similar.

kanav99:
I fear this breaks `DynamicalODEFunction`.

ChrisRackauckas:
I reverted this due to the errors. I'm surprised it ends up downstream breaking. Feel free to open this PR again so  we can track it down, but for now we need to get a hotfix in.

jlumpe:
What set of tests do I need to run to check for downstream problems?

kanav99:
Checkout Partitioned Method Test in OrdinaryDiffEq [here](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/algconvergence/partitioned_methods_tests.jl).

jlumpe:
Ok, finally got a chance to take a look at this. The problem seems to be that `DynamicalODEFunction.f1` and `.f2` are themselves instances of `ODEFunction`, but they are expected to have the signature `f([du, ]u, v, p, t)`.

It seems to me that `(::ODEFunction{iip})(...)` should have a fixed signature, but I'm not sure how to fix its use in `DynamicalODEFunction` in a backwards-compatible way. One thing I can think of is to introduce a new type `DynamicalODEComponentFunction` which is basically just like `ODEFunction` but it takes an extra argument.

As far as I can tell (from just looking at `diffeqfunction.jl`) the only other instance where this happens is in `SplitFunction`, but in that case `f1` and `f2` are called with the usual signature so it's not a problem.

ChrisRackauckas:
That fallback is so that dynamical ODE functions can also be used in first order methods. I think we definitely want to keep that, but the design for doing so is a little awkward.

PR #240, opened 2019-06-06
Full immutable return of DESolutions

ChrisRackauckas:


ChrisRackauckas:
Needs upstream tests.

PR #241, opened 2019-06-07
Fix https://github.com/vchuravy/GPUifyLoops.jl/issues/82

YingboMa:


YingboMa:
This doesn't fix the issue, but it fixes https://github.com/vchuravy/GPUifyLoops.jl/issues/82#issuecomment-499674008

PR #246, opened 2019-06-23
monte carlo overhaul

ChrisRackauckas:


ChrisRackauckas:
Causes warnings but no breakage without the next Monte Carlo, so merging.

PR #256, opened 2019-06-30
W transform in Newton iteration

YingboMa:
Ref: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/798


ChrisRackauckas:
Breaking and requires an update of StochasticDiffEq too?

PR #207, opened 2019-04-14
Safer epsilon in callback and handle callback switch off

ChrisRackauckas:
We may want to try/catch using a dual number as well?

PR #257, opened 2019-07-01
Better epsilon handling in callbacks

kanav99:
Lets discuss about how we could actually better our epsilon handling. For those unfamiliar with this - when we hit a callback condition, i.e. `condition = 0` we want the exact time where the event occured. As there is nothing like "exact" in floating point, we calculate time to be the previous nearest floating point time. But when we do this, we are actually going back in time. This could cause infinite loop of events. Right now, we use derivate to calculate the condition in some near future and use the value as the `prev_condition`. But this had failed too (Ref #207).
In this PR, I have added two few tweaks - 
1. Fixed epsilon handling in negetive dt case - because `max` function actually calculates the smaller step (magnitude wise) of the two in case of negetive dt.
2. Buggy time increment in of future. [here](https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/257/files#diff-f2ec50e45e2c0d098a69b4e7c72f86f3R434) if Bottom theta is not 0, then multiplying by 5 doesn't seem right to me. Though a rare case, it could be problematic when `dt` is small.

This needs to be merged with care. @ChrisRackauckas

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -282,6 +282,7 @@ end
 
     # Evaluate condition slightly in future
     abst = integrator.tprev+max(integrator.dt/10000,sign(integrator.dt)*100*eps(integrator.t))
+    abst = integrator.tprev+integrator.tdir*max(abs(integrator.dt/10000),100*eps(integrator.t))
     tmp_condition = get_condition(integrator, callback, abst)
 
     # Sometimes users may "switch off" the condition after crossing
@@ -360,6 +361,7 @@ end
 
     # Evaluate condition slightly in future
     abst = integrator.tprev+max(integrator.dt/10000,sign(integrator.dt)*100*eps(integrator.t))
+    abst = integrator.tprev+integrator.tdir*max(abs(integrator.dt/10000),100*eps(integrator.t))
     tmp_condition = get_condition(integrator, callback, abst)
 
     # Sometimes users may "switch off" the condition after crossing
@@ -429,10 +431,13 @@ function find_callback_time(integrator,callback::ContinuousCallback,counter)
             # But floating point error may make the end point negative
 
             sign_top = sign(zero_func(top_Θ))
-            bottom_θ += 2eps(typeof(bottom_θ))
+            offset = 2eps(typeof(bottom_θ))
+            bottom_θ += offset
             iter = 1
             while sign(zero_func(bottom_θ)) == sign_top && iter < 12
-              bottom_θ *= 5
+              bottom_θ -= offset
+              offset *= 5
+              bottom_θ += offset
               iter += 1
             end
             iter == 12 && error("Double callback crossing floating pointer reducer errored. Report this issue.")
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -429,14 +429,10 @@ function find_callback_time(integrator,callback::ContinuousCallback,counter)
             # But floating point error may make the end point negative
 
             sign_top = sign(zero_func(top_Θ))
-            offset = 2eps(typeof(bottom_θ))
-            bottom_θ += offset
+            bottom_θ += 2eps(typeof(bottom_θ))
             iter = 1
             while sign(zero_func(bottom_θ)) == sign_top && iter < 12
-              bottom_θ -= offset
-              offset *= 5
-              bottom_θ += offset
-              iter += 1
+              bottom_θ *= 5
             end
             iter == 12 && error("Double callback crossing floating pointer reducer errored. Report this issue.")
           end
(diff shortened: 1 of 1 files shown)


src/callbacks.jl
```diff
@@ -282,6 +282,7 @@ end
 
     # Evaluate condition slightly in future
     abst = integrator.tprev+max(integrator.dt/10000,sign(integrator.dt)*100*eps(integrator.t))
```
ChrisRackauckas: why keep the previous line?

src/callbacks.jl
```diff
@@ -360,6 +361,7 @@ end
...
```
ChrisRackauckas: why keep the previous line?

src/callbacks.jl
kanav99: kept that for quick testing, removing them

src/callbacks.jl
```diff
@@ -429,10 +431,13 @@ function find_callback_time(integrator,callback::ContinuousCallback,counter)
             # But floating point error may make the end point negative
 
             sign_top = sign(zero_func(top_Θ))
-            bottom_θ += 2eps(typeof(bottom_θ))
+            offset = 2eps(typeof(bottom_θ))
+            bottom_θ += offset
             iter = 1
             while sign(zero_func(bottom_θ)) == sign_top && iter < 12
-              bottom_θ *= 5
+              bottom_θ -= offset
```
ChrisRackauckas: Why subtract?

Let me describe this a bit. Here is catching a rare case where the endpoint is between 0 and dt/interp_points and the callback derivative overrides the value at 0. So let's say condition(0) is negative and condition(0+) is facing positive, meaning that if the condition had no floating point error it would be 0 and any slightly non-zero theta would make it positive. This means we're looking for the place where it dips below zero in that range, but in order for Roots.jl to work we need to have it be an interval with a positive on the left side, otherwise Roots.jl will error (needs to be a bounding interval). So we push the left until it gets to positive, knowing that a positive value must be right nearby. However, if it keeps pushing and pushing and one isn't there, we error because that shouldn't ever happen. 

Given that's the case, I don't think we need to ever negative offset. That wouldn't fix an error IIUC

ChrisRackauckas: Most likely the error might be using "derivative based" when it's inapplicable, which `abs(zero_func(bottom_θ)) < 100abs(integrator.last_event_error)` should be checking for a `u` change to make sure that it is.

ChrisRackauckas:
Does this fix @lstagner's example?

src/callbacks.jl
kanav99: Okay I missed the fact that we had a check of `prev_sign_index == 1` then the point of having an offset is of no use, I will remove that. But I still have some concerns. If we increment slowly this way in increasing the bottom theta, why dont we use the same way in detecting the event occurance? Why do we just jump a great deal here - https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/callbacks.jl#L284

kanav99:
The example provided was a JLD2 file, so it won't get affected by local changes in DiffEqBase

src/callbacks.jl
kanav99: Sorry I missed your second comment, I do agree you upon that. Also i believe we cannot use the same time incrementer for all the condition functions, it should depend on the nature of function

ChrisRackauckas:
Well this is at least progress. @lstagner will need to run it to make sure it fixes his case.

@JuliaRegistrator register()

lstagner:
Thanks. Initial tests seem to indicate that the issue has been fixed though I can still get an `"Event repeated at the same time. Please report this error"` but it is not fatal and it can be fixed by running the solver non-adaptively.

Using the same input files as before the MWE for the new error is
```julia
@load "equilibrium.jld2"
gcp = GCParticle{Float64}(5.0, 0.13091203265306123, 1.9, -0.2342, 3.344496968928368e-27, 1)
o = get_orbit(MB, gcp, vacuum=true, verbose=true, maxiter=0)
Adaptive
ErrorException("Event repeated at the same time. Please report this error")
Trapped Orbit:
 τₚ = 78.749 μs
 τₜ = 147.666 μs
```

ChrisRackauckas:
At least it's just floating point stuff now. I'm not sure we can solve that well adpatively since 5.0 and 3.34e-27 are more than eps(Float64) apart: it might be just random at these units where that's adaptively resolvable?

lstagner:
I'll look into scaling my inputs so that this wont be a problem but its a low priority since the non-adaptive fallback resolves the issue. Thanks for all your help!

PR #268, opened 2019-07-03
Generalize islinear

devmotion:
This PR generalizes islinear to DiffEqFunctions. Moreover, I updated some outdated documentation of DiffEqOperators and replaced some calls of `A_mul_B!`.

The last two commits are not needed, but I think the name `du_cache` is a bit more appropriate and for consistency with Base and `islinear` `is_constant` could be replaced with `isconstant`.

### DIFF
--- a/src/diffeq_operator.jl
+++ b/src/diffeq_operator.jl
@@ -8,6 +8,7 @@ using LinearAlgebra
    update_coefficients!(A,u,p,t) that changes the internal coefficients, and a
    out-of-place form B = update_coefficients(A,u,p,t).
 3. is_constant(A) trait for whether the operator is constant or not.
+4. islinear(A) trait for whether the operator is linear or not.
 =#
 
 Base.eltype(L::AbstractDiffEqOperator{T}) where T = T
@@ -16,6 +17,7 @@ update_coefficients(L,u,p,t) = L
 
 # Traits
 is_constant(L::AbstractDiffEqOperator) = false
+islinear(::AbstractDiffEqOperator) = false
 has_expmv!(L::AbstractDiffEqOperator) = false # expmv!(v, L, t, u)
 has_expmv(L::AbstractDiffEqOperator) = false # v = exp(L, t, u)
 has_exp(L::AbstractDiffEqOperator) = false # v = exp(L, t)*u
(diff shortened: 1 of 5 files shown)


src/diffeq_operator.jl
```diff
@@ -2,25 +2,27 @@ using LinearAlgebra
...
+3. isconstant(A) trait for whether the operator is constant or not.
+4. islinear(A) trait for whether the operator is linear or not.
 =#
 
 Base.eltype(L::AbstractDiffEqOperator{T}) where T = T
 update_coefficients!(L,u,p,t) = nothing
 update_coefficients(L,u,p,t) = L
 
 # Traits
-is_constant(L::AbstractDiffEqOperator) = false
```
ChrisRackauckas: is this downstream breaking?

src/diffeq_operator.jl
```diff
@@ -42,45 +44,47 @@ has_ldiv!(L::AbstractDiffEqOperator) = false # ldiv!(du, L, u)
...
+expmv!(v,L::AbstractDiffEqLinearOperator,u,p,t) = mul!(v,exp(L,t),u)
 # Factorizations have no fallback and just error
 
 """
 AffineDiffEqOperator{T} <: AbstractDiffEqOperator{T}
 
 `Ex: (A₁(t) + ... + Aₙ(t))*u + B₁(t) + ... + Bₙ(t)`
 
-AffineDiffEqOperator{T}(As,Bs,u_cache=nothing)
+AffineDiffEqOperator{T}(As,Bs,du_cache=nothing)
```
ChrisRackauckas: is this downstream breaking?

src/diffeq_operator.jl
```diff
@@ -118,3 +122,5 @@ function update_coefficients!(L::AffineDiffEqOperator,u,p,t)
     for A in L.As; update_coefficients!(A,u,p,t); end
     for B in L.Bs; update_coefficients!(B,u,p,t); end
 end
+
+@deprecate is_constant(L::AbstractDiffEqOperator) isconstant(L)
```
ChrisRackauckas: got it

src/diffeq_operator.jl
devmotion: It could be downstream breaking if someone is accessing `u_cache` directly, otherwise it should be fine. I just ran the DiffEqOperators tests and they passed. Is there any other package to test?

devmotion: Or should we deprecate it using `getproperty`?

ChrisRackauckas: If DiffEqOperators.jl passes, let's just do it.

PR #271, opened 2019-07-04
Fix missed callback

### LINKED ISSUE #266: Callback issue with condition is zero at tprev
In below code snippet, the comment is counter intuitive because when derivative positive at a zero it is an upcrossing and vice versa. But, in general, we cannot make it up because there is also higher order roots. For my case, `previous_condition` is actually 0 and slight future condition is somewhat negative and next condition is positive. What next is it is treated as event occured and `find_zero` errors out with non-bracketing interval because actual start of the interval is at slight future and `tprev` is not at slight future in `zero_func`
https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/bd02c314adf08d0bb5223ad6e2d33383bb05e902/src/callbacks.jl#L346-L371
https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/bd02c314adf08d0bb5223ad6e2d33383bb05e902/src/callbacks.jl#L417-L420

Also, for avoiding similar error, exact zero at the end of interval checked but not at the beginning.
https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/bd02c314adf08d0bb5223ad6e2d33383bb05e902/src/callbacks.jl#L421-L423

kanav99:
Fixes #266. Actually the problem was that by default `event_last_error` is set to zero. When an event occurs at any timesteps, we do not update `event_last_error` (see [here](https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/callbacks.jl#L421)) So as long as the first non-timstep callback occurs, `event_last_error` is zero, hence the code wont enter [here](https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/callbacks.jl#L346) even if the `prev_condition` was zero. Therefore the callback got missed as we didn't calculate prev_sign in near future.

This code used to throw only a single ping, but is fixed in the branch -
```julia
using OrdinaryDiffEq
function f(u,p,t)
  u
end

condition(u,t,integrator) = (t - 0.5)*(t - 0.6)
affect!(integrator) = println("ping")
cb = ContinuousCallback(condition,affect!)
u0 = 1.0
tspan = (0.0,1.0)
prob = ODEProblem(f,u0,tspan)
sol = solve(prob,Tsit5(),callback=cb,dt=0.1,adaptive=false)
```

@tomaklutfu Please see if this branch fixes your issue. 
@ChrisRackauckas I hope that changing `<` with `<=` won't cause any accuracy issue because of increase of range.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -265,7 +265,7 @@ end
   next_sign = @view(integrator.callback_cache.next_sign[1:callback.len])
 
 
-  if integrator.event_last_time == counter && minimum(ODE_DEFAULT_NORM(previous_condition[ivec],integrator.t)) < 100ODE_DEFAULT_NORM(integrator.last_event_error,integrator.t)
+  if integrator.event_last_time == counter && minimum(ODE_DEFAULT_NORM(previous_condition[ivec],integrator.t)) <= 100ODE_DEFAULT_NORM(integrator.last_event_error,integrator.t)
 
     # If there was a previous event, utilize the derivative at the start to
     # chose the previous sign. If the derivative is positive at tprev, then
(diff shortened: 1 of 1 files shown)


tomaklutfu:
Can I privately share the code to you? This does not solve my actual issue.

tomaklutfu:
Also it seems to me that what new comments says is not in the logic
`tmp_condition > previous_condition` means positive derivative but it still sets it to 1.0 and vice versa.
https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/f131bee44db242a2013a14ddea4126d830d6d633/src/callbacks.jl#L368-L370

ChrisRackauckas:
We should make that an upstream test.

PR #272, opened 2019-07-04
Expand iterative solver support

ChrisRackauckas:
@YingboMa I think we can't do the stationary methods because they can't do the preconditioners?


YingboMa:
I don't think we can do stationary methods either. They are likely to fail on the majority of the problems.

PR #276, opened 2019-07-06
Replace Winv with Wfact

ChrisRackauckas:


ChrisRackauckas:
Indeed. I think it's just Rosenbrock23 and Rosenbrock32 that use the untransformed version? Need to double check, but it would be good to complete that.

PR #278, opened 2019-07-06
Yingbo's allocation reduction suggestion for CuArray linear solve

ChrisRackauckas:
We really need some GPU tests


ChrisRackauckas:
@YingboMa `ldiv!` is missing for CuArrays.

PR #284, opened 2019-07-08
Concretize function overloads

devmotion:
I guess there's no harm in making the function overloads more concrete.

However, since the documentation at http://docs.juliadiffeq.org/latest/features/performance_overloads.html seems both incomplete and outdated, it's very likely that some signatures are incorrect.

### DIFF
--- a/src/diffeqfunction.jl
+++ b/src/diffeqfunction.jl
@@ -190,61 +190,92 @@ end
 
 ######### Backwards Compatibility Overloads
 
-(f::ODEFunction)(args...) = f.f(args...)
-(f::ODEFunction)(::Type{Val{:analytic}},args...) = f.analytic(args...)
-(f::ODEFunction)(::Type{Val{:tgrad}},args...) = f.tgrad(args...)
-(f::ODEFunction)(::Type{Val{:jac}},args...) = f.jac(args...)
-(f::ODEFunction)(::Type{Val{:Wfact}},args...) = f.Wfact(args...)
-(f::ODEFunction)(::Type{Val{:Wfact_t}},args...) = f.Wfact_t(args...)
-(f::ODEFunction)(::Type{Val{:paramjac}},args...) = f.paramjac(args...)
+for F in (:ODEFunction, :SDEFunction)
+  @eval begin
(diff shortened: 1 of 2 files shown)


devmotion:
Just noticed that this won't work since `DynamicalODEFunction` converts `f1` and `f2` to an `ODEFunction`. I'm wondering whether that's reasonable at all? It seems in all other use cases an `ODEFunction` has the signature `(du, u, p, t)` (if it is in-place) and `(u, p, t)` if it is out-of-place.

src/diffeqfunction.jl
```diff
@@ -190,61 +190,92 @@ end
...
-  f.f1(du.x[1],u.x[1],u.x[2],p,t)
-  f.f2(du.x[2],u.x[1],u.x[2],p,t)
+for F in (:ODEFunction, :SDEFunction)
+  @eval begin
+    (f::$F{false})(u, p, t) = f.f(u, p, t)
+    (f::$F{true})(du, u, p, t) = f.f(du, u, p, t)
+
+    (f::$F)(::Type{Val{:analytic}}, u0, p, t) = f.analytic(u0, p, t)
+
+    (f::$F{false})(::Type{Val{:tgrad}}, u, p, t) = f.tgrad(u, p, t)
```
ChrisRackauckas: we've already deprecated and removed these overloads. I don't think we want to support this interface anymore.

src/solutions/dae_solutions.jl
```diff
@@ -27,7 +27,7 @@ function build_solution(
   if has_analytic(prob.f)
     u_analytic = Vector{typeof(prob.u0)}()
     for i in 1:size(u,1)
-      push!(u_analytic,prob.analytic(prob.u0,t[i]))
+      push!(u_analytic, prob.analytic(prob.du0, prob.u0, prob.p, t[i]))
```
ChrisRackauckas: this should probably get its own PR so we can merge it.

ChrisRackauckas:
I think the right way to go is to just delete these overloads: nothing should be using them anymore and they haven't been in the docs for about 2 years.

PR #288, opened 2019-07-10
[WIP] Make nlsolve thread safe

kanav99:
For this to be thread safe, we need to not edit any values from integrator. Right now we use `u` from integrator hence this is not thread safe. And having an extra `tmp_u` would hit all the other solvers. Should we have a new custom `ThreadedNLSolver`? This has 2 benefits, first - we don't play around the previously stable methods hence not breaking them. Secondly - we can have higher customization for such NLSolver, because in parallel methods, it's not necessary that you have 5 different `W` for 5 different threads.

### DIFF
--- a/src/nlsolve/newton.jl
+++ b/src/nlsolve/newton.jl
@@ -161,7 +161,7 @@ end
       @.. u = uprev+z
       update_coefficients!(W,u,p,tstep)
     end
-    cache.linsolve(vecdz,W,vecztmp,iter == 1 && new_W; Pl=ScaleVector(weight, true), Pr=ScaleVector(weight, false), tol=lintol)
+    nlsolver.linsolve(vecdz,W,vecztmp,iter == 1 && new_W; Pl=ScaleVector(weight, true), Pr=ScaleVector(weight, false), tol=lintol)
 
     if has_destats(integrator)
       integrator.destats.nsolve += 1
--- a/src/nlsolve/newton.jl
+++ b/src/nlsolve/newton.jl
@@ -145,8 +145,8 @@ end
     end
 
     # evaluate function
-    @.. u = tmp + γ*z
-    f(k, u, p, tstep)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
>Right now we use u from integrator hence this is not thread safe.

But we don't write to `u` at all? No `nlsolver` should be updating it directly. Instead,  `@.. u = tmp + γ*z` should probably just be made `@.. dz = tmp + γ*z` and then we'd do `f(k, dz, p, tstep)` and `@.. dz = uprev+z`. It looks like that wouldn't change the cache size and is still safe.

kanav99:
Yeah that's a really great idea. Also, do we need `u` in calculating residuals?

ChrisRackauckas:
You do, but it doesn't need to be "u"

kanav99:
I will run OrdinaryDiffEq tests once, then it can be merged. We dont need a new nlsolver if this passes :tada:

ChrisRackauckas:
@yingboma take a look at the calculate residuals here. It's not using the updated u, which might actually be a good thing since it was using an incorrect max in early iterations and when diverging, so using the one from the previous stage might be more stable. It's at least worth a try

ChrisRackauckas:
@yingboma take a look at the calculate residuals here. It's not using the updated u, which might actually be a good thing since it was using an incorrect max in early iterations and when diverging, so using the one from the previous stage might be more stable. It's at least worth a try

ChrisRackauckas:
For consistency in the results, we may want to do similar changes to the `u` that is used in the residual of constantcache version

kanav99:
I am just doing the J move in this PR only. Seems like the first milestone is very basic and almost done.

kanav99:
@ChrisRackauckas this PR is good to go. Requesting reviews. Next step would be to pass J in the nlsolve generator in OrdinaryDiffEq.

PR #291, opened 2019-07-13
No need of J in NLNewtonConstantCache

kanav99:
Continuation of #288 I missed this fact. Testing OrdinaryDiffEq before merging.


ChrisRackauckas:
No, we will need J in OOP for the Jacobian reuse strategies.

PR #300, opened 2019-07-21
structured matrix fast paths

ChrisRackauckas:
```julia
using LinearAlgebra, DifferentialEquations
f = (du,u,p,t) -> du .= u ./ t
jac = (J,u,p,t) -> (J[1,1] = 1/t; J[2,2] = 1/t; J)
jp = Diagonal(zeros(2))
fun = ODEFunction(f; jac=jac, jac_prototype=jp)
prob = ODEProblem(fun,ones(2),(1.0,10.0))
solve(prob,Rosenbrock23(autodiff=false))
```

### DIFF
--- a/src/diffeqfunction.jl
+++ b/src/diffeqfunction.jl
@@ -272,7 +272,7 @@ function ODEFunction{iip,true}(f;
                   end
                  end
 
-                 if jac_prototype !== nothing && colorvec = nothing && ArrayInterface.fast_matrix_colors(jac_prototype)
+                 if jac_prototype !== nothing && colorvec == nothing && ArrayInterface.fast_matrix_colors(jac_prototype)
                    _colorvec = ArrayInterface.matrix_colors(jac_prototype)
                  else
                    _colorvec = colorvec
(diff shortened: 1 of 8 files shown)


src/linear_nonlinear.jl
```diff
@@ -58,7 +58,9 @@ function (p::DefaultLinSolve)(x,A,b,update_matrix=false;tol=nothing, kwargs...)
       else
         p.A = lu!(A)
       end
-    elseif typeof(A) <: SparseMatrixCSC
+    elseif typeof(A) <: Union{Tridiagonal,Symmetric,Hermitian}
```
YingboMa: What about `bunchkaufman` for the `Symmetric` and `Hermitian`? I think `SymTridiagonal` is missing.

ChrisRackauckas:
https://github.com/JuliaLang/julia/issues/32722 from the W scaling causes Symmetric and Hermitian to fail.

src/diffeqfunction.jl
```diff
@@ -259,21 +259,30 @@ function ODEFunction{iip,true}(f;
...
+
                  if jac == nothing && isa(jac_prototype, AbstractDiffEqLinearOperator)
                   if iip
                     jac = update_coefficients! #(J,u,p,t)
                   else
                     jac = (u,p,t) -> update_coefficients!(deepcopy(jac_prototype),u,p,t)
                   end
                  end
+
+                 if jac_prototype !== nothing && colorvec == nothing && ArrayInterface.fast_matrix_colors(jac_prototype)
```
devmotion: We could/should use `colorvec === nothing` here.

PR #304, opened 2019-07-31
Include missing test files in runtests.jl

jlumpe:
There were several files in the `test/` directory that were not included in test sets in `test/runtests.jl`. I'm mostly interested in restoring `diffeqfunction_tests.jl` so I can add to it but figured I should do the rest at the same time.

`internal_euler_test.jl` and `traits_tests.jl` both cause errors. Maybe they are obsolete and should be removed entirely but I just commented them out in `runtests.jl` to make it obvious they were not being run.

I also moved "destats_tests.jl" to the `downstream/` directory because it imports `OrdinaryDiffEq`.


ChrisRackauckas:
The traits tests should just be removed. This:

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/test/internal_euler_test.jl#L4-L6

should be updated to the "newer" syntax and use ODEFunction.

jlumpe:
Fixed internal Euler test, also required some fixes in implementation as well. Refactored DiffEqFunction test file and added tests for `SplitFunction` and `DynamicalODEFunction`. This should now catch any problems created by implementing #233 or #284 instead of breaking things downstream.

test/internal_euler_test.jl
```diff
@@ -1,9 +1,11 @@
 using DiffEqBase, DiffEqBase.InternalEuler
 
 # Try it
-ff(t,u) = u
-ff(::Val{:jac}, t, u) = 1
-ff(::Type{Val{:analytic}},t,u0) = u0*exp(t)
+ff = ODEFunction(
+    (u, p, t) -> u,
+    jac=(u, p, t) -> u,
```
devmotion: The jacobian should be `(u, p, t) -> 1`.

ChrisRackauckas:
```suggestion
    jac=(u, p, t) -> 1,
```

jlumpe: Fixed.

jlumpe: Fixed.

PR #315, opened 2019-08-11
Restructure non-linear solvers

devmotion:
This draft reorganizes the fields of `NLSolver` and its caches.

Currently some fields such as the Jacobian or `linsolve` that are only used by the Newton solver are part of the `NLSolver` struct, but IMO they should rather be part of the `NLNewtonCache`. Moreover, some duplication can be removed by adding an `alg` field (similar to the `ODEIntegrator` struct). The resulting type parameter can also be used for dispatching `nlsolve!`, removing the need for the additional `nlcache` input argument. Moreover, most resize operations can be implemented by overloading `Base.resize!`.

The changes in this draft are breaking and require downstream changes in OrdinaryDiffEq (e.g., in `derivative_utils.jl`).

In addition to the changes in this PR, the implementations of `nlsolve!` can probably be restructured similarly to the implementation of `solve!` in OrdinaryDiffEq by adding `init!`, `loopheader!` etc equivalents since the overall structure of all implementations and the convergence/divergence checks are identical.


src/nlsolve/type.jl
```diff
@@ -66,18 +59,22 @@ NLNewton(; κ=1//100, max_iter=10, fast_convergence_cutoff=1//5, new_W_dt_cutoff
 
 # caches
 
-mutable struct NLNewtonCache{W,J,T,C} <: AbstractNLSolverCache
+mutable struct NLNewtonCache{W,J,T,du1Type,ufType,jcType,lsType,uType} <: AbstractNLSolverCache
   new_W::Bool
   W::W
   J::J
   W_dt::T
-  new_W_dt_cutoff::C
```
ChrisRackauckas: @YingboMa is this no longer used?

devmotion: It's still available as `nlsolver.alg.new_W_dt_cutoff` (but it's not used in the current implementation of `nlsolve!`).

devmotion:
Is/should `integrator.cache.atmp` be available for all algorithms? Then we should use it instead of `ztmp` in

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/b1f8a7e288da68f7c498a97eebabb958312edff9/src/nlsolve/functional.jl#L226

devmotion:
> Is/should integrator.cache.atmp be available for all algorithms?

Currently it's not, e.g., `Rosenbrock23Cache` has no field `atmp`. Is this a bug?

devmotion:
Maybe it's possible to make this (almost) non-breaking by using `getproperty` hacks? I'll play around with this tomorrow.

devmotion:
With the latest changes, this PR does not break OrdinaryDiffEq master anymore (but displays many warnings).

I'm not fully satisfied with it though.I think we should get rid of `@getiipnlsolvefields` and `@getoopnlsolvefields` completely, at some point we have to do this breaking change anyway. After https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/874 and https://github.com/JuliaDiffEq/StochasticDiffEq.jl/pull/224 are released, at least it would not break the current release of OrdinaryDiffEq and StochasticDiffEq.

Additionally, if the direct accesses to `new_W_dt_cutoff` are updated in OrdinaryDiffEq and StochasticDiffEq the fields in the cache could be removed again.

ChrisRackauckas:
How about, go directly for the end goal of just making it as clean as possible, and we'll do a breaking update to DiffEqBase. Nothing that is alg-specific should be in the shared cache, the macros shouldn't exist, it should be thread-safe, there shouldn't be alg-specific accesses into the caches, etc.

devmotion:
> alg-specific accesses into the caches

What do you mean?

Generally, I think maybe we should switch to semver package bounds in at least OrdinaryDiffEq and StochasticDiffEq, similar to what's done in HomotopyContinuation. That would make breaking changes in upstream packages a lot easier, since if the version tag is increased appropriately they shouldn't be available for older versions of OrdinaryDiffEq or StochasticDiffEq.

ChrisRackauckas:
Those should be using semvar already, it's just that semvar is broken if things are grabbing internals.

devmotion:
Hmm, I guess, there's some misunderstanding and I should have explained it more clearly. I thought maybe we should use [Caret specifiers](https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1) explicitly or at least no ">=" bounds for at least DiffEqBase in OrdinaryDiffEq and StochasticDiffEq. Currently, ArrayInterface and julia are the only dependencies in OrdinaryDiffEq that are upper bounded, otherwise we just use ">=" in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/893bc163525ca36dc7bd6f89095a1a05a62cc612/Project.toml#L27-L38.

ChrisRackauckas:
Oh yes, we should get everything to 1.0.0 and then take off the >=. 1.0.0 is necessary since otherwise every minor change is considered breaking and gets an upper bound.

src/nlsolve/functional.jl
```diff
@@ -25,12 +25,15 @@ Equations II, Springer Series in Computational Mathematics. ISBN
 978-3-642-05221-7. Section IV.8.
 [doi:10.1007/978-3-642-05221-7](https://doi.org/10.1007/978-3-642-05221-7)
 """
-@muladd function nlsolve!(nlsolver::NLSolver, nlcache::Union{NLFunctionalConstantCache,NLAndersonConstantCache}, integrator)
+@muladd function nlsolve!(nlsolver::NLSolver{<:Union{NLFunctional,NLAnderson},false}, integrator)
```
ChrisRackauckas: use an abstract type?

ChrisRackauckas:
what is this waiting on?

PR #318, opened 2019-08-14
has_{property} functions check property exists

jlumpe:
`has_jac(f::AbstractDiffEqFunction)` and related functions check whether the given property is not `nothing`, but fail when the type lacks the field entirely. I added a check that the property exists (using the existing `__has_{property}` functions). This should make implementing new function types a bit simpler.

I also merged the `extended_functions.jl` file into `diffeqfunction.jl` because the functions it defines are only used there.

I'm pretty certain the `__has_paramderiv` function isn't used anymore either (definitely not in this package, I don't think downstream packages should be using it) so I removed that as well.


src/diffeqfunction.jl
```diff
@@ -750,16 +750,26 @@ DDEFunction(f::DDEFunction; kwargs...) = f
...
 has_invW(f::AbstractDiffEqFunction) = false
-has_analytic(f::AbstractDiffEqFunction) = f.analytic != nothing
-has_jac(f::AbstractDiffEqFunction) = f.jac != nothing
-has_tgrad(f::AbstractDiffEqFunction) = f.tgrad != nothing
-has_Wfact(f::AbstractDiffEqFunction) = f.Wfact != nothing
-has_Wfact_t(f::AbstractDiffEqFunction) = f.Wfact_t != nothing
-has_paramjac(f::AbstractDiffEqFunction) = f.paramjac != nothing
-has_syms(f::AbstractDiffEqFunction) = :syms ∈ fieldnames(typeof(f)) && f.syms != nothing
-has_colorvec(f::AbstractDiffEqFunction) = f.colorvec != nothing
+has_analytic(f::AbstractDiffEqFunction) = __has_analytic(f) && f.analytic != nothing
```
devmotion: We should check `... !== nothing` or `!isnothing(...)`

devmotion:
I think it's good to add this additional check, but I'm wondering if we should remove the `__has_...` methods and just add a `isdefined` check to the `has_...` methods.

It's a bit unclear to me why this simple check needs an "alias". I see it's currently used in the implementation of `Base.convert(::Type{AbstractDiffEqFunction{iip}},f) where iip` - but couldn't all these conversion be implemented with generated functions? The fields of the types are known at compile time, so we could just build all those `isdefined(f, :fieldname)` if-else-checks automatically for all fieldnames of the desired `AbstractDiffEqFunction` :thinking:

ChrisRackauckas:
>I'm pretty certain the __has_paramderiv function isn't used anymore either (definitely not in this package, I don't think downstream packages should be using it) so I removed that as well.

DiffEqSensitivity.jl is using it. In fact, we need to go deeper here an have a vjp (v'*J) analytical solution function for accelerating adjoints as well.

ChrisRackauckas:
>It's a bit unclear to me why this simple check needs an "alias". I see it's currently used in the implementation of Base.convert(::Type{AbstractDiffEqFunction{iip}},f) where iip - but couldn't all these conversion be implemented with generated functions? The fields of the AbstractDiffEqFunction types are known at compile time, so we could just build all those isdefined(f, :fieldname) if-else-checks automatically for all fieldnames of the desired AbstractDiffEqFunction 🤔

It probably could be. It's really just used during `remake` IIRC.

jlumpe:
Used `!==` for comparison to `nothing`. How are you guys feeling about this? It seems like deprecating the `__has_*` functions would be a good idea.

Is there some way to make these repetitive definitions more condensed? I'd use `@eval` in a loop over all the property symbols but I don't that would play nicely with precompilation.

devmotion:
> Used !== for comparison to nothing. How are you guys feeling about this?

We should definitely use `!==`.

> Is there some way to make these repetitive definitions more condensed?

I tried to use `@eval` in this PR: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/319

jlumpe:
> How are you guys feeling about this?

Sorry that wasn't clear, I mean about the PR as a whole. Think it needs more work or is it good to go as-is?

ChrisRackauckas:
Yup, sorry this just got lost. Nicely done, this is a good addition.

PR #319, opened 2019-08-14
Use code generation and deprecate __has_ property functions

devmotion:
Similar to what I speculated about in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/318#issuecomment-521131006. Actually I think probably one should not use generated functions since the function body would have to be generated for every different type of `f` (correct me if that's wrong).


ChrisRackauckas:
This PR is stale. The deprecation has gone on so long though that we can delete these, but they are in SciMLBase now.

PR #322, opened 2019-08-15
Add Pkg dependency and separate Project.toml

devmotion:


devmotion:
Tests fail on Julia 1.1 due to the callback issue which was just fixed.

With the nightlies, however, tests fail due to some error in DynamicPolynomials which was fixed in version 0.2.1 (https://github.com/JuliaAlgebra/DynamicPolynomials.jl/issues/43). It seems we're stuck with DynamicPolynomials v0.1.4 although the current version is 0.3.2.

devmotion:
It seems we're restricted to DynamicPolynomials v0.1.4 since we use HomotopyContinuation v0.7.0, which explicitly specifies `DynamicPolynomials = "^0.1.2"` (https://github.com/JuliaHomotopyContinuation/HomotopyContinuation.jl/blob/8cfd7f0aa886840c6e7d0bc81a16326a420ea435/Project.toml#L23).

ChrisRackauckas:
It's DiffEqBiological

devmotion:
But then why is DynamicPolyniamls v0.3.2 used in the DiffEqBiological test? https://travis-ci.org/JuliaDiffEq/DiffEqBiological.jl/jobs/572073906

At the moment I just assume that some other packages (or DiffEqBase itself) bound some of the dependencies of HomotopyContinuation. Since the package dependencies in HomotopyContinuation are specified in a quite restrictive manner, maybe these different bounds can only be satisfied by HomotopyContinuation v0.7.0.

devmotion:
The same issue occurs if we combine all DiffEq dependencies in DiffEqProblemLibrary: https://travis-ci.org/JuliaDiffEq/DiffEqProblemLibrary.jl/jobs/570298145

devmotion:
Pkg's dependency management can be so confusing :see_no_evil: 

I think I finally figured out what's happening and how to fix it but I still don't fully understand why the latest release of HomotopyContinuation is used in the DiffEqBiological tests but we only get the old version HomotopyContinuation v0.7.0 in the DiffEqBase and DiffEqProblemLibrary tests.

The special thing about HomotopyContinuation v0.7.0 seems to be that it is compatible with Parameters v0.11.0, although both the [Project.toml](https://github.com/JuliaHomotopyContinuation/HomotopyContinuation.jl/blob/v0.7.0/Project.toml) and the [REQUIRE](https://github.com/JuliaHomotopyContinuation/HomotopyContinuation.jl/blob/v0.7.0/REQUIRE) seem to rule out that version 0.11.0. Also weird that version v0.7.1 never shows up although the Project.toml and REQUIRE are identical, apart from the updated version number in the Project.toml.

DiffEqBiological specifies both Parameters and HomotopyContinuation as explicit dependencies in its Project.toml, and hence immediately at the beginning of the tests Pkg tries to find two compatible versions of these packages. We're lucky (or maybe just because in lexicographic order H comes first) and Pkg selects Parameters 0.10.3 and HomotopyContinuation 1.1.0 in https://travis-ci.org/JuliaDiffEq/DiffEqBiological.jl/jobs/572502432.

However, when package conflicts are resolved in DiffEqBase and DiffEqProblemLibrary, Parameters is prioritized (maybe since it's explicitly mentioned in DiffEqBase's Project.toml and DiffEqBase is also the first entry in DiffEqProblemLibrary's Project.toml?). Hence we end up with Parameters v0.11.0, and at the time when we add DiffEqBiological in the downstream tests or as a second dependency in DiffEqProblemLibrary, a quite old version of HomotopyContinuation is selected because all (?) newer versions are incompatible with Parameters v0.11.0.

By explicitly restricting Parameters to version 0.10.3 in DiffEqBase's Project.toml or allowing Parameters 0.11.0 in HomotopyContinuation's Project.toml, the issues are resolved and the tests run successfully (apart from the callback statistics). This seems to confirm my hypothesis.

To conclude, I still don't understand why Pkg thinks that HomotopyContinuation v0.7.0 is compatible with Parameters 0.11.0, but I've opened a PR to HomotopyContinuation that adds support for Parameters 0.11.0.

devmotion:
OK, I give up :smile: 

On the nightlies we now get the newly released version of HomotopyContinuation with support for Parameters v0.11.0 and the tests pass. However, since HomotopyContinuation only supports Julia 1.1 and newer, we still get the old versions of HomotopyContinuation and DynamicPolynomials in the tests with Julia 1.0. So far that seems to be OK since the problem with DynamicPolynomials does not show up on Julia 1.0.

PR #326, opened 2019-08-20
`resize!` from #315

kanav99:
Credits @devmotion. I just took the resize implementation from #315 as I explained in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/325#issuecomment-522927866

### DIFF
--- a/src/nlsolve/utils.jl
+++ b/src/nlsolve/utils.jl
@@ -347,86 +347,99 @@ DiffEqBase.@def getoopnlsolvefields begin
   uf = nlsolver.uf
 end
 
+## resize NLSolver
+
 function nlsolve_resize!(integrator::DEIntegrator, i::Int)
-  if !isdefined(integrator.cache, :nlsolver)
-    return nothing
-  end
-  alg = integrator.alg; nlsolver = integrator.cache.nlsolver
+  (isdefined(integrator.alg, :nlsolve) && isdefined(integrator.cache, :nlsolver)) || return
+
+  nlsolver = integrator.cache.nlsolver
   if nlsolver isa AbstractArray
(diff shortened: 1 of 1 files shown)


devmotion:
As I mentioned in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/325#issuecomment-523217736, I think we should reconsider the order in which we want to add changes. The PR from which the resize! implementation is taken is outdated and should be closed. So IMO it feels wrong to add changes from that PR.

kanav99:
Even though the PR was closed, I dont think considering this resize is bad, its more extendible more clean.

devmotion:
I agree that it's better than the current version but I don't see in what way it would be better than the changes in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/325? On the contrary, I think the changes in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/325/ are superior since `resize!(::NLSolver, ::Int)` does not consider any special cases for `NLAnderson` anymore. Since hopefully the registry will be updated soon with the correctly upper bounded DiffEqBase dependencies for OrdinaryDiffEq and StochasticDiffEq, I think the main goal should be to agree on a final version of https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/325.

IMO, by merging this PR we do not gain much at the moment, it seems it would mostly be helpful for https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/878. I'm pretty sure we would have to rebase https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/325/, which provides similar but slightly improved resize changes.

ChrisRackauckas:
Looks like we are going straight to the newer nlsolvers.

PR #331, opened 2019-09-05
Capture kwargs in problem construction

ChrisRackauckas:
This is a multiple birds in one stone fix. In Pumas I realized it would be easier to use the Ensemble interface if I could add a tstop to the Ensemble. The fact that `callback` was in the problem always kind of annoyed me because it was an odd one out. Using `prob` remake stuff for parameter estimation couldn't handle possibly changing arguments like `d_discontinuities`. etc.

But if a `DEProblem` now captures the `kwargs`, then all of these are handled. In the higher level functions we just splat them in with the other `kwargs` and none of the solvers need to even explicitly handle this. `callback` is then automatically handled because it's a standard common interface argument.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "5.20.1"
+version = "6.0.0"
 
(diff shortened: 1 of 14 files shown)


ChrisRackauckas:
With the change to the nlsolvers update, should this alone be 6.0 @devmotion ?

devmotion:
Yep, I'd suggest aiming for adding the nonlinear solvers in 7.0

PR #332, opened 2019-09-06
Base functionality of StochasticDelayDiffEq

HTSykora:
Added files, types and functions necessary to include StochasticDelayDiffEq.jl in the near future

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -345,10 +345,6 @@ $(TYPEDEF)
 """
 abstract type AbstractSensitivitySolution{T,N} end
 
-"""
-$(TYPEDEF)
-"""
-abstract type AbstractSDDESolution{T,N} <: AbstractRODESolution{T,N} end
 # Misc
 """
 $(TYPEDEF)
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -123,6 +123,18 @@ Base for types which define jump problems.
 """
 abstract type AbstractJumpProblem{P,J} <: DiffEqBase.DEProblem end
 
+"""
+$(TYPEDEF)
+
+Base for types which define SDDE problems.
+"""
+abstract type AbstractSDDEProblem{uType,tType,lType,isinplace,ND} <: DEProblem end
+
+"""
+$(TYPEDEF)
+"""
+abstract type AbstractConstantLagSDDEProblem{uType,tType,lType,isinplace,ND} <:
+                      AbstractSDDEProblem{uType,tType,lType,isinplace,ND} end
 # Algorithms
 """
 $(TYPEDEF)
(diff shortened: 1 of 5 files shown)


src/DiffEqBase.jl
```diff
@@ -324,6 +345,10 @@ $(TYPEDEF)
 """
 abstract type AbstractSensitivitySolution{T,N} end
 
+"""
+$(TYPEDEF)
+"""
+abstract type AbstractSDDESolution{T,N} <: AbstractRODESolution{T,N} end
```
ChrisRackauckas: is this needed?

HTSykora:
Yeah, you are right, you can ditch that, I accidently included it when adding the stuff

ChrisRackauckas:
This seems to be throwing a bunch of method definition warnings.

HTSykora:
Oh, damn, when I only included DiffEqBase it didnt for me, so it slipped my attention... But it will be due to the @add_kwonly macro before the SDDEFunction... I'll look into it tomorrow as soon as I'm able

Sorry about it.

ChrisRackauckas:
No worries! Take your time.

src/problems/sdde_problems.jl
```diff
@@ -0,0 +1,45 @@
...
+  noise::NP
+  constant_lags::lType
+  dependent_lags::lType2
+  callback::C
+  noise_rate_prototype::ND
+  seed::UInt64
+  neutral::Bool
+  order_discontinuity_t0::Int
+  
+  @add_kwonly function SDDEProblem{iip}(f::AbstractSDDEFunction{iip}, g, u0, h, tspan, p = nothing;
```
devmotion:
```suggestion
  @add_kwonly function SDDEProblem{iip}(f::AbstractSDDEFunction{iip}, g, u0, h, tspan, p = NullParameters();
```

src/problems/sdde_problems.jl
```diff
@@ -0,0 +1,45 @@
...
+                                       neutral = f.mass_matrix !== I && det(f.mass_matrix) != 1,
+                                       order_discontinuity_t0 = 0,
+                                       callback = nothing) where {iip}
+    _tspan = promote_tspan(tspan)
+    new{typeof(u0),typeof(_tspan),typeof(constant_lags),typeof(dependent_lags),isinplace(f),
+        typeof(p),typeof(noise),typeof(f),typeof(g),typeof(h),typeof(callback),typeof(noise_rate_prototype)}(
+        f, g, u0, h, _tspan, p, noise, constant_lags, dependent_lags, callback, noise_rate_prototype, seed, neutral, order_discontinuity_t0)
+  end
+
+  function SDDEProblem{iip}(f::AbstractSDDEFunction{iip}, g , h, tspan, p = nothing;
```
devmotion: You need to specify some types, otherwise this method cannot be distinguished from the standard definition above without `p`. I suggest
```suggestion
  function SDDEProblem{iip}(f::AbstractSDDEFunction{iip}, g , h, tspan::Tuple, p = NullParameters();
```
as done for the DDEProblems, although I'm not completely satisfied with this solution since it won't work if `tspan` is parameter-dependent (there's some additional discussion in an open PR, if I remember correctly about some other aspect of this convenience function).

HTSykora:
Thanks for the suggestions, they seem to solve the method definition warnings

PR #339, opened 2019-09-17
check has_Wfact directly on f not on f.f

ChrisRackauckas:


devmotion:
I think that's the correct approach, but apparently we either need to change downstream calls of the linsolve initialization or add `has_Wfact` and `has_Wfact_t` for stuff like `UJacobianWrapper`.

devmotion:
What's the status of this PR? BTW, there are issues with data arrays on Julia nightlies, it seems.

ChrisRackauckas:
I didn't check downstream yet. I might just dedicate my computer to running this later today.

PR #341, opened 2019-09-19
allow for arbitrary array types in CallbackCache

ChrisRackauckas:


src/callbacks.jl
```diff
@@ -637,7 +637,15 @@ mutable struct CallbackCache{conditionType,signType}
   prev_sign::signType
 end
 
-function CallbackCache(max_len,conditionType::Type,signType::Type)
+function CallbackCache(u,max_len,::Type{conditionType},::Type{signType}) where {conditionType,signType}
```
devmotion: What's the purpose of the `where` clauses? IMO all type annotations could/should be removed.

src/callbacks.jl
```diff
@@ -637,7 +637,15 @@ mutable struct CallbackCache{conditionType,signType}
...
+    tmp_condition = similar(u, conditionType, max_len)
+    previous_condition = similar(u, conditionType, max_len)
+    next_sign = similar(u, signType, max_len)
+    prev_sign = similar(u, signType, max_len)
+    CallbackCache{typeof(tmp_condition),typeof(next_sign)}(tmp_condition,previous_condition,next_sign,prev_sign)
+end
+
+function CallbackCache(max_len,::Type{conditionType},::Type{signType}) where {conditionType,signType}
```
devmotion: Same here

src/callbacks.jl
ChrisRackauckas: `::Type` doesn't specialize.

devmotion: But isn't it sufficient to remove the type annotation completely?

ChrisRackauckas: IIRC no, you need to specialize it so it transforms it from `DataType`.

ChrisRackauckas: IIRC, this and how `::Function` works are the two exceptions to the auto-specialization rule.

devmotion: Oh interesting, I didn't know that

devmotion: But wait, isn't it slightly different for `Function` in that if the type annotation is removed, it is actually specialized? E.g., `f(g) = g(0)` should specialize whereas `f(g::Function) = g(0)` doesn't?

ChrisRackauckas: Yes, it's the opposite way with `Function`, with the caveat that `Function` will also specialize there if it's called... two oddly specific and weird heuristics.

PR #344, opened 2019-09-23
Fix VectorContinousCallback

kanav99:
Reviews @ChrisRackauckas, if you want to dig into Callback Code again :stuck_out_tongue:

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -475,7 +475,7 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
         bottom_θ = typeof(integrator.t)(0)
       end
       if callback.rootfind && !isdiscrete(integrator.alg)
-        minΘ = nextfloat(1.0)
+        minΘ = nextfloat(typeof(integrator.t)(1))
         min_event_idx = -1
         for idx in event_idx
           zero_func = (Θ) -> begin
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -475,7 +475,7 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
         bottom_θ = typeof(integrator.t)(0)
       end
       if callback.rootfind && !isdiscrete(integrator.alg)
-        minΘ = nextfloat(typeof(integrator.t)(1))
+        minΘ = nextfloat(top_Θ)
         min_event_idx = -1
         for idx in event_idx
           zero_func = (Θ) -> begin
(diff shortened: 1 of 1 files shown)


src/callbacks.jl
```diff
@@ -475,7 +475,7 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
         bottom_θ = typeof(integrator.t)(0)
       end
       if callback.rootfind && !isdiscrete(integrator.alg)
-        minΘ = 1.0
+        minΘ = nextfloat(1.0)
```
ChrisRackauckas: what is this actually doing? And why only VectorContinuousCallback?

kanav99: For VectorContinuousCallback, we can have multiple events in same interval (in the same interpolation interval too), so we have to find the exact time where the event happened and then find the minimum time. That's what `min_theta` is about. We find the minimum time where an event happens. We started `min_theta` with 1.0. But in the case when the only event that happens is at 1.0, the `theta < min_theta` branch will be missed. So I used the next float to solve this issue.
But before merging this, I want to make sure we haven't already handled this case somewhere else. If we have already handled this, this problem might be of a larger issue.

devmotion: Why not just set it to `Inf`?

ChrisRackauckas: Inf is Float64. We should probably make this match `nextfloat(typeof(integrator.t)(1))` (and this might need a comment)

ChrisRackauckas:
@kanav99 can you confirm that this works with the new downstream tests?

src/callbacks.jl
ChrisRackauckas: yeah that might be better

kanav99:
Sorry I was away today. I get some type conversion errors in CallbackCache, might be due to recent changes. Fixing them

kanav99:
Done https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/938/commits/ac05f7e6a113935542d4c3ce67df069c9f58968e.
I will run full tests now, repeat tests work now.

kanav99:
OrdinaryDiffEq tests pass, DiffEqBase is failing on master itself

kanav99:
Why are Downstream tests only failing for Julia 1?

devmotion:
I'm not sure, I've seen this before in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/339 but assumed that these downstream test errors were some spurious artefacts since the Travis CI push tests still passed (only the PR tests failed). The test failure of the Core test group on Julia nightly is not an artefact, though; the downstream test failures on Julia nightly should be fixed by https://github.com/JuliaMath/Quadmath.jl/pull/40.

PR #345, opened 2019-09-24
WIP add krylovkit backend

ChrisRackauckas:
@rveltz


ChrisRackauckas:
@YingboMa could the scaling be done before the call? That would at least give KrylovKit and GPU GMRES for now.

rveltz:
You mean like in [here](https://github.com/rveltz/PseudoArcLengthContinuation.jl/blob/master/src/LinearSolver.jl#L145) ?

ChrisRackauckas:
yup, that's it.

PR #347, opened 2019-10-07
fix complex adaptive norms on GPU

ChrisRackauckas:


ChrisRackauckas:
https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/460

Now it works, but still throws the error. MWE:

```julia
using CuArrays
f(u,nothing,t) = 1/2 .*u
u0 = cu(rand(32,32).+ 1im*rand(32,32));
using OrdinaryDiffEq
prob = ODEProblem(f,u0,(0.0,1.0))
sol = solve(prob,Tsit5())
```

```
┌ Warning: calls to Base intrinsics might be GPU incompatible
│   exception =
│    You called hypot(x::T, y::T) where T<:Number in Base.Math at math.jl:541, maybe you intended to call hypot(x::Float64, y::Float64) 
in CUDAnative at C:\Users\accou\.julia\packages\CUDAnative\UWBIY\src\device\cuda\math.jl:291 instead?
│    Stacktrace:
│     [1] hypot at math.jl:541
│     [2] #23 at C:\Users\accou\.julia\packages\GPUArrays\fLiQ1\src\broadcast.jl:49
└ @ CUDAnative C:\Users\accou\.julia\packages\CUDAnative\UWBIY\src\compiler\irgen.jl:116
```

@maleadt maybe that warning doesn't need to exist?

maleadt:
> @maleadt maybe that warning doesn't need to exist?

We assume Base intrinsics are incompatible if there's a CUDAnative alternative, and in this case `CUDAnative.hypot` exists. Does `Base.hypot` just work on the GPU? If so, maybe we should add a whitelist of supported intrinsics to CUDAnative.

ChrisRackauckas:
Yeah, it looks like it just needs to be whitelisted:

```julia
using CuArrays, LinearAlgebra
x = rand(Float32) + rand(Float32)*im
norm.(cu([x])) == norm.([x])
hypot.(cu([real(x)]),cu([imag(x)])) == hypot.([real(x)],[imag(x)])
```

PR #349, opened 2019-10-10
Use optimized findall on GPU

ChrisRackauckas:
Requires https://github.com/JuliaGPU/CuArrays.jl/pull/446

### DIFF
--- a/src/init.jl
+++ b/src/init.jl
@@ -136,6 +136,14 @@ function __init__()
     function LinearAlgebra.ldiv!(x::CuArrays.CuArray,_qr::CuArrays.CUSOLVER.CuQR,b::CuArrays.CuArray)
       _x = UpperTriangular(_qr.R) \ (_qr.Q' * reshape(b,length(b),1))
       x .= vec(_x)
+      unsafe_free!(_x)
+    end
+    function findall_events(affect!,affect_neg,prev_sign::CuArrays.CuArray,next_sign::CuArrays.CuArray)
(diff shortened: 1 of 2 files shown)


src/init.jl
```diff
@@ -136,6 +136,16 @@ function __init__()
...
       x .= vec(_x)
+      CuArrays.CUDAnative.unsafe_free!(_x)
+    end
+    function findall_events(affect!,affect_neg!,prev_sign::CuArrays.CuArray,next_sign::CuArrays.CuArray)
+      hasaffect::Bool = affect! !== nothing
+      hasaffectneg::Bool = affect_neg! !== nothing
+      f = (p,n)-> ((p < 0 && hasaffect) || (p > 0 && hasaffectneg)) && p*n<=0
+      A = map(f,prev_sign,next_sign)
+      out = findall(A)
+      CuArrays.CUDAnative.unsafe_free!(A)
```
ChrisRackauckas: @maleadt where is `unsafe_free!` defined?

maleadt: CuArrays

maleadt: It's a feature of the memory pool, eagerly putting stuff you know will be unused back into the pool before waiting for the GC to catch up (and because calling `finalize` is too slow).

ChrisRackauckas: I see, thanks. Seems like it could be very useful for this case, where the normal code is non-allocating but here I do have to pay the price of an allocation just to get the GPU code working.

ChrisRackauckas: The other path is using iteration on the indices in order to `findall` on an iterator. `findall` doesn't allow giving two arguments, other than by `zip`. I think it's just a missing Base abstraction that's causing the difficulty here.

PR #367, opened 2019-10-31
Remove `isconstant` export

YingboMa:


ChrisRackauckas:
Make sure this doesn't break DiffEqOperators

PR #371, opened 2019-11-09
Add required ldiv! overloads

ChrisRackauckas:
https://github.com/JuliaLang/julia/issues/31278

@antoine-levitt


antoine-levitt:
Not sure how that is going to interact with nlsolve. Why not put those lines in base?

ChrisRackauckas:
sure, do it.

PR #388, opened 2019-11-23
Allowing solve.jl to merge user defined callbacks with those defined …

PiotrSokol:
…in the problem.

As discussed on Slack, this PR allows `_prob`-defined callbacks to be merged with those supplied by the user.
The merge is done by defining a `CallbackSet` comprised of both the `_prob` callbacks and the user defined callbacks.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -35,6 +35,12 @@ end
 
 function solve_call(_prob,args...;kwargs...)
   if :kwargs ∈ propertynames(_prob)
+    if haskey(_prob.kwargs,:callback) && haskey(kwargs, :callback)
+      kwargs_temp = NamedTuple{Base.diff_names(Base._nt_names(
+      values(kwargs)), (:callback,))}(values(kwargs))
+      callbacks = NamedTuple{(:callback,)}( [DiffEqBase.CallbackSet(_prob.kwargs.callback, values(kwargs).callback )] )
+      kwargs = merge(kwargs_temp, callbacks)
+    end
     __solve(_prob,args...;_prob.kwargs...,kwargs...)
   else
     __solve(_prob,args...;kwargs...)
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -33,9 +33,9 @@ function init(prob::DEProblem,args...;kwargs...)
   end
 end
 
-function solve_call(_prob,args...;kwargs...)
+function solve_call(_prob,args...;merge_callbacks = true,kwargs...)
   if :kwargs ∈ propertynames(_prob)
-    if haskey(_prob.kwargs,:callback) && haskey(kwargs, :callback)
+    if merge_callbacks && haskey(_prob.kwargs,:callback) && haskey(kwargs, :callback)
       kwargs_temp = NamedTuple{Base.diff_names(Base._nt_names(
       values(kwargs)), (:callback,))}(values(kwargs))
       callbacks = NamedTuple{(:callback,)}( [DiffEqBase.CallbackSet(_prob.kwargs.callback, values(kwargs).callback )] )


src/solve.jl
```diff
@@ -35,6 +35,12 @@ end
 
 function solve_call(_prob,args...;kwargs...)
   if :kwargs ∈ propertynames(_prob)
+    if haskey(_prob.kwargs,:callback) && haskey(kwargs, :callback)
```
ChrisRackauckas: I'm not sure we actually want to force this argument to work differently from the others? If we do, it should have a kwarg for it.

PiotrSokol: what would you suggest? The convenience of this approach is that adding another callback to the set does not necessarily change the solution, and the elements of the set are not mutually exclusive.

ChrisRackauckas: Add a keyword argument like `merge_callbacks=false` by default. It seems weird to have just one keyword argument act differently from the others, so making it switchable is fine.

PiotrSokol: Ok, cool! Where do you suggest that kwarg be defined? One option would be to put 
```
if haskey(_prob.kwargs,:callback) && haskey(kwargs, :callback) && get(kwargs, :merge_callbacks, false)
```
but I'm not sure if solve scans for valid kwargs somewhere else.

ChrisRackauckas: it at least doesn't scan for valid kwargs right now. It would be very very difficult to do correctly (though it would be nice to add) and highly breaking.

PiotrSokol: You mean scanning for valid kwargs?
Either way, thoughts on having that check in `solve_call`?

ChrisRackauckas: >You mean scanning for valid kwargs?

yes

>Either way, thoughts on having that check in solve_call?

looks good to me

PiotrSokol: o7, will do. btw, I think having this merging behavior as the default for adjoint sensitivity problems might be good, since the adjoints heavily rely on callbacks.
Do you think that would be useful? And if so should it be a kwarg at the level of `adjoint_sensitivities_u0` and `adjoint_sensitivities` in [DiffEqSensitivities](https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/blob/master/src/adjoint_sensitivity.jl)?

ChrisRackauckas: `adjoint_sensitivities` should at least merge it with any callback sent to `solve` since that's a "secret" callback. I guess it'll confuse people about ones from `prob` as well... maybe the only sensible default is to merge then? @tkoolen might have opinions on this, but we need to settle this, pick one default that usually works, and then just have an option to flip it here.

ChrisRackauckas: But yes, this high level is where to capture it so every package acts uniformly.

PiotrSokol: Ok, I'll open a PR for `DiffEqSensitivities.jl` and propose to describe how those are computed for discrete targets -- that should explain why there are callbacks defined in `prob` and explain why the merging of callbacks is there by default. I think that should be described clearly in the documentation, but I believe that sits under `DifferentialEquations`.

I can add the documentation there, but could do so only after Dec 1st.

ChrisRackauckas: no worries. We'll pick it back up then. It's most important that we do this right.

ChrisRackauckas:
Not merging callbacks seems to have been a disaster, so I'm liking this more now. I'll add in the kwarg and finish this up.

ChrisRackauckas:
Pumas had an issue, DiffEqSensitivity had an issue, etc... it seems the only sensible default is to merge callbacks, since if some higher level package wants to add a callback it can screw up what the user gave far too easily. This is a little breaking, but since it's a revert of something we did before, I'm going to let it slide through and hopefully no one complains too much. If anything, this direction is safe, since the people who wanted merging now just merge before `solve`, which is a workaround that's no longer necessary but won't break. Hopefully nothing relied on the callbacks being different, but after giving it a try... it's such a weird behavior and I'm not sure anyone could build sensible generic code with that.

PR #391, opened 2019-11-28
Fix fastpow zero limit

ChrisRackauckas:
A simple fix for the zero case.

### DIFF
--- a/src/fastpow.jl
+++ b/src/fastpow.jl
@@ -83,5 +83,11 @@ end
 """
     fastpow(x::Real, y::Real) -> Float32
 """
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Not sufficient to fix the Pumas tests, but probably good to do anyways? Branch prediction should essentially make this free.

src/fastpow.jl
```diff
@@ -83,5 +83,13 @@ end
 """
     fastpow(x::Real, y::Real) -> Float32
 """
-@inline fastpow(x::Real, y::Real) = _exp2(Float32(y) * fastlog2(convert(Float32, x)))
+@inline function fastpow(x::Real, y::Real)
+    if x == zero(x)
```
devmotion:
```suggestion
    if iszero(x)
```

PR #396, opened 2019-12-02
fastpow should just be abstractfloat

ChrisRackauckas:


ChrisRackauckas:
Actually, it'll make sense to work on integers. Good test.

PR #406, opened 2019-12-22
add jvp,vjp

ChrisRackauckas:
Not only is encoding the `jvp` into the Jacobian type a little odd, it also makes it fairly hard to document and define `vjp`s, and then use them in sensitivity analysis. So I was thinking that we might as well leave `jac_prototype` just for the Jacobian type definition, and then define `jvp` separately. This removes the requirement of having it as a `DiffEqOperator` and all of that stuff that seems to be getting in the way. Just a function `jvp(out,v,u,p,t) = f'(u)*v` and `vjp(out,v,u,p,t) = v'*f'(u)` for things like ModelingToolkit to generate and for DiffEqSensitivity to use. 

As for `Wvp`, I think that we can still generate that `WOperator`, which just uses a `jvp` call in there. The mass matrix would then still have to be handled separately, but I don't see a major issue with that? Maybe down the line add a function, maybe, but I'm not convinced yet.

@yingboma


ChrisRackauckas:
one question I've been thinking about is, should we hoist preconditioner and preconditioner setup to this level?

PR #407, opened 2019-12-25
add Zygote and ChainRules adjoint rules to the ODEFunction

ChrisRackauckas:
@YingboMa I realized that this would make it fairly clean and easy for the DiffEqSensitivity implementations, since now it doesn't have to have a branch to handle the existence of a vjp but is just built into the `autojacvec` implementation once it's Zygote.

@oxinabox this commented part is what I mean though, where in the sensitivity analysis implementations I want to calculate `df/du *v`for `f(u,p,t)`, but there doesn't seem an idea for that Jacobian-vector product in ChainRules. It's not issue though an easy to work around, but food for thought.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "6.10.0"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
I realized it probably makes a lot more sense for ModelingToolkit.jl to generate the full combined forward+pullback calculation so it can reuse values. I'll change it towards that.

PR #411, opened 2020-01-03
add ZygoteRule forwarding on vjps

ChrisRackauckas:


ChrisRackauckas:
Fixes https://github.com/FluxML/ZygoteRules.jl/issues/7#issuecomment-570564246

Issue #410, opened 2020-01-03
Type piracy turned into method overwriting


KristofferC:
After https://github.com/JuliaLang/julia/pull/33806, loading this package gives:

```
WARNING: Method definition ldiv!(AbstractArray{T, N} where N where T, AbstractArray{T, 2} where T, AbstractArray{T, N} where N where T) in module LinearAlgebra at /workspace/srcdir/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/generic.jl:247 overwritten in module DiffEqBase at /home/pkgeval/.julia/packages/DiffEqBase/8fdjX/src/init.jl:6.
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition ldiv!(AbstractArray{T, N} where N where T, LinearAlgebra.Diagonal{T, V} where V<:AbstractArray{T, 1} where T, AbstractArray{T, N} where N where T) in module LinearAlgebra at /workspace/srcdir/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/diagonal.jl:430 overwritten in module DiffEqBase at /home/pkgeval/.julia/packages/DiffEqBase/8fdjX/src/init.jl:7.
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition ldiv!(AbstractArray{T, N} where N where T, LinearAlgebra.Diagonal{T, V} where V<:AbstractArray{T, 1} where T, AbstractArray{T, N} where N where T) in module LinearAlgebra at /workspace/srcdir/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/diagonal.jl:430 overwritten in module DiffEqBase at /home/pkgeval/.julia/packages/DiffEqBase/8fdjX/src/init.jl:7.
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition ldiv!(AbstractArray{T, N} where N where T, AbstractArray{T, 2} where T, AbstractArray{T, N} where N where T) in module LinearAlgebra at /workspace/srcdir/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/generic.jl:247 overwritten in module DiffEqBase at /home/pkgeval/.julia/packages/DiffEqBase/8fdjX/src/init.jl:6.
  ** incremental compilation may be fatally broken for this module **
```

ChrisRackauckas:
JuliaRegistrator register()

KristofferC:
Missed the `@`

[...]

PR #412, opened 2020-01-03
remove piracy overload on 1.4

ChrisRackauckas:
@KristofferC 1.4 is the correct bound?

### DIFF
--- a/src/init.jl
+++ b/src/init.jl
@@ -1,10 +1,12 @@
 value(x) = x
 cuify(x) = error("To use LinSolveGPUFactorize, you must do `using CuArrays`")
 
-# Piracy, should get upstreamed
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Fixes https://github.com/JuliaDiffEq/DiffEqBase.jl/issues/410

KristofferC:
Not really, it is better to use `contrib/commit-name.sh` to get the exact version.

ChrisRackauckas:
I'll find out how to run that some day, but at least this is better than nothing.

ChrisRackauckas:
The change is at least going into 1.4 and not 1.5 or a 1.3.x, right?

KristofferC:
This will still error on all RCs now.

PR #413, opened 2020-01-03
Better bound

ChrisRackauckas:
Better bound than https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/412 . Thanks @KristofferC and @thofma !

### DIFF
--- a/src/init.jl
+++ b/src/init.jl
@@ -1,7 +1,7 @@
 value(x) = x
 cuify(x) = error("To use LinSolveGPUFactorize, you must do `using CuArrays`")
 
-if VERSION < v"1.4"
+if VERSION < v"1.4.0-DEV.635"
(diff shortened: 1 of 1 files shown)


src/init.jl
```diff
@@ -1,7 +1,7 @@
 value(x) = x
 cuify(x) = error("To use LinSolveGPUFactorize, you must do `using CuArrays`")
 
-if VERSION < v"1.4"
+if VERSION < v"1.4.0-DEV.635"
```
asinghvi17: Might want to add a `@static` here?  Version change will cause precompile anyway.

YingboMa: No, one doesn't need `@static` for statements on the top-level.

andreasnoack:
Would be great with a bugfix release with this commit

PR #423, opened 2020-01-21
Simpler Tracker overloads

ChrisRackauckas:
Possibly could help it GPU better


ChrisRackauckas:
Supersceded by https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/430

PR #426, opened 2020-01-23
Add Function Wrappers from DiffEqDiffTools

kanav99:
These wrappers just got removed from FiniteDiff


sethaxen:
This PR seems to have broken a number of passing tests on Manifolds.jl, but only on certain builds: fails on Julia 1.0 for os x, linux, and windows and on Julia 1.4 for linux only. For Julia 1.3 and 1.5, all builds pass.

This came up on this PR: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/426. The failing test is https://github.com/JuliaNLSolvers/Manifolds.jl/blob/a06ed1e586e807949e460c672cf7818a3a30069d/test/metric.jl#L44, and the relevant ODE is specified here: https://github.com/JuliaNLSolvers/Manifolds.jl/blob/115b3532b21892a1a4e273d98c35f99a883666b2/src/ode.jl. An example failing build is https://travis-ci.org/JuliaNLSolvers/Manifolds.jl/jobs/641184933#L331.

@ChrisRackauckas or @kanav99, could you give any guidance on how to repair this?

kanav99:
Hey  @sethaxen Can you restart your tests? Seems like OrdinaryDiffEq is stuck to an older version but DiffEqBase got upgraded to latest one. I would need to check versions of OrdinaryDiffEq were compatible on Julia 1.0.

ChrisRackauckas:
Yeah, the versioning was weird on this because of the repo rename, but if you restart the tests it should all be good.

sethaxen:
I restarted the tests, and while the test suite is still running, it has already failed with the same error on Julia 1.0. See https://travis-ci.org/JuliaNLSolvers/Manifolds.jl/builds/641255525.

Oh, and my bad, the Julia 1.4 error on Linux is something else entirely, so this appears to just be a 1.0 problem.

kanav99:
Can you tell me the exact version of OrdinaryDiffEq your package is using?

sethaxen:
It looks like all of the failing builds are using OrdinaryDiffEq 5.27.1

ChrisRackauckas:
Most of this should be bounded such that v1.0 isn't effected. I don't know the best way forward here: we need to raise lower bounds and re-tag versions that are v1.0 compatible and suss out exactly where the issue comes from, but in reality v1.0 is essentially unsupported by most packages which makes it hard.

sethaxen:
Yeah, it sounds tricky to suss out. Our package still supports 1.0 because it's an LTS release, but ODE is just an optional dependency for one feature, so we'll change official support for that feature to v1.1 and up and disable these tests for 1.0. Thanks to you both for the quick replies!

PR #431, opened 2020-02-04
Fix warn macro typo (warn -> @warn)

Max-Manning:


ChrisRackauckas:
Thanks! Will do a patch release with this.

PR #438, opened 2020-02-14
Add prob2dtmin

YingboMa:


src/utils.jl
```diff
@@ -234,3 +234,15 @@ end
...
 function unwrap_cache end
+
+# TODO: would be good to have dtmin a function of dt
+prob2dtmin(prob) = prob2dtmin(prob.tspan, one(eltype(prob.tspan)))
+function prob2dtmin(tspan, ::AbstractFloat)
+  t1, t2 = tspan
+  # handle eps(Inf) -> NaN
+  t1f, t2f = map(isfinite, tspan)
+  !t1f && throw(ArgumentError("t0 in the tspan `(t0, t1)` must be finite"))
+  return t1f & t2f ? max(eps(t1), eps(t2)) : eps(t1)
```
ChrisRackauckas: (0.0,Inf) will throw this off

ChrisRackauckas:
```suggestion
  return t1f & t2f ? max(eps(t1), eps(t2)) : max(eps(typeof(t1),eps(t1))
```

YingboMa:
Ummm, I am not sure why would this cause downstream failure.

ChrisRackauckas:
https://travis-ci.org/JuliaDiffEq/DiffEqBase.jl/jobs/650270321#L781

random floating point error.

PR #445, opened 2020-02-24
better handling of array tspan

ChrisRackauckas:
We should either try to automatically convert or throw a nice error message. I think this is a good middle ground. Comments @yingboma?

### DIFF
--- a/src/problems/problem_utils.jl
+++ b/src/problems/problem_utils.jl
@@ -8,6 +8,7 @@ promote_tspan((t1,t2)::Tuple{T,S}) where {T,S} = promote(t1, t2)
 promote_tspan(tspan::Number) = (zero(tspan),tspan)
 promote_tspan(tspan::Nothing) = (nothing,nothing)
 promote_tspan(tspan::Function) = tspan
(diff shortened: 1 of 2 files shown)


src/problems/problem_utils.jl
```diff
@@ -8,6 +8,7 @@ promote_tspan((t1,t2)::Tuple{T,S}) where {T,S} = promote(t1, t2)
 promote_tspan(tspan::Number) = (zero(tspan),tspan)
 promote_tspan(tspan::Nothing) = (nothing,nothing)
 promote_tspan(tspan::Function) = tspan
+promote_tspan(tspan::AbstractArray) = length(tspan) == 2 ? (first(tspan),last(tspan)) : throw(error("The length of tspan must be two (and preferably, tspan should be a tuple, i.e. (0.0,1.0)). If you are trying to include other values for saving reasons, note see the [common solver arguments page](https://docs.juliadiffeq.org/latest/basics/common_solver_opts/) for information on the saving command saveat."))
```
YingboMa:
```suggestion
promote_tspan(tspan::AbstractArray) = length(tspan) == 2 ? (promote(first(tspan),last(tspan))) : throw(error("The length of tspan must be two (and preferably, tspan should be a tuple, i.e. (0.0,1.0)). If you are trying to include other values for saving reasons, note see the [common solver arguments page](https://docs.juliadiffeq.org/latest/basics/common_solver_opts/) for information on the saving command saveat."))
```

We need this to handle
```julia
julia> Real[0, 1.0]
2-element Array{Real,1}:
 0
 1.0
```

PR #446, opened 2020-02-27
Stop using Compat

andreasnoack:
Compat 3 has been hold back by other packages until but with Compat 3, the use of `Compat.TypeUtils` causes test failures in Pumas.


ChrisRackauckas:
Still true after the FunctionWrappers tag?

andreasnoack:
I think the change here suffices. It's because we have a few tests that use `@test_nowarn` so it's only when those tests hit one of the `Compat.TypeUtils` deprecations that we get the failure and I guess that we don't use `@test_nowarn` for any of the tests that end up calling the parts of `FunctionWrappers` that has deprecations.

andreasnoack:
I take that back. When I tested it locally, I'd also downgraded Compat to version 2, so I actually haven't tested if the `FunctionWrappers` tag is needed.

PR #449, opened 2020-03-04
Allso setting both jac_prototype and sparsity

mauro3:
Needed to make the last iteration of https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/1054 work.

### DIFF
--- a/src/diffeqfunction.jl
+++ b/src/diffeqfunction.jl
@@ -27,6 +27,7 @@ struct ODEFunction{iip,F,TMM,Ta,Tt,TJ,JVP,VJP,JP,TW,TWt,TPJ,S,TCV} <: AbstractOD
   jvp::JVP
   vjp::VJP
   jac_prototype::JP
+  sparsity::JP
   Wfact::TW
   Wfact_t::TWt
   paramjac::TPJ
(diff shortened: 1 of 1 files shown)


mauro3:
@ChrisRackauckas: presumably the positional constructors of `ODEFunction` and friends are not part of the public API?  Otherwise this PR would have been breaking.

Also, does this need updates in other DiffEq packages too?

mauro3:
Searching for "sparsity", I found a few needing updates:
- https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/15764865c86d8e258fecd8e9df303f84c965b5e2/src/functionwrapper.jl#L27
- https://github.com/JuliaDiffEq/DiffEqDocs.jl/blob/22ee8499aeb1fd946b6623bf7c40c5ec0a32e668/docs/src/tutorials/advanced_ode_example.md#speeding-up-jacobian-calculations
- https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/9df57b1900df1051f77e111661cd5335a70c79ba/src/derivative_wrappers.jl (several places)

ChrisRackauckas:
@mauro3 I did the DelayDiffEq one: could you take the StochasticDiffEq one? It should be about the same as here. In fact, we should sooner or later set StochasticDiffEq to just use this part of OrdinaryDiffEq, but that's a larger ordeal.

ChrisRackauckas:
yeah the positional constructors are just for `remake`, so it's not public.

mauro3:
There is also DiffEqSensitivity
https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/659059359?utm_medium=notification&utm_source=github_status
with error `LoadError: type ODEForwardSensitivityFunction has no field sparsity`

PR #456, opened 2020-03-13
Try to fix callbacks

kanav99:
The new `find_bracket` feature doesn't work either, in fact it brings an unexpected issue. I think we should think of a new approach to fix this floating point issue.

1. In case of repeated event, currently we check using derivative. If value of condition is small enough and is increasing, we say that sign is +ve on the left hand (similar for decreasing ). But when we find the callback time, we use actual value of condition at some point ahead until it switches sign. I think this is something wrong, for both detection and time calculation we should do the same. Otherwise such errors are bound to happen.

2. We also do need a `OnceContinuousCallback` that only runs once. or maybe a good way to switch off the callback. A part of code can be simplified if we make a way to switch off the callback which is documented.

3. I support complete refactoring of the callback code as this issue might be something else.

I have some private code that had been shared to me by users. Unless all of that is working, this isnt fixed

@ChrisRackauckas @YingboMa @devmotion do you have any thoughts?

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -569,7 +569,8 @@ end
 function bisection(f, tup, tdir; maxiters=100)
   x0, x1 = tup
   fx0x1 = f(x0) * f(x1)
-  (fx0x1 >= zero(fx0x1)) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
+  fzero = zero(fx0x1)
+  (fx0x1 >= fzero) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
   left = x0
   right = x1
   prevfloat_tdir = isone(tdir) ? prevfloat : nextfloat
@@ -577,7 +578,7 @@ function bisection(f, tup, tdir; maxiters=100)
   while true
     iter += 1
     iter == maxiters && error("Maxiters exceeded in bisection. Please report the error in DiffEqBase")
-    f(left) * f(right) >= zero(fx0x1) && error("Unexpected values in bisection. Please report the error in DiffEqBase.")
+    f(left) * f(right) >= fzero && error("Unexpected values in bisection. Please report the error in DiffEqBase.")
     mid = (left + right) / 2
     y = f(mid)
     if iszero(y)
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -571,15 +571,15 @@ function bisection(f, tup, tdir)
   @assert f(x0) * f(x1) < 0.0
   left = x0
   right = x1
-  prevfloat_tdir = (tdir === 1.0 ? prevfloat : nextfloat)
+  prevfloat_tdir = isone(tdir) ? prevfloat : nextfloat
   while true
     @assert f(left) * f(right) < 0.0
     mid = (left + right) / 2
     y = f(mid)
-    if y === 0.0
+    if iszero(y)
       # we are in the region of zero
       mid = prevfloat_tdir(mid)
-      while f(mid) === 0.0
+      while iszero(f(mid))
         mid = prevfloat_tdir(mid)
       end
       return mid
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -568,7 +568,8 @@ end
 # then either f(nextfloat(r)) == 0 or f(nextfloat(r)) * f(r) < 0
 function bisection(f, tup, tdir; maxiters=100)
   x0, x1 = tup
-  (f(x0) * f(x1) >= 0.0) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
+  fx0x1 = f(x0) * f(x1)
+  (fx0x1 >= zero(fx0x1)) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
   left = x0
   right = x1
   prevfloat_tdir = isone(tdir) ? prevfloat : nextfloat
@@ -576,7 +577,7 @@ function bisection(f, tup, tdir; maxiters=100)
   while true
     iter += 1
     iter == maxiters && error("Maxiters exceeded in bisection. Please report the error in DiffEqBase")
-    f(left) * f(right) >= 0.0 && error("Unexpected values in bisection. Please report the error in DiffEqBase.")
+    f(left) * f(right) >= zero(fx0x1) && error("Unexpected values in bisection. Please report the error in DiffEqBase.")
     mid = (left + right) / 2
     y = f(mid)
     if iszero(y)
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -414,7 +414,7 @@ end
   if callback.interp_points!=0
     addsteps!(integrator)
   end
-  Θs = range(typeof(integrator.t)(0), stop=typeof(integrator.t)(1), length=callback.interp_points)
+  ts = range(integrator.tprev, stop=integrator.t, length=callback.interp_points)
   interp_index = 0
   # Check if the event occured
   previous_condition = @views(integrator.callback_cache.previous_condition[1:callback.len])
@@ -471,8 +471,8 @@ end
     interp_index = callback.interp_points
   end
   if callback.interp_points!=0 && !isdiscrete(integrator.alg) && length(prev_sign) != length(event_idx) # Use the interpolants for safety checking
-    for i in 2:length(Θs)
-      abst = integrator.tprev+integrator.dt*Θs[i]
+    for i in 2:length(ts)
+      abst = ts[i]
       new_sign = get_condition(integrator, callback, abst)
       _event_idx = findall_events(callback.affect!,callback.affect_neg!,prev_sign,new_sign)
       if length(_event_idx) != 0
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
>I think this is something wrong, for both detection and time calculation we should do the same.

Why? What's a better scheme?

ChrisRackauckas:
>or maybe a good way to switch off the callback. A part of code can be simplified if we make a way to switch off the callback which is documented.

A good way to switch on and off callbacks is probably the answer.

kanav99:
Okay so I wrote a custom `bisection` (very rough needs refining) but I realized two things:
1. Using `Θs` instead of `ts` was a bad choice because if we change `dt` later on, an if that's too big or too smaller than last step, then `abst = integrator.tprev + integrator.dt*Θ` could do funny stuff. The MWE I had made `dt` too small so basically `abst` completely same during the 12 loop. 
2. Certainly Roots.jl still have some bugs. Changing the `bisection` to `find_zero(...., BisectionExact())` made `event_detection_tests.jl` error.

Its late night here and would continue making it better in the morning.

kanav99:
Running OrdinaryDiffEq test locally. @ChrisRackauckas please check the `bisection` function

src/callbacks.jl
```diff
@@ -560,7 +560,37 @@ end
...
+
+# rough implementation, needs multiple type handling
+# always ensures that if r = bisection(f, (x0, x1))
+# then either f(nextfloat(r)) == 0 or f(nextfloat(r)) * f(r) < 0
+function bisection(f, tup, tdir)
+  x0, x1 = tup
+  @assert f(x0) * f(x1) < 0.0
+  left = x0
+  right = x1
+  prevfloat_tdir = (tdir === 1.0 ? prevfloat : nextfloat)
```
devmotion: What's the reason for not writing
```suggestion
  prevfloat_tdir = isone(tdir) ? prevfloat : nextfloat
```

src/callbacks.jl
```diff
@@ -560,7 +560,37 @@ end
...
+  while true
+    @assert f(left) * f(right) < 0.0
+    mid = (left + right) / 2
+    y = f(mid)
+    if y === 0.0
```
devmotion: Same here, why not
```suggestion
    if iszero(y)
```
?

src/callbacks.jl
```diff
@@ -560,7 +560,37 @@ end
...
+      # we are in the region of zero
+      mid = prevfloat_tdir(mid)
+      while f(mid) === 0.0
```
devmotion: And here?
```suggestion
      while iszero(f(mid))
```

src/callbacks.jl
kanav99: Oh yes, AD test would fail otherwise, my bad

devmotion: The `@assert` statements could be problematic as well, I guess, but I assumed these are mainly for debugging right now?

kanav99:
AD tests still fail after the change

kanav99:
Any help why these AD test are failing? I plugged `@show typeof(...)` in many places inside `bisection`
```julia
function bisection(f, tup, tdir)
  x0, x1 = tup
  @show typeof(x0)
  @show typeof(x1)
  @assert f(x0) * f(x1) < 0.0
  left = x0
  right = x1
  prevfloat_tdir = isone(tdir) ? prevfloat : nextfloat
  while true
    @assert f(left) * f(right) < 0.0
    mid = (left + right) / 2
    @show typeof(mid)
    y = f(mid)
    @show typeof(y)
    if iszero(y)
      # we are in the region of zero
      mid = prevfloat_tdir(mid)
      @show typeof(mid)
      while iszero(f(mid))
        mid = prevfloat_tdir(mid)
        @show typeof(mid)
      end
      return mid
    end
    (left === mid || right === mid #= i dont think second condition would happen =#) && return left
    if sign(y) === sign(f(left))
      left = mid
      @show typeof(left)
    else
      right = mid
      @show typeof(right)
    end
  end
end
```
All of them are duals which means the types are getting inserted well. 

Test code:
```julia
using Test
using OrdinaryDiffEq, Calculus, ForwardDiff

function f(du,u,p,t)
  du[1] = -p[1]
  du[2] = p[2]
end

for x in 0:0.001:5
  called = false
  function test_f(p)
    cb = ContinuousCallback((u,t,i) -> u[1], (integrator)->(called=true;integrator.p[2]=zero(integrator.p[2])))
    prob = ODEProblem(f,eltype(p).([1.0,0.0]),eltype(p).((0.0,1.0)),copy(p))
    integrator = init(prob,Tsit5(),abstol=1e-14,reltol=1e-14,callback=cb)
    step!(integrator)
    solve!(integrator).u[end]
  end
  p = [2.0, x]
  called = false
  findiff = Calculus.finite_difference_jacobian(test_f,p)
  @test called
  called = false
  fordiff = ForwardDiff.jacobian(test_f,p)
  @test called
  @test findiff ≈ fordiff
end
```

kanav99:
I realized its the problem due to branching inside the Bisection Methods. This fails even on `Roots.Bisection`. I think we should make a branch for AD and Non-AD cases.

ChrisRackauckas:
It's sad we can't use a safe one with AD. Is there any way to stabilize a false position method?

ChrisRackauckas:
Actually let's go with this for now, and then in the future change it so that we take over the dual dispatch and utilize the pushforward derivative, which can be directly applied at the solution. The adjoint version can be found in https://math.mit.edu/~stevenj/18.336/adjoint.pdf . We can do something similar (it's just the implicit function theorem).

kanav99:
> It's sad we can't use a safe one with AD. Is there any way to stabilize a false position method?

Problem with AlefeldPotraShi right now is that it throws convergence error when I set tolerance zero, otherwise there is no problem. I will wait for it to be stable.

 > Actually let's go with this for now, and then in the future change it so that we take over the dual dispatch and utilize the pushforward derivative, which can be directly applied at the solution. The adjoint version can be found in https://math.mit.edu/~stevenj/18.336/adjoint.pdf . We can do something similar (it's just the implicit function theorem).

Sounds like a plan. I am running OrdinaryDiffEq test. Most of the important one's have passed. All the new MWEs passed. @HelgavonLichtenstein

src/callbacks.jl
kanav99: I can change them to proper errors. But those are very important assertions, in case of any edge case, we need a solid starting point to debug

kanav99:
Test pass @ChrisRackauckas , should I add some tests for the new bisection function?

ChrisRackauckas:
Yes, just add a few tests so we trust the bisection method, and let's see how this goes.

kanav99:
OrdinaryDiffEq tests running here - https://colab.research.google.com/drive/1okm12KhJIyEhfQYHj-MbDykv0T_rLHXM

kanav99:
All integrator and interface tests pass. Running convergence tests, but they don't have any callbacks. Can you look at the latest commit @ChrisRackauckas @devmotion for any errors?

ChrisRackauckas:
The last test is fine. This looks good to merge then. You happy with it?

kanav99:
Done. It looks good to me to be merged.

src/callbacks.jl
```diff
@@ -568,15 +568,16 @@ end
 # then either f(nextfloat(r)) == 0 or f(nextfloat(r)) * f(r) < 0
 function bisection(f, tup, tdir; maxiters=100)
   x0, x1 = tup
-  (f(x0) * f(x1) >= 0.0) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
+  fx0x1 = f(x0) * f(x1)
+  (fx0x1 >= zero(fx0x1)) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
```
devmotion: Maybe just compute `zero(fx0x1)` once in the beginning instead of recomputing it in the `while`? Although I guess it shouldn't have a severe impact on performance, might be compiled away?

kanav99: I hope that zeros are easily compiled. Still let me do that.

YingboMa: Zero isn't free to compute. Doing manual loop invariant code motion is better.

kanav99:
One more thing, I have set maxiters to 100 right now. iters bisection takes is proportional to the log of time interval. For `x1 - x0 = 25.967830436978204`, it takes 50 iters. So a limit of `maxiters = 100` would be fine?

YingboMa:
Do we have to set `maxiters`? Stiff solvers could potentially step for a very long interval.

ChrisRackauckas:
I don't think we need to set maxiters. 2^100 = 1.2676506e+30, so the default is essentially going until floating point error, unless it doesn't on a problem where people want really high precision, in which case having maxiters doesn't make sense because they want high precision.

kanav99:
I want to set it for now, I have made sure this loop wont be stuck in an infinite loop, but just for safety. I can increase it, but I think we should set it for now, just to prevent infinite loops.

devmotion:
>  I think we should set it for now, just to prevent infinite loops.

Couldn't you check something else like that you actually shrink the bracketing interval to make sure it does terminate at some point? Something that's independent of the number of iteration steps?

kanav99:
I have already added this line - https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/456/files#diff-f2ec50e45e2c0d098a69b4e7c72f86f3R591 which checks if mid is one of the endpoints of bracket, that should probably always check for infinite loops. If you think thats enough, I will remove the maxiters.

kanav99:
Some good news from JuliaMath/Roots.jl#174 too

ChrisRackauckas:
yeah, that's really good news and probably the final solution, but let's at least get this through for now.

PR #476, opened 2020-04-09
Fix bisection to make it not return tprev

kanav99:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/590
The condition function in the issue was such that value of condition at `bottom_t` was -700 and `nextfloat(bottom_t)` was 2000 so our bisection function would have returned `bottom_t`.  There's nothing much we can do in such condition functions, but throwing an error just because there isn't much precision in a very very specific case wasn't good either.
So I made a kwarg which tells if the bisection function can return left most value or not. 

@ChrisRackauckas @BeastyBlacksmith

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -717,7 +717,7 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
               end
               iter == 12 && error("Double callback crossing floating pointer reducer errored. Report this issue.")
             end
-            Θ = bisection(zero_func, (bottom_t,top_t), integrator.tdir)
+            Θ = bisection(zero_func, (bottom_t,top_t), integrator.tdir; noleft=(bottom_t==integrator.tprev))
             if integrator.tdir * Θ < integrator.tdir * min_t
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Won't we still need to take the left value there to enforce manifold constraints? We just need to up the tolerance on the next detection when we see this.

kanav99:
Yes you are right. I am thinking if that error (this one https://github.com/SciML/DiffEqBase.jl/blob/master/src/callbacks.jl#L758) even makes sense in the current scenario or not, now that bisection is guaranteed to return the left value of actual root. I think we should remove that completely?

ChrisRackauckas:
That error still makes sense, because it prevents an infinite loop of always re-finding the same root.

ChrisRackauckas:
that was a mistake

ChrisRackauckas:
@kanav99 reopen if you want to discuss more.

ChrisRackauckas:
If anything, the option should be on the ContinuousCallback side so the user could choose left/right, and default to left.

PR #485, opened 2020-04-20
Fix bisection to make it not return tprev #2

kanav99:
@ChrisRackauckas As you said in the last PR, to make an option in ContinuousCallback for noleft thing, if we did that, we would get the error for `cbtime = 0`

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -567,7 +567,7 @@ end
 # rough implementation, needs multiple type handling
 # always ensures that if r = bisection(f, (x0, x1))
 # then either f(nextfloat(r)) == 0 or f(nextfloat(r)) * f(r) < 0
-function bisection(f, tup, tdir; maxiters=100)
+function bisection(f, tup, tdir; maxiters=100, noleft=false)
   x0, x1 = tup
   fx0x1 = f(x0) * f(x1)
   fzero = zero(fx0x1)
@@ -600,7 +600,7 @@ function bisection(f, tup, tdir; maxiters=100)
         end
       end
     end
-    (left === mid || right === mid) && return left
+    (left === mid || right === mid) && return ((left === tup[1]) && noleft ? right : left)
     if sign(y) === sign(f(left))
       left = mid
     else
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
>if we did that, we would get the error for cbtime = 0

What do you mean?

kanav99:
See we set the noleft flag here https://github.com/SciML/DiffEqBase.jl/pull/485/files#diff-f2ec50e45e2c0d098a69b4e7c72f86f3R653 . See that this is done only if the interval to be solved starting from tprev of the integrator steps.

And then inside the bisection, https://github.com/SciML/DiffEqBase.jl/pull/485/files#diff-f2ec50e45e2c0d098a69b4e7c72f86f3R603 , we only return the right if the actual zero is in between `left` and `nextfloat(left)` because if we returned `left` that is `tprev`, `cb_time = callback time - tprev` inside the apply_callback would be zero and we would get the error https://github.com/SciML/DiffEqBase.jl/blob/master/src/callbacks.jl#L757

kanav99:
This is a very very specific flag, which comes into effect only when the callback goes off between `tprev` and `nextfloat(tprev)`

ChrisRackauckas:
is there a case where this is not am accidental retrigger of the same event?

kanav99:
Nope, though the perfect solution would be to remove the error https://github.com/SciML/DiffEqBase.jl/blob/master/src/callbacks.jl#L757 completely, but I understand that it is important in the sense that it helps us track bugs.

ChrisRackauckas:
No, that wouldn't be a solution. That error is catching infinite loops. It looks like this is changing infinite loops to something that moves 1 floating point value forward at a time, which isn't better?

kanav99:
This PR violates our very original assumption that if a callback happens, integrator should never step ahead of callback time, but step backward if needed. But if we are keeping that error, we have to step forward in the case when callback occurs somewhere between `tprev` and `nextfloat(tprev)` because if we step backward, we would get the error, which should not happen.
This is the tradeoff here: either remove the error or violate the moving backward assumption

ChrisRackauckas:
Well removing the error doesn't fix the issue because then you just get an infinite loop.

Maybe have a field in the callback for enforceleft?

If we add this, then we need a test that describes why we need it.

kanav99:
So if the user set enforceleft, we would need to make that error not happen when this flag is set right? then that would make if go infinite loop again, should we leave it for the user to handle?

ChrisRackauckas:
That error should never be removed. It makes no sense to allow an infinite loop. We already know that if you have a callback at `t`, and then the next step causes a callback at `t`, and that the next step causes a callback at `t`, etc., that the user wants to exit. We should be hitting the callback at exactly `t` only once: it's an error to hit it twice. You don't want to bounce the ball twice just because of numerical error: when the ball hits the ground it bounces once.

kanav99:
Yes you are right, so by default, lets keep the error, but lets allow the user to take wild path of choosing the left one.
Slight doubt: was that error made because we don't trust our differential check branch here https://github.com/SciML/DiffEqBase.jl/blob/master/src/callbacks.jl#L512 ? if we assume that this branch always protects us from repeat events, would that error still makes sense?

PR #487, opened 2020-04-22
Callback Fix: Allowing 0 `cb_time`

kanav99:
Very small changes to code and removed that error.
Scripts tested on:

```julia
using OrdinaryDiffEq, Test

lastt = 0.0
f = function (du,u,p,t)
  du[1] = u[2]
  du[2] = -p[1]
end
function condition(u,t,integrator) # Event when event_f(u,t) == 0
  u[1]
end
function affect!(integrator)
  @test integrator.u[1] >= 0
  global lastt
  println("bounce at interval $(integrator.t - lastt)")
  lastt = integrator.t
  integrator.u[2] = -integrator.u[2]
end
cb2 = ContinuousCallback(condition,affect!)
tspan = (0.0,Float64(Inf))
u0 = [50000.0,0.0]
p = 9.8
prob = ODEProblem(f,u0,tspan,p)
integ = init(prob,Tsit5(),callback=cb2,maxiters=Inf,progress=true)
solve!(inter)
```
This one makes the ball go negative (as dt goes very big) on master. Works fine on this PR(watch out, tspan is infinite long so the script wont stop). Reduced dt after callback and solved an `eps` related blunder.

```julia
f = (u,p,t) -> u

function condition(u,t,i)
    (t-0.10)*(t-0.20)*(t-0.30)*(t-0.40)*(t-0.50)*(t-0.60)*(t-0.70)*(t-0.80)*(t-0.90)
end

function affect!(i)
    println("affect!")
    @show i.t
end
cb = ContinuousCallback(condition,affect!)
tspan = (0.0, 1.0)
u0 = 1.0
prob = ODEProblem(f,u0,tspan)
solve(prob,Tsit5(),callback=cb,adaptive=false,dt=0.1)
```
This one throws the non-bracketing error or the cb_time is 0 error. This PR fixes it too.
@ChrisRackauckas

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -447,7 +447,7 @@ end
     end
 
     # Evaluate condition slightly in future
-    abst = integrator.tprev+integrator.tdir*max(abs(integrator.dt/10000),100*eps(typeof(integrator.t)))
+    abst = integrator.tprev+integrator.tdir*max(abs(integrator.dt/10000),100*eps(integrator.t))
     tmp_condition = get_condition(integrator, callback, abst)
 
     # Sometimes users may "switch off" the condition after crossing
@@ -525,7 +525,7 @@ end
     end
 
     # Evaluate condition slightly in future
-    abst = integrator.tprev+integrator.tdir*max(abs(integrator.dt/10000),100*eps(typeof(integrator.t)))
+    abst = integrator.tprev+integrator.tdir*max(abs(integrator.dt/10000),100*eps(integrator.t))
     tmp_condition = get_condition(integrator, callback, abst)
 
     # Sometimes users may "switch off" the condition after crossing
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -756,9 +756,6 @@ end
 function apply_callback!(integrator,callback::Union{ContinuousCallback,VectorContinuousCallback},cb_time,prev_sign,event_idx)
 
   change_t_via_interpolation!(integrator,integrator.tprev+cb_time)
-  if integrator.opts.adaptive
-    integrator.dtpropose = integrator.dt/10
-  end
 
   # handle saveat
   _, savedexactly = savevalues!(integrator)
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -71,7 +71,7 @@ Contains a single callback whose `condition` is a continuous function. The callb
   if the starting condition is less than the tolerance from zero, then no root will be detected.
   This is to stop repeat events happening just after a previously rootfound event.
 """
-struct ContinuousCallback{F1,F2,F3,F4,T,T2,I} <: AbstractContinuousCallback
+struct ContinuousCallback{F1,F2,F3,F4,T,T2,I,R} <: AbstractContinuousCallback
   condition::F1
   affect!::F2
   affect_neg!::F3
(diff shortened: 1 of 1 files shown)


kanav99:
This or #485 are needed, as described in the comment https://github.com/SciML/DiffEqBase.jl/pull/485#issuecomment-616542934

src/callbacks.jl
```diff
@@ -447,7 +447,7 @@ end
     end
 
     # Evaluate condition slightly in future
-    abst = integrator.tprev+integrator.tdir*max(abs(integrator.dt/10000),100*eps(integrator.t))
+    abst = integrator.tprev+integrator.tdir*max(abs(integrator.dt/10000),100*eps(typeof(integrator.t)))
```
ChrisRackauckas, line 450: why typeof here? Seems better to have it adapting

src/callbacks.jl
```diff
@@ -754,10 +754,11 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
 end
 
 function apply_callback!(integrator,callback::Union{ContinuousCallback,VectorContinuousCallback},cb_time,prev_sign,event_idx)
-  if cb_time == zero(typeof(integrator.t))
-    error("Event repeated at the same time. Please report this error")
-  end
+
   change_t_via_interpolation!(integrator,integrator.tprev+cb_time)
+  if integrator.opts.adaptive
+    integrator.dtpropose = integrator.dt/10
```
ChrisRackauckas, line 760: I think this will fail for Sundials and ODEInterfaceDiffEq

src/callbacks.jl
```diff
@@ -754,10 +754,11 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
 end
 
 function apply_callback!(integrator,callback::Union{ContinuousCallback,VectorContinuousCallback},cb_time,prev_sign,event_idx)
-  if cb_time == zero(typeof(integrator.t))
-    error("Event repeated at the same time. Please report this error")
```
ChrisRackauckas, line 758: how is it correct to remove this and still apply the affect?

src/callbacks.jl
kanav99: hmm I thought so, need to do it for every repository instead

src/callbacks.jl
kanav99: in our repeat detection code, we will never get 0 because we shift `bottom_t` a slight ahead. so in case of repeat detection, there won't be this case.

ChrisRackauckas: Explain that. If we never get to zero, how is this error triggered? It should only be triggered when we're at exactly zero.

kanav99: There are two cases that should be considered:
1. Case of repeated events. In this case, this loop https://github.com/SciML/DiffEqBase.jl/blob/master/src/callbacks.jl#L648 will increase the bottom t and will never be zero.
2. Normal events: In this case, if we use the current setup, if the event occurs between `tprev` and `nextfloat(tprev)` our bisection would return tprev as the root. which is okay, not a problem but that error stops the integration. I say that it's not a problem because even we take a `dt=0` step, next step would be handled due to the loop https://github.com/SciML/DiffEqBase.jl/blob/master/src/callbacks.jl#L648. 
But yes, we need better `dt` handling than this. We don't want dt = 0 for a non callback iteration
What do you suggest I should set dt?

src/callbacks.jl
kanav99: why? we want just ahead of time right? why should that depend on t?

src/callbacks.jl
ChrisRackauckas: How is 1 and 2 interacting? If bottom_t is the next float above 0, then the case of 2 should never hit this error?

ChrisRackauckas: If that's really the case, then (2) should be erroring at https://github.com/SciML/DiffEqBase.jl/blob/master/src/callbacks.jl#L651, right?

src/callbacks.jl
ChrisRackauckas: because we want ahead of time by current floating point epsilon, which changes with the size of t since floating point error is relative.

src/callbacks.jl
kanav99: Nope because it wont enter that branch as `integrator.event_last_time` would not be set in (2) https://github.com/SciML/DiffEqBase.jl/blob/master/src/callbacks.jl#L635

src/callbacks.jl
kanav99: okay I will revert this change

src/callbacks.jl
ChrisRackauckas: Oh I see. Wait... how the hell did we get 1 floating point away from the event in more than one example? That seems insanely improbable!

src/callbacks.jl
```diff
@@ -98,11 +99,13 @@ ContinuousCallback(condition,affect!,affect_neg!;
                    rootfind=true,
                    save_positions=(true,true),
                    interp_points=10,
+                   dtrelax=1.0,
```
devmotion, line 102: I guess it's not a good idea to enforce Float64 here, could you use
```suggestion
                   dtrelax=1,
```
instead?

ChrisRackauckas: Otherwise it will automatically upscale dt to Float64, so it'll cause failures for example if someone was solving with Float32. `1` is much safer.

kanav99: understood, mde the changes

kanav99:
Test pass, that was travis bug

ChrisRackauckas:
Callbacks are fixed and will never have an issue again.

PR #488, opened 2020-04-22
SteadyStateSolution has no field t error

frankschae:
This additional function is necessary for the implementation of  https://github.com/SciML/DiffEqSensitivity.jl/issues/182
where concrete_solve will be called (which currently always returns RecursiveArrayTools.DiffEqArray(sol.u,sol.t).
The steady state solution object however does not have a field t)

Does this need a test somewhere?


ChrisRackauckas:
Yeah we should probably at least add some simple downstream tests to concrete_solve

ChrisRackauckas:
Was missing the test dependency.

PR #491, opened 2020-04-24
Adds is_integrator_adaptive to integrator interface

kanav99:
@ChrisRackauckas this is non-breaking, making a PR to Sundials

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -574,3 +574,10 @@ function step!(integ::DEIntegrator, dt::Real, stop_at_tdt = false)
 end
 
 has_destats(i::DEIntegrator) = false
+
+"""
+    is_integrator_adaptive(i::DEIntegrator)
+
+Checks if the integrator is adaptive
+"""
+is_integrator_adaptive(integrator::DEIntegrator) = integrator.opts.adaptive
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -580,4 +580,5 @@ has_destats(i::DEIntegrator) = false
 
 Checks if the integrator is adaptive
 """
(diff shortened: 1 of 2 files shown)


src/integrator_interface.jl
```diff
@@ -574,3 +574,10 @@ function step!(integ::DEIntegrator, dt::Real, stop_at_tdt = false)
 end
 
 has_destats(i::DEIntegrator) = false
+
+"""
+    is_integrator_adaptive(i::DEIntegrator)
+
+Checks if the integrator is adaptive
+"""
+is_integrator_adaptive(integrator::DEIntegrator) = integrator.opts.adaptive
```
ChrisRackauckas, line 583: if isdefined default

kanav99: I followed the `check_error` approach in this. All repositories have `integrators.opts.adaptive` except Sundials. So I thought this should be the case. Also wouldn't isdefined way cause extra ops?

devmotion, line 583: Maybe one could just use `isadaptive`? I guess, since it is called with `DEIntegrator` it should be clear that it checks if an integrator is adaptive.

ChrisRackauckas: constant propogation should get rid of any cost. Are you sure it's just Sundials? ODEInterfaceDiffEq does events and will need this too. Also, LSODA, DASKR, DASSL, etc. don't have this, but won't immediately error because of no events.

kanav99: Agreed

kanav99: Okay if ops are same, then its good. So is the Sundials PR required now? Should be closed?

ChrisRackauckas:
This should be good.

Issue #507, opened 2020-05-15
Improve API of `DEDataArray` for fields that are `AbstractArray` but immutable


ronisbr:
Hi!

I am trying to create a workspace using `DEDataArray` as follow:

```Julia
mutable struct Workspace{T} <: DEDataVector{T}
    x::Vector{T} = zeros(12)
    q::Quaternion = Quaternion(1,0,0,0)
```

where `Quaternion` is defined in [ReferenceFrameRotations.jl](https://github.com/ronisbr/ReferenceFrameRotations.jl). Here, I get the following error:

```
ERROR: LoadError: setindex! not defined for Quaternion{Float64}
Stacktrace:
 [1] error(::String, ::Type{T} where T) at ./error.jl:42
 [2] error_if_canonical_setindex(::IndexLinear, ::Quaternion{Float64}, ::Int64) at ./abstractarray.jl:1081
 [3] setindex! at ./abstractarray.jl:1072 [inlined]
 [4] copyto!(::Quaternion{Float64}, ::Quaternion{Float64}) at ./multidimensional.jl:962
 [5] recursivecopy!(::Quaternion{Float64}, ::Quaternion{Float64}) at /Users/ronan.arraes/.julia/packages/RecursiveArrayTools/YIS5B/src/utils.jl:25
 [6] macro expansion at /Users/ronan.arraes/.julia/packages/DiffEqBase/8A5uo/src/data_array.jl:60 [inlined]
 [7] recursivecopy!(::Workspace{Float64}, ::Workspace{Float64}) at /Users/ronan.arraes/.julia/packages/DiffEqBase/8A5uo/src/data_array.jl:60
[...]

PR #508, opened 2020-05-16
Support for immutable fields derived from AbstractArray

ronisbr:
This PR closes the issue #507. Notice that, due to the discussion in #507, I had to move the test file `data_array_tests.jl` to the beginning of the set to test the new code.


ronisbr:
I have absolutely no idea what the cause of the error in Travis. Does it seem related to this modification?

ChrisRackauckas:
It's a stochastic thing that can happen due to lack of some optimizations causing a numerical error. I'll fix up that test: it shouldn't use numerical differentiation for this reason.

Issue #519, opened 2020-05-25
Collisionhandling of instantanious, depend events


freemin7:
To help out with some errors in occurring in #515 and to allow handling of proper collisions  [There is no way to handle this with idx only](https://youtu.be/7H4VUebyDa0). A change to VectorCallback or the creation of the new callback is required.

There are still open questions how to do this without allocations and what kind of array to pass.
@kanav99 offered to help me and to figure things out.

devmotion:
Just copied from my discussion on Slack with @freemin7:

> Due to https://github.com/SciML/OrdinaryDiffEq.jl/blob/a5fe013d4c377ef4a39fd7d12bdc54e4558e11a2/src/integrators/integrator_utils.jl#L251 and similar line,  you have to change integrator.vector_event_last_time as soon as you return an event_idx that is not an Int. Even if vector_event_last_time is moved to the callback cache (in which case it has to be an array as well since otherwise you can't have pooled and non-pooled VectorContinousCallbacks) you have to change all integrators. I mean, it will affect many packages but I think that's fine if it works and improves callback handling. BTW I always thought it's a bit weird to have this special field for VectorContinuousCallback (which is misused even if there are only regular callbacks), so maybe it would actually be a good idea to separate that from the integrator and remove it all together from the integrators.

So my suggestion would be to
- move `integrator.vector_event_last_time::Int` to `callback_cache.event_last_time::Vector{Int}`
  - in principle it could be of type `Int` if all `VectorContinuousCallback`s do not pool events, but I'm not sure if this optimization is worth it since it might make the implementation unnecessarily complex if it's not clear if we are dealing with integers or arrays of integers
- remove the handling of VectorContinuousCallback-specific implementation details from all integrators (if possible)
  - it seems apart from setting `integrator.vector_event_last_time` that's mainly the creation of the callback cache. I guess this logic is quite similar for different integrators, so maybe it could be moved to `DiffEqBase.callback_cache` (or something similar)

At a first glance that seems doable, but since I'm not familiar with all details of VectorContinuousCallback it would be good if some other people could comment on this proposal.

@kanav99 @ChrisRackauckas

kanav99:
Yes I think we should move it to callback_cache, after all it's supposed to store all that. I also think that we should set the value of `event_last_time` and `vector_event_last_time` inside the DiffEqBase code.

freemin7:
[...]

PR #523, opened 2020-05-31
[WIP] Adding pooled callbacks to VectorContinousCallback

freemin7:
I need help with this. I am currently running into:
`ERROR (update): MethodError: Cannot convert an object of type Array{Int64,1} to an object of type Int64`
at `/.julia/packages/OrdinaryDiffEq/JrtsK/src/integrators/integrator_utils.jl:251`
Looks like `integrator.vector_event_last_time`  expects to be an Integer.


freemin7:
I am not convinced that it handles non root-finding callbacks correctly. There are currently no test on callbacks at all.
This WIP PI is related #519 .

kanav99:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L251 this is where you are failing. You have to make it's type `Union{Int,Array{Int}}`

ChrisRackauckas:
How come this isn't caught by the tests?

kanav99:
Because they are behind the `pool_events` and we never reach them

ChrisRackauckas:
Ahh okay, so yeah it definitely needs a downstream test.

freemin7:
Well [test/callbacks.jl](https://github.com/SciML/DiffEqBase.jl/blob/master/test/callbacks.jl) is quiet sparse in general. I am not sure if normal callbacks get tested appropriately by it.

devmotion:
A lot should be covered by downstream tests such as https://github.com/SciML/DiffEqBase.jl/blob/master/test/downstream/event_detection_tests.jl.

src/callbacks.jl
```diff
@@ -209,14 +213,13 @@ function VectorContinuousCallback(condition,affect!,len;
                    affect_neg! = affect!,
                    interp_points=10,
                    dtrelax=1,
-                   abstol=10eps(),reltol=0)
+                   abstol=10eps(),reltol=0, pooltol=missing, pool_events=false)
```
devmotion, line 216: IMO usually in DiffEq default values are set to `nothing` (as indicated by the docstring) or a sensible default, maybe one could use the same as `abstol`?

src/callbacks.jl
```diff
@@ -754,6 +757,16 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
...
+      if callback.pool_events
+        tmp = get_condition(integrator, callback, integrator.dt + new_t)
+        if callback.pooltol isa Missing
+          # This is still dubious
+          pool_tol = eps(integrator.t + new_t) + eps(typeof(tmp[end]))
+        else
+          pool_tol = callback.pooltol
+        end
+        min_event_idx = findall(x-> abs(x) < pool_tol, tmp)
+      end
```
devmotion, line 769: I'm a bit surprised that one has to redo all calculations here. Is it not possible to change lines such as https://github.com/SciML/DiffEqBase.jl/blob/a13b48b67489750e28a042d151befe966f53f4a6/src/callbacks.jl#L735-L738 such that `min_t` and `min_event_idx` are not replaced but that one gets a list of `min_t` for every index, which then allows to select all indices within some tolerance later? And to change https://github.com/SciML/DiffEqBase.jl/blob/a13b48b67489750e28a042d151befe966f53f4a6/src/callbacks.jl#L751 and https://github.com/SciML/DiffEqBase.jl/blob/a13b48b67489750e28a042d151befe966f53f4a6/src/callbacks.jl#L755 to not just select the first index?

kanav99: Yes we discussed that, it is the initial PR, so lets get that perfect

src/callbacks.jl
freemin7: This is very prototypie but thanks for your input, the doc strings aren't really updated ...

src/callbacks.jl
freemin7: The code is not optimized since it does not work yet.
I have decided to keep the pooled callback logic out of branches for different types of time resolving, so there is just a single place to change it.

src/callbacks.jl
freemin7: The convention for the pooltol is also still in discussion, so i won't merge that proposal but keep in mind that doc strings would need to be added

src/callbacks.jl
ChrisRackauckas: I don't think this is necessarily the right direction. You're not looking for if the other values are near zero at the same time, you're looking for if there is another zero crossing nearby. One decent approximation would be to find the earliest event, and the move forward in time by some tolerance, and check the condition. You'd then take all of the events that were triggered in that time, and you'd know whether they are up or down crossings. Multiple crossings are extremely unlikely if this is on the size of a*eps(t). This would be insensitive to the relative size of the values.

PR #528, opened 2020-06-08
Use JuliaGPU's GitLab CI setup

christopher-dG:


Project.toml
```diff
@@ -62,13 +62,15 @@ ZygoteRules = "0.2"
 julia = "1.3"
 
 [extras]
+CuArrays = "3a865a2d-5b23-5a0f-bc46-62713ec82fae"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
```
devmotion, line 69: Currently downstream dependencies are part of a separate project: https://github.com/SciML/DiffEqBase.jl/blob/master/test/downstream/Project.toml Actually, I don't remember why that's the case.

ChrisRackauckas: it cut down on CI times a lot to only pull in the downstream dependencies for the tests that needed them

christopher-dG: Well currently the GPU group expects OrdinaryDiffEq to be installed. Should the GPU group also be activating the downstream env then?

christopher-dG: Since I'm seeing this "install test deps as required on a per-group basis" pattern a lot in these packages, I can eventually add something to the CI template to facilitate it, like:

```yml
variables:
  CI_JULIA_PACKAGES: CuArrays, OrdinaryDiffEq
```

But that's not great for local testing.

ChrisRackauckas: If you `]test DiffEqBase` then those tests aren't run by default anyways.

>Well currently the GPU group expects OrdinaryDiffEq and CuArrays to be installed. Should the GPU group also be activating the downstream env then?

Yes

christopher-dG:
Ok this one is good to go I think. The failed job is on the old mirrored repo with no runners properly set up for it.

ChrisRackauckas:
Will that failed job not exist in the future runs?

christopher-dG:
It shouldn't, I've deactivated the repo mirroring on the juliadiffeq repo.

PR #529, opened 2020-06-09
make dual norms include the dual terms

ChrisRackauckas:
This changes the norms on dual numbers to include the dual portions. This means that the solution to the ODE under automatic differentiation is not necessarily the same as without. However, https://github.com/SciML/DiffEqSensitivity.jl/issues/273 nicely highlights some edge cases that show that this is somewhat necessary. In that case, the integrator has zero error, and so time steps are unaffected by tolerances which causes gradients to explode since they are not tolerance controlled. This change adds them to the summation and the length, making the solution by dual numbers equivalent to the result of forward sensitivity analysis under the default norm.

@yingboma what do you think?

@pkofod you might want to test this with Pumas. It should give it some more robustness.

Fixes https://github.com/SciML/DiffEqSensitivity.jl/issues/273

### DIFF
--- a/src/init.jl
+++ b/src/init.jl
@@ -54,7 +54,7 @@ function __init__()
     totallength(x::AbstractArray) = sum(totallength,x)
 
     @inline ODE_DEFAULT_NORM(u::ForwardDiff.Dual,::Any) = sqrt(sse(u))
-    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual},t::Any) = sqrt(sum(x->sse(x),u) / totallength(u))
+    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual},t::Any) = sqrt(sum(sse,u) / totallength(u))
     @inline ODE_DEFAULT_NORM(u::ForwardDiff.Dual,::ForwardDiff.Dual) = sqrt(sse(u))
     @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual},::ForwardDiff.Dual) = sqrt(sum(x->sse(x),u) / totallength(u))
 
--- a/src/init.jl
+++ b/src/init.jl
@@ -48,7 +48,7 @@ function __init__()
     @inline fastpow(x::ForwardDiff.Dual, y::ForwardDiff.Dual) = x^y
 
     sse(x::Number) = x^2
-    sse(x::ForwardDiff.Dual) = sse(ForwardDiff.value(x)) + sum(sse.(ForwardDiff.partials(x)))
+    sse(x::ForwardDiff.Dual) = sse(ForwardDiff.value(x)) + sum(sse, ForwardDiff.partials(x))
     totallength(x::Number) = 1
     totallength(x::ForwardDiff.Dual) = totallength(ForwardDiff.value(x)) + sum(totallength.(ForwardDiff.partials(x)))
     totallength(x::AbstractArray) = sum(totallength,x)
--- a/src/init.jl
+++ b/src/init.jl
@@ -47,19 +47,16 @@ function __init__()
 
     @inline fastpow(x::ForwardDiff.Dual, y::ForwardDiff.Dual) = x^y
 
-    # Support adaptive with non-dual time
-    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual},::Any) = sqrt(sum(UNITLESS_ABS2∘value,u) / length(u))
-    @inline ODE_DEFAULT_NORM(u::ForwardDiff.Dual,::Any) = abs(value(u))
+    sse(x::Number) = x^2
+    sse(x::ForwardDiff.Dual) = sse(ForwardDiff.value(x)) + sum(sse.(ForwardDiff.partials(x)))
+    totallength(x::Number) = 1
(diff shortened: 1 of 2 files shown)


pkofod:
Is there a specific way in which this might impact Pumas or is it just a general thing?

ChrisRackauckas:
It'll change the time steps so that the partial terms are also controlled by the ODE solver tolerance. Effectively it's acknowledging that AD is generally unstable, so you do need to use a slightly different method when AD is involved, but a choice of norm is all that's necessary to make it equivalent to traditional forward sensitivity analysis.

src/init.jl
```diff
@@ -47,19 +47,16 @@ function __init__()
...
-    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual},::ForwardDiff.Dual) = sqrt(sum(UNITLESS_ABS2,u) / length(u))
-    @inline ODE_DEFAULT_NORM(u::ForwardDiff.Dual,::ForwardDiff.Dual) = abs(u)
-
-    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual{<:Any,<:ForwardDiff.Dual}},::ForwardDiff.Dual) = sqrt(sum(UNITLESS_ABS2∘value,u) / length(u))
-    @inline ODE_DEFAULT_NORM(u::ForwardDiff.Dual{<:Any,ForwardDiff.Dual},::ForwardDiff.Dual) = abs(value(u))
-
-    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual{<:Any,<:ForwardDiff.Dual}},::ForwardDiff.Dual{<:Any,ForwardDiff.Dual}) = sqrt(sum(UNITLESS_ABS2,u) / length(u))
-    @inline ODE_DEFAULT_NORM(u::ForwardDiff.Dual{<:Any,ForwardDiff.Dual},::ForwardDiff.Dual{<:Any,ForwardDiff.Dual}) = abs(u)
+    sse(x::Number) = x^2
+    sse(x::ForwardDiff.Dual) = sse(ForwardDiff.value(x)) + sum(sse.(ForwardDiff.partials(x)))
```
YingboMa, line 51:
```suggestion
    sse(x::ForwardDiff.Dual) = sse(ForwardDiff.value(x)) + sum(sse, ForwardDiff.partials(x))
```

src/init.jl
```diff
@@ -47,19 +47,16 @@ function __init__()
...
+    totallength(x::Number) = 1
+    totallength(x::ForwardDiff.Dual) = totallength(ForwardDiff.value(x)) + sum(totallength.(ForwardDiff.partials(x)))
```
YingboMa, line 53:
```suggestion
    totallength(x::ForwardDiff.Dual) = totallength(ForwardDiff.value(x)) + sum(totallength, ForwardDiff.partials(x))
```

src/init.jl
```diff
@@ -47,19 +47,16 @@ function __init__()
...
+    totallength(x::AbstractArray) = sum(totallength,x)
+
+    @inline ODE_DEFAULT_NORM(u::ForwardDiff.Dual,::Any) = sqrt(sse(u))
+    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual},t::Any) = sqrt(sum(x->sse(x),u) / totallength(u))
```
YingboMa, line 57:
```suggestion
    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual},::Any) = sqrt(sum(sse,u) / totallength(u))
```

ChrisRackauckas: No, we don't want to use the value part here: we do want to recurse.

src/init.jl
```diff
@@ -47,19 +47,16 @@ function __init__()
...
-    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual{<:Any,<:ForwardDiff.Dual}},::ForwardDiff.Dual{<:Any,ForwardDiff.Dual}) = sqrt(sum(UNITLESS_ABS2,u) / length(u))
-    @inline ODE_DEFAULT_NORM(u::ForwardDiff.Dual{<:Any,ForwardDiff.Dual},::ForwardDiff.Dual{<:Any,ForwardDiff.Dual}) = abs(u)
+    sse(x::Number) = x^2
+    sse(x::ForwardDiff.Dual) = sse(ForwardDiff.value(x)) + sum(sse, ForwardDiff.partials(x))
+    totallength(x::Number) = 1
+    totallength(x::ForwardDiff.Dual) = totallength(ForwardDiff.value(x)) + sum(totallength, ForwardDiff.partials(x))
+    totallength(x::AbstractArray) = sum(totallength,x)
+
+    @inline ODE_DEFAULT_NORM(u::ForwardDiff.Dual,::Any) = sqrt(sse(u))
+    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual},t::Any) = sqrt(sum(x->sse(x),u) / totallength(u))
```
YingboMa, line 57:
```suggestion
    @inline ODE_DEFAULT_NORM(u::AbstractArray{<:ForwardDiff.Dual},t::Any) = sqrt(sum(sse,u) / totallength(u))
```

YingboMa:
It looks good to me. I wanted to do this change before, but the concern is that, adaptive solvers don't land the same steps anymore. I am not sure if that's desired.

However, it does make sense to bound the error on derivatives to get a more accurate approximation to the derivative.

PR #531, opened 2020-06-11
Use fastmath on abs

ChrisRackauckas:
This results in dramatic speedups on complex numbers, and since this is only used in error norm heuristics, it doesn't need to be very accurate anyways.


ChrisRackauckas:
Good idea? Bad idea?

ChrisRackauckas:
Is there a way to make the adaptivity calculation just treat it as an array of two values? That would be sufficient.

YingboMa:
Now that I think about it. Fast complex abs is not anymore unsafe than what we have now. I don’t do overflow safe norm in the real case either.

PR #535, opened 2020-06-15
Check pointer not values when deciding whether to remake

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/620

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.38.2"
+version = "6.38.3"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
@christopher-dG would you happen to know why this doesn't seem to be triggering the GPU tests?

christopher-dG:
Sorry I wasn't in time for this. [It looks like they did run](https://gitlab.com/JuliaGPU/DiffEqBase.jl/-/jobs/594625216), so I'm guessing it wasn't showing up as a "status check" of this PR?

ChrisRackauckas:
Yeah, it didn't show up as a status check. That's curious.

christopher-dG:
My guess is that the token I used to set it up no longer has permissions to update commit statuses now that I don't have push access to these repos anymore.

ChrisRackauckas:
Oh... I'll just give you access to those repos then. Need to find that whole list again.

ChrisRackauckas:
You should have the right access through @SciML/gpu now

christopher-dG:
Cool, next time a PR pops up on one of those repos we can see if it's working.

PR #538, opened 2020-06-18
IncrementingODEProblem and IncrementingODEFunction

kanav99:
This is the bare minimum we need for incrementing forms, we can discuss what all things might make sense.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.jl.cov
 *.jl.*.cov
 *.jl.mem
+*.jl.*.mem
 Manifest.toml
 .DS_Store
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -590,6 +590,7 @@ export DAEProblem, DAESolution
 export DDEProblem
 export SDDEProblem
 export PDEProblem
+export IncrementingODEProblem
 
 export BVProblem, TwoPointBVProblem
 
@@ -598,7 +599,7 @@ export remake
 export DEDataArray, DEDataVector, DEDataMatrix
 
 export ODEFunction, DiscreteFunction, SplitFunction, DAEFunction, DDEFunction,
-       SDEFunction, SplitSDEFunction, RODEFunction, SDDEFunction
+       SDEFunction, SplitSDEFunction, RODEFunction, SDDEFunction, IncrementingODEFunction
 
 export ContinuousCallback, DiscreteCallback, CallbackSet, VectorContinuousCallback
 
--- a/src/diffeqfunction.jl
+++ b/src/diffeqfunction.jl
@@ -1641,3 +1641,24 @@ function Base.convert(::Type{SDDEFunction{iip}},f,g) where iip
               tgrad=tgrad,jac=jac,jvp=jvp,vjp=vjp,Wfact=Wfact,
               Wfact_t=Wfact_t,paramjac=paramjac,syms=syms,colorvec=colorvec)
 end
+
+struct IncrementingODEFunction{iip,F} <: AbstractODEFunction{iip}
+  f::F
+end
+
+function IncrementingODEFunction{iip}(f) where iip
+  IncrementingODEFunction{iip, typeof(f)}(f)
+end
+function IncrementingODEFunction(f)
+  IncrementingODEFunction{isinplace(f, 5), typeof(f)}(f)
+end
+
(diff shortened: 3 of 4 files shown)


kanav99:
Idea: We can add conversions to and from standard ODE problem using Williamson wrappers

kanav99:
Updated to support kwargs, now we can do scaling like suggested by Simon, I think this is good to merge, unless we need to convert them to args instead of kwargs

ChrisRackauckas:
Let's let @simonbyrne take a look. If it matches what they need, then we can do our merging brigade.

simonbyrne:
Can you expand a bit on how these would work? How would a user interact with them?

kanav99:
You can see this comment https://github.com/SciML/OrdinaryDiffEq.jl/pull/1185#issuecomment-650531272 it shows how a user can define a `f` for the problem

simonbyrne:
How are they related? i.e. when do I use IncrementingODEFunction vs IncrementingODEProblem?

ChrisRackauckas:
It might make sense to just have IncrementingODEFunction. IncrementingODEProblem is just a shorthand where you give it `f` and it builds an ODEProblem with an IncrementingODEFunction, but I'm not sure it's necessary. SecondOrderODEProblem and such are a bit different from this because they require different inputs and for example make an ArrayPartition to put together initial conditions, but IncrementingODEProblem just passes all of the arguments forward so it doesn't do all that much.

simonbyrne:
> It might make sense to just have IncrementingODEFunction.

That makes sense, I'll try that with the CliMA solvers.

simonbyrne:
Would be good to document the interface: see my comments here:  https://github.com/SciML/OrdinaryDiffEq.jl/pull/1185#issuecomment-650894826

kanav99:
Yes IncrementingODEProblem won't be needed, but it is a convenient constructor for getting directly f -> prob, so I would really like it to stay (lol, between my two comments, GitHub went down)

ChrisRackauckas:
It does match the others, so it is consistent. And we always document by problem. So I don't see any harm in it, though it's interesting that it's not entirely necessary.

devmotion:
Would it make sense to just define the convenient constructor but not make it a separate type?

ChrisRackauckas:
Yes, that's what we do with the others. There's also a trait that is set in the ODEProblem so you could know how it was constructed, though that has largely been supersceded now by using the DiffEqFunction types. So, no need for another problem type (though you build a fake one to use the "inner constructor")

simonbyrne:
I think it would make sense to define an additional abstract type in between, i.e.
```julia
abstract type AbstractIncrementingODEFunction <: AbstractODEFunction{true}
end
```
so that users could subtype that directly rather than use a wrapper, and then just define

```julia
const IncrementingODEProblem{uType,tType,P,F<: AbstractIncrementingODEFunction,K,PT} = ODEProblem{uType,tType,true,P,F,K,PT}
```

simonbyrne:
to be more specific, the following seems to work pretty well:
```julia
abstract type AbstractIncrementingODEFunction <: DiffEqBase.AbstractODEFunction{true}
end
struct IncrementingODEFunction{F} <: AbstractIncrementingODEFunction
    f::F
end

const IncrementingODEProblem{uType,tType,P,F<: AbstractIncrementingODEFunction,K,PT} =
    ODEProblem{uType,tType,true,P,F,K,PT}

IncrementingODEProblem(f::AbstractIncrementingODEFunction, args...; kwargs...) = 
    ODEProblem(f, args...; kwargs...)

IncrementingODEProblem(f, args...; kwargs...) =
    ODEProblem(IncrementingODEFunction(f), args...; kwargs...)
```

simonbyrne:
Or could we make it an alternative value for the `iip` parameter? Though I imagine there are a lot of places which assume it is a `Bool`.

ChrisRackauckas:
A lot of places assume it's a bool. I think it would be very breaking to change it. Though, IncrementingODEFunction would always be in-place, so I think it doesn't need it, so in theory we could equate `:incrementing` with `true` but... I can see this being really breaking in weird ways.

simonbyrne:
Perhaps just consider it a request for your next rewrite 😄

ChrisRackauckas:
So are we doing this?

kanav99:
Yes, it's ready to merge according to me!

PR #540, opened 2020-06-25
Make DiffEqArrayOperator ExponentialUtilities.jl compatible

YingboMa:
This is also related to https://github.com/SciML/OrdinaryDiffEq.jl/blob/91175ba20bf3e604f1b3150ec7a1ac49083f6487/src/derivative_utils.jl#L169-L170

It seems that we don't have the ability to express when we can reasonably "concretize" the operator and ask for the type of the concretized operator.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.39.0"
+version = "6.39.1"
 
 [deps]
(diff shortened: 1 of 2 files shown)


YingboMa:
This PR fixes the test failure seen in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1189

ChrisRackauckas:
We should probably formalize the concretization interface, which already exists in the DiffEqOperators derivative operators.

ChrisRackauckas:
Why would it need to similar on the operator though? It should be matrix-free?

YingboMa:
We need a concrete matrix for the subspace.

PR #541, opened 2020-06-26
Add `apply_only_on_accept`

YingboMa:
This PR adds `apply_only_on_accept` that defaults to true, when false, the
callback is applied even if the step is rejected due to various reasons.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -346,11 +346,11 @@ has_continuous_callback(cb::Nothing) = false
 # Callback handling
 #======================================================#
 function callbackcondition(callback, args...)
+  integrator = args[end]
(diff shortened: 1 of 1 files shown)


src/callbacks.jl
```diff
@@ -810,7 +823,7 @@ end
 #Base Case: Just one
 @inline function apply_discrete_callback!(integrator,callback::DiscreteCallback)
   saved_in_cb = false
-  if callback.condition(integrator.u,integrator.t,integrator)
+  if callbackcondition(callback,integrator.u,integrator.t,integrator)
```
ChrisRackauckas, line 826: this doesn't get called until an accept though?

YingboMa:
Yeah, I have a local change that moves `handle_callbacks!`.

PR #545, opened 2020-07-05
Remake for DEAlgorithm

kanav99:
Probably enough 🤣 @ChrisRackauckas


kanav99:
Not enough sorry, need to do something about chunksize and AD

PR #566, opened 2020-08-16
add literal u adjoint

ChrisRackauckas:


src/zygote.jl
```diff
@@ -15,25 +15,16 @@ ZygoteRules.@adjoint function ODESolution{T1,T2,T3,T4,T5,T6,T7,T8,T9,T10,T11
...
-
-ZygoteRules.@adjoint function getindex(sol::DESolution, i, j...)
-  function DESolution_getindex_adjoint(Δ)
-    Δ′ = zero(sol)
-    Δ′[i,j...] = Δ
-    (Δ′, map(_ -> nothing, i)...)
+ZygoteRules.@adjoint function ZygoteRules.literal_getproperty(sol::DESolution, ::Val{:u})
+  function solu_adjoint(Δ)
+        zerou = zero(sol.prob.u0)
+        _du = @. ifelse(du == nothing,(zerou,),du)
```
devmotion, line 23: This seems wrong, `du` is not defined. I guess it should be
```suggestion
        _du = @. ifelse(Δ === nothing, (zerou,), Δ)
```
?

ChrisRackauckas: Yeah... https://github.com/SciML/DiffEqBase.jl/commit/8e1a627d10ec40a112460860cd13948cc0532c63 patched 😆 . Sorry about that. This just needs a lot more machinery for a nice test, so the real test of it is going to be in https://github.com/SciML/DiffEqSensitivity.jl/pull/325

Issue #564, opened 2020-08-07
"Non bracketing interval passed in bisection method. Please report the error in DiffEqBase."


samuela:
I just got an error with the following code:
```julia
using DifferentialEquations
using Plots

gravity = 9.8
stiffness = 500
equilibrium_length = 1
T = 5.0

f(u, p, t) = begin
    x1, x2, dx1, dx2 = u
    length = abs(x2 - x1)
    spring_force = stiffness * (equilibrium_length - length)
    ddx1 = -gravity - spring_force
    ddx2 = -gravity + spring_force
    if x1 <= 0
        ddx1 = max(0, ddx1)
    end
    if x2 <= 0
        ddx2 = max(0, ddx2)
    end
    [dx1, dx2, ddx1, ddx2]
end

sol = solve(
    ODEProblem(f, [5.0, 6.0, 0.0, 0.0], (0.0, T)),
    # Euler(),
    # dt=0.005,
    Rosenbrock23(),
    callback = ContinuousCallback((u, _, _) -> u[1], (integrator) -> (integrator.u[1] = 0; integrator.u[3] = 0)),
    # callback = ContinuousCallback((u, _, _) -> u[1], (integrator) -> (integrator.u[3] = 0)),
    reltol = 1e-3,
    abstol = 1e-3,
)

@show sol.destats

# plot(sol, label = ["x1" "x2" "dx1" "dx2"], layout = (4, 1))
plot(
    [t -> sol(t)[1], t -> sol(t)[2] - sol(t)[1]],
    0.0,
    T,
    label = ["height" "spring length"],
    layout = (2, 1),
)
```
Here's the complete error:
```
ERROR: LoadError: Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.
Stacktrace:
 [1] error(::String) at ./error.jl:33
[...]

PR #574, opened 2020-08-27
Callback bug fixes

pengwyn:
In using VectorContinuousCallback in my code I came across several bugs, leading to exceptions, which I think I've fixed here.

1. The check `integrator.vector_event_last_time == event_idx` compared an Int to a Vector{Int} so it would always be false. I think it should be `idx` instead of `event_idx`.
2. The `diff_t` used in the nudging of the bottom t to find a different sign for bottom and top was using machine epsilon of the type, not the actual time. In my particular case, t was very large causing this to fail.
3. The check between `find_callback_time` and `bisection` was inconsistent. The nudger in `find_callback_time` was happy with a zero value for bottom_t, but `bisection` wanted a non-zero value.

I wanted to write some tests for this but ran out of time and figured that these are simple enough fixes. All the existing tests of `DiffEqBase` and `OrdinaryDiffEq` pass on my machine.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -661,7 +661,8 @@ function find_callback_time(integrator,callback::ContinuousCallback,counter)
             diff_t = integrator.tdir*2eps(bottom_t)
             bottom_t += diff_t
             iter = 1
-            while sign(zero_func(bottom_t)) == sign_top && iter < 12
+            # This check should match the same check in bisection
+            while sign(zero_func(bottom_t)) * sign_top >= zero(sign_top) && iter < 12
               diff_t *= 5
               bottom_t = integrator.tprev + diff_t
               iter += 1
@@ -728,7 +729,8 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
               diff_t = integrator.tdir * 2eps(bottom_t)
               bottom_t += diff_t
               iter = 1
-              while sign(zero_func(bottom_t)) == sign_top && iter < 12
+              # This check should match the same check in bisection
+              while sign(zero_func(bottom_t)) * sign_top >= zero(sign_top) && iter < 12
                 diff_t *= 5
                 bottom_t = integrator.tprev + diff_t
                 iter += 1
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -717,7 +717,7 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
             Θ = top_t
           else
             if integrator.event_last_time == counter &&
-              integrator.vector_event_last_time == event_idx &&
+              integrator.vector_event_last_time == idx &&
               abs(zero_func(bottom_t)) <= 100abs(integrator.last_event_error) &&
               prev_sign_index == 1
 
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -658,7 +658,7 @@ function find_callback_time(integrator,callback::ContinuousCallback,counter)
             # But floating point error may make the end point negative
 
             sign_top = sign(zero_func(top_t))
-            diff_t = integrator.tdir*2eps(typeof(bottom_t))
+            diff_t = integrator.tdir*2eps(bottom_t)
             bottom_t += diff_t
             iter = 1
             while sign(zero_func(bottom_t)) == sign_top && iter < 12
@@ -725,7 +725,7 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
               # But floating point error may make the end point negative
 
               sign_top = sign(zero_func(top_t))
-              diff_t = integrator.tdir * 2eps(typeof(bottom_t))
+              diff_t = integrator.tdir * 2eps(bottom_t)
               bottom_t += diff_t
               iter = 1
               while sign(zero_func(bottom_t)) == sign_top && iter < 12


src/callbacks.jl
```diff
@@ -717,18 +718,19 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
...
               # But floating point error may make the end point negative
 
               sign_top = sign(zero_func(top_t))
-              diff_t = integrator.tdir * 2eps(typeof(bottom_t))
+              diff_t = integrator.tdir * 2eps(bottom_t)
               bottom_t += diff_t
               iter = 1
-              while sign(zero_func(bottom_t)) == sign_top && iter < 12
+              # This check should match the same check in bisection
+              while sign(zero_func(bottom_t)) * sign_top >= zero(sign_top) && iter < 12
```
kanav99, line 733: This is probably one of the issues I was looking for..

src/callbacks.jl
```diff
@@ -717,18 +718,19 @@ function find_callback_time(integrator,callback::VectorContinuousCallback,counte
...
+              integrator.vector_event_last_time == idx &&
               abs(zero_func(bottom_t)) <= 100abs(integrator.last_event_error) &&
               prev_sign_index == 1
 
               # Determined that there is an event by derivative
               # But floating point error may make the end point negative
 
               sign_top = sign(zero_func(top_t))
-              diff_t = integrator.tdir * 2eps(typeof(bottom_t))
+              diff_t = integrator.tdir * 2eps(bottom_t)
```
kanav99, line 729: @ChrisRackauckas I am not good with eps, but should this be okay? I think this would be okay

kanav99:
@pengwyn thank you so much for your contribution! I agree with all 3 of the points you made.

> All the existing tests of DiffEqBase and OrdinaryDiffEq pass on my machine. 

Awesome! I assume your code works well with these changes too.

kanav99 (approved):

src/callbacks.jl
```diff
@@ -658,10 +658,11 @@ function find_callback_time(integrator,callback::ContinuousCallback,counter)
             # But floating point error may make the end point negative
 
             sign_top = sign(zero_func(top_t))
-            diff_t = integrator.tdir*2eps(typeof(bottom_t))
+            diff_t = integrator.tdir*2eps(bottom_t)
             bottom_t += diff_t
             iter = 1
-            while sign(zero_func(bottom_t)) == sign_top && iter < 12
+            # This check should match the same check in bisection
+            while sign(zero_func(bottom_t)) * sign_top >= zero(sign_top) && iter < 12
```
devmotion, line 665: Maybe computing `zero(sign_top)` before the loop would be slightly more efficient.

src/callbacks.jl
devmotion, line 733: Same here, maybe one could move `zero(sign_top)` out of the loop.

src/callbacks.jl
pengwyn: Shouldn't this be inlined by the compiler? Not that it is hard to move it outside the loop, but it is slightly less readable. Actually to contradict myself, I ran `@code_warntype` on it and it couldn't infer the result of `zero_func`.

Originally I thought that this duplicated check (between `bisection` and `find_callback_time`) would be best represented once as a function like:
```julia
different_sign(x::T,y::T) = sign(x) * sign(y) >= zero(T)
```
However, this wouldn't work in the scenario of Unitful quantities (there's a bit of asymmetry there between `zero(x)`, `one(x)` and `oneunit(x)`). But this would work:
```julia
different_sign(x::T,y::T) = sign(x) * sign(y) >= zero(sign(x))
```
as long as `zero(sign(x))` is inlined and the sign(x) calculation elided.

Alternatively, it could be argued that sign(x) is always one of [0, +1, -1] and so we could put directly:
```julia
different_sign(x::T,y::T) = sign(x) * sign(y) >= 0
```

pengwyn:
> @pengwyn thank you so much for your contribution! I agree with all 3 of the points you made.

@kanav99 I'm glad you find it useful! I hope it helps with #564 too.

> Awesome! I assume your code works well with these changes too.

Yeap, it's hitting the right benchmarks in my physics problem on everything I've tried so far!

src/callbacks.jl
pengwyn: I just realised that was a brain dump and not a proper reply. My question is should I a) keep the same style (and move the zero(..) out of the for loop)? or b) turn it into a separate function and put that into both `find_callback_time` and `bisection`?

ChrisRackauckas:
We should check how many of https://github.com/SciML/DifferentialEquations.jl/labels/callbacks are fixed by this, and get tests on them. I assume that the fix of making the `eps(t)` instead of `eps(typeof(t))` would fix more than a couple of them.

src/callbacks.jl
ChrisRackauckas: IIRC LLVM usually will handle this case if it's type-stable.

src/callbacks.jl
ChrisRackauckas: >Actually to contradict myself, I ran @code_warntype on it and it couldn't infer the result of zero_func.

Oh no, how come?

src/callbacks.jl
ChrisRackauckas: yes, this is much better. I didn't catch this before, but you almost never want to do `eps(typeof(t))` and always `eps(t)` since floating point is relative.

src/callbacks.jl
ChrisRackauckas: yeah, this is one of those things where it's like... huh, yeah... that's umm... wow how did anything ever work before? 😆

src/callbacks.jl
pengwyn: It turned out to be 3 bugs, one in my external code and two more in the callbacks.jl file. I'll open another pull request about them.

PR #576, opened 2020-08-27
Reintroduce DataStructures bound

devmotion:
DataStructures is not imported anymore, but the package is definitely not compatible with DataStructures 0.17 anymore. Have to fix the registry as well.


ChrisRackauckas:
We need to make sure v6.46 also has the bound though in General because otherwise people will keep getting served that version and have the issue.

devmotion:
Ah GPUCompiler is not compatible with DataStructures 0.18, hence tests are failing.

PR #577, opened 2020-08-28
Fixed inference issues in bisection and CallbackCache

pengwyn:
In following up my inference issues identified in #574, I think I found another couple of bugs.

1. The `prevfloat_tdir` variable in `bisection(...)` doesn't seem to be expanded into its two possibilities by the compiler, so the return value of the function was an `Any`. I "fixed" this by putting it as an argument to the function with a default. This feels like a dirty hack to me... is there a better way to get the same result?

2. One of the `CallbackCache` constructors (which is used only for `VectorContinuousCallback`) was setting the parameters to be `Array{Float64}` without a dimension, so it would appear as `Array{Float64,N}` in my integrator object. I don't think Array was the intent anyway, so I changed it to be the same as the other constructor, which is `typeof(tmp_condition)`.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -882,7 +882,7 @@ function CallbackCache(u,max_len,::Type{conditionType},::Type{signType}) where {
     previous_condition = similar(u, conditionType, max_len)
     next_sign = similar(u, signType, max_len)
     prev_sign = similar(u, signType, max_len)
-    CallbackCache{typeof(tmp_condition),typeof(next_sign)}(tmp_condition,previous_condition,next_sign,prev_sign)
+    CallbackCache(tmp_condition,previous_condition,next_sign,prev_sign)
 end
 
 function CallbackCache(max_len,::Type{conditionType},::Type{signType}) where {conditionType,signType}
@@ -890,5 +890,5 @@ function CallbackCache(max_len,::Type{conditionType},::Type{signType}) where {co
     previous_condition = zeros(conditionType, max_len)
     next_sign = zeros(signType, max_len)
     prev_sign = zeros(signType, max_len)
-    CallbackCache{typeof(tmp_condition),typeof(next_sign)}(tmp_condition,previous_condition,next_sign,prev_sign)
+    CallbackCache(tmp_condition,previous_condition,next_sign,prev_sign)
 end
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -590,7 +590,10 @@ function bisection(f, tup, tdir ; maxiters=1000)
   fx0x1 = f(x0) * f(x1)
   fzero = zero(fx0x1)
   (fx0x1 >= fzero) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
-  prevfloat_tdir(t,forward=isone(tdir)) = forward ? prevfloat(t) : nextfloat(t)
+  # The comment line below does not allow for inference, yet the second version does.
+  forward=isone(tdir)
+  #prevfloat_tdir = forward ? prevfloat : nextfloat
+  prevfloat_tdir(t) = forward ? prevfloat(t) : nextfloat(t)
   left = x0
   right = x1
   iter = 0
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -585,12 +585,12 @@ end
 # rough implementation, needs multiple type handling
 # always ensures that if r = bisection(f, (x0, x1))
 # then either f(nextfloat(r)) == 0 or f(nextfloat(r)) * f(r) < 0
-function bisection(f, tup, tdir, prevfloat_tdir=isone(tdir) ? prevfloat : nextfloat
-                   ; maxiters=1000)
+function bisection(f, tup, tdir ; maxiters=1000)
   x0, x1 = tup
   fx0x1 = f(x0) * f(x1)
   fzero = zero(fx0x1)
   (fx0x1 >= fzero) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
+  prevfloat_tdir(t,forward=isone(tdir)) = forward ? prevfloat(t) : nextfloat(t)
   left = x0
   right = x1
   iter = 0
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -590,9 +590,7 @@ function bisection(f, tup, tdir ; maxiters=1000)
   fx0x1 = f(x0) * f(x1)
   fzero = zero(fx0x1)
   (fx0x1 >= fzero) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
-  # The comment line below does not allow for inference, yet the second version does.
   forward=isone(tdir)
-  #prevfloat_tdir = forward ? prevfloat : nextfloat
   prevfloat_tdir(t) = forward ? prevfloat(t) : nextfloat(t)
   left = x0
   right = x1
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -585,14 +585,14 @@ end
 # rough implementation, needs multiple type handling
 # always ensures that if r = bisection(f, (x0, x1))
 # then either f(nextfloat(r)) == 0 or f(nextfloat(r)) * f(r) < 0
-function bisection(f, tup, tdir; maxiters=1000)
+function bisection(f, tup, tdir, prevfloat_tdir=isone(tdir) ? prevfloat : nextfloat
+                   ; maxiters=1000)
   x0, x1 = tup
   fx0x1 = f(x0) * f(x1)
   fzero = zero(fx0x1)
   (fx0x1 >= fzero) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
   left = x0
   right = x1
-  prevfloat_tdir = isone(tdir) ? prevfloat : nextfloat
   iter = 0
   while true
     iter += 1
@@ -890,5 +890,5 @@ function CallbackCache(max_len,::Type{conditionType},::Type{signType}) where {co
     previous_condition = zeros(conditionType, max_len)
     next_sign = zeros(signType, max_len)
     prev_sign = zeros(signType, max_len)
-    CallbackCache{Array{conditionType},Array{signType}}(tmp_condition,previous_condition,next_sign,prev_sign)
+    CallbackCache{typeof(tmp_condition),typeof(next_sign)}(tmp_condition,previous_condition,next_sign,prev_sign)
 end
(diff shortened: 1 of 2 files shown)


src/callbacks.jl
```diff
@@ -890,5 +890,5 @@ function CallbackCache(max_len,::Type{conditionType},::Type{signType}) where {co
     previous_condition = zeros(conditionType, max_len)
     next_sign = zeros(signType, max_len)
     prev_sign = zeros(signType, max_len)
-    CallbackCache{Array{conditionType},Array{signType}}(tmp_condition,previous_condition,next_sign,prev_sign)
+    CallbackCache{typeof(tmp_condition),typeof(next_sign)}(tmp_condition,previous_condition,next_sign,prev_sign)
```
devmotion, line 893: Are the type parameters not inferred automatically by the constructor?

src/callbacks.jl
```diff
@@ -585,14 +585,14 @@ end
...
-function bisection(f, tup, tdir; maxiters=1000)
+function bisection(f, tup, tdir, prevfloat_tdir=isone(tdir) ? prevfloat : nextfloat
+                   ; maxiters=1000)
   x0, x1 = tup
   fx0x1 = f(x0) * f(x1)
   fzero = zero(fx0x1)
   (fx0x1 >= fzero) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
   left = x0
   right = x1
-  prevfloat_tdir = isone(tdir) ? prevfloat : nextfloat
```
devmotion, line 595: Alternatively, it might be possible to just define a function `prevfloat_tdir(t, tdir) = isone(tdir) ? prevfloat(t) : nextfloat(t)` (maybe make `tdir` a boolean `forward` here) and call it in the function. I'd assume it's inferrable.

src/callbacks.jl
pengwyn: They would normally be inferred - except the call there forces the parameters to be non-concrete.

Unless you mean, why put explicit types in there at all? I think it's needed to disambiguate from the hand-written CallbackCache(...) constructor.

ChrisRackauckas: Yeah, the types can probably be completely removed from this. But the types should've been `Array{conditionType,1}` etc., and that might've been the original issue.

ChrisRackauckas, line 893:
```suggestion
    CallbackCache(tmp_condition,previous_condition,next_sign,prev_sign)
```

ChrisRackauckas: No, that type designation is necessary to not collide with the other method.

devmotion: Yeah that's what I meant - I thought that probably the types can be removed completely.

devmotion: > No, that type designation is necessary to not collide with the other method.

I see :+1:

src/callbacks.jl
pengwyn: Yeap that works! I just pushed a new commit.

src/callbacks.jl
pengwyn: ... or not. I guess it's because the hand-written constructors are only triggered in the special case of Type{}s and they always recall the constructor with instantiated values.

I changed over both versions of the constructor to do this now (also matching @ChrisRackauckas's suggestion too)

src/callbacks.jl
```diff
@@ -585,14 +585,14 @@ end
 # rough implementation, needs multiple type handling
 # always ensures that if r = bisection(f, (x0, x1))
 # then either f(nextfloat(r)) == 0 or f(nextfloat(r)) * f(r) < 0
-function bisection(f, tup, tdir; maxiters=1000)
+function bisection(f, tup, tdir ; maxiters=1000)
   x0, x1 = tup
   fx0x1 = f(x0) * f(x1)
   fzero = zero(fx0x1)
   (fx0x1 >= fzero) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
+  prevfloat_tdir(t,forward=isone(tdir)) = forward ? prevfloat(t) : nextfloat(t)
```
devmotion, line 593: I thought about moving it outside of the function actually since it is defined in the same way always. You would then just evaluate `isone(tdir)` once and call a two argument function.

pengwyn: I've changed it to calculated `isone(tdir)` only once now. Weirdly, the difference between a definition of `prevfloat_tdir = ...` or `prevfloat_tdir(t) = ...(t)` is enough to allow for inference. I can vaguely understand this as the difference between a variable and a hidden-scope function definition, but it seems quite surprising to me.

I've kept it as a nested function with the idea that it's only used once inside of `bisection`.

src/callbacks.jl
```diff
@@ -590,7 +590,10 @@ function bisection(f, tup, tdir ; maxiters=1000)
   fx0x1 = f(x0) * f(x1)
   fzero = zero(fx0x1)
   (fx0x1 >= fzero) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
-  prevfloat_tdir(t,forward=isone(tdir)) = forward ? prevfloat(t) : nextfloat(t)
+  # The comment line below does not allow for inference, yet the second version does.
+  forward=isone(tdir)
+  #prevfloat_tdir = forward ? prevfloat : nextfloat
+  prevfloat_tdir(t) = forward ? prevfloat(t) : nextfloat(t)
```
devmotion, line 596: Haha I guess I'm not explaining well enough what I mean. I thought one doesn't have to use nested functions at all. Nevertheless, in this form maybe it would be inferrable even without defining `forward` at all?
I.e.
```suggestion
  prevfloat_tdir(t) = isone(tdir) ? prevfloat(t) : nextfloat(t)
```

pengwyn: Oh yes it definitely infers in the suggestion you've made. I thought the concern was on the duplicated checks of isone(tdir) everytime prevfloat_tdir is called.

Are nested functions considered bad practice? I like the idea here as it is not something that another function should casually be using.

devmotion: No it's fine. And written in this way the check should only be performed once. It is just that if you would define a function `prevfloat_tdir(t, forward)` then one should precompute `isone(tdir)` instead of doing it over and over again in the loop.

pengwyn: Okay, good stuff. I cleaned up the comments and left it inside and with the precomputed forward.

src/callbacks.jl
```diff
@@ -585,14 +585,15 @@ end
 # rough implementation, needs multiple type handling
 # always ensures that if r = bisection(f, (x0, x1))
 # then either f(nextfloat(r)) == 0 or f(nextfloat(r)) * f(r) < 0
-function bisection(f, tup, tdir; maxiters=1000)
+function bisection(f, tup, tdir ; maxiters=1000)
```
devmotion, line 588: Just had one additional idea: Could we just make this
```suggestion
function bisection(f, tup, forward; maxiters=1000)
```
instead? It seems `tdir` is never actually used apart from the evaluation `isone(tdir)`.

src/callbacks.jl
```diff
@@ -585,14 +585,15 @@ end
...
   x0, x1 = tup
   fx0x1 = f(x0) * f(x1)
   fzero = zero(fx0x1)
   (fx0x1 >= fzero) && error("Non bracketing interval passed in bisection method. Please report the error in DiffEqBase.")
+  forward=isone(tdir)
```
devmotion, line 593: Then this line could be removed:
```suggestion
```

pengwyn: Yeap that makes sense. In fact, it begs the question if the variable `tdir` in the integrator should be a singleton type, analogous to `Forward` and `Backward` for `sort`. This would provide a tiny optimisation whenever `tdir` is used.

I'll change the current pull request to make the bisection function arg forward instead of tdir, but what about another change in a different pull request to modify how tdir is stored?

Speaking of which, the number of parameters in the definition of the structs (e.g. ODEIntegrator) is massive. I made a [package](https://github.com/pengwyn/AutoParameters.jl) to deal with exactly this problem. I'll open an issue to mention it, see if it tickles anyone's fancy.

pengwyn: I included a Bool check in `forward::Bool` as well just in case there is some other function I'm not aware of using `bisection`.

ChrisRackauckas: >Yeap that makes sense. In fact, it begs the question if the variable tdir in the integrator should be a singleton type, analogous to Forward and Backward for sort. This would provide a tiny optimisation whenever tdir is used.

I'm not sure that's a real optimization, and it would make for more compilation that is required. Probably not the best idea in here.

>I included a Bool check in forward::Bool as well just in case there is some other function I'm not aware of using bisection.

There shouldn't be, but safety first!

>Speaking of which, the number of parameters in the definition of the structs (e.g. ODEIntegrator) is massive. I made a package to deal with exactly this problem. I'll open an issue to mention it, see if it tickles anyone's fancy.

That seems like a good solution.

pengwyn: > I'm not sure that's a real optimization, and it would make for more compilation that is required. Probably not the best idea in here.

Ah I didn't think of that. I'll leave it alone then.

src/callbacks.jl
```diff
@@ -882,13 +882,13 @@ function CallbackCache(u,max_len,::Type{conditionType},::Type{signType}) where {
...
-    CallbackCache{typeof(tmp_condition),typeof(next_sign)}(tmp_condition,previous_condition,next_sign,prev_sign)
+    CallbackCache(tmp_condition,previous_condition,next_sign,prev_sign)
 end
 
 function CallbackCache(max_len,::Type{conditionType},::Type{signType}) where {conditionType,signType}
     tmp_condition = zeros(conditionType, max_len)
     previous_condition = zeros(conditionType, max_len)
     next_sign = zeros(signType, max_len)
     prev_sign = zeros(signType, max_len)
-    CallbackCache{Array{conditionType},Array{signType}}(tmp_condition,previous_condition,next_sign,prev_sign)
```
ChrisRackauckas, line 893: won't this hit the wrong method?

pengwyn: I originally thought so, but actually it's not a problem because the explicit constructors require a Type of some kind, whereas this is calling the default constructor method with instances.

ChrisRackauckas:
Looks like this hit a test error https://travis-ci.com/github/SciML/DiffEqBase.jl/jobs/379012045#L727-L731

devmotion:
The type annotation `forward::Bool` leads to a method ambiguity error when calling https://github.com/pengwyn/DiffEqBase.jl/blob/5844608b2817826bed4f59618428a1e301874a43/src/init.jl#L97.

ChrisRackauckas:
@kanav99 be aware of this.

PR #578, opened 2020-08-30
Add missing has_continuous_callback(::VectorContinuousCallback) method

mchitre:
Fix for https://github.com/SciML/DifferentialEquations.jl/issues/662

Reproduced here for quick reference:

---

When trying to forward diff through an ODE with a `VectorContinuousCallback`, I’m getting:
```
ERROR: MethodError: no method matching has_continuous_callback(::DiffEqBase.VectorContinuousCallback{...})
```

Tracing through a little more, I find that `has_continuous_callback()` is called if `ForwardDiff` is loaded, but they are only defined for:

```
has_continuous_callback(cb::DiscreteCallback) = false
has_continuous_callback(cb::ContinuousCallback) = true
has_continuous_callback(cb::CallbackSet) = !isempty(cb.continuous_callbacks)
has_continuous_callback(cb::Nothing) = false
```
and not for `VectorContinuousCallback`.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -333,6 +333,7 @@ Base.isempty(cb::AbstractDiscreteCallback) = false
 
 has_continuous_callback(cb::DiscreteCallback) = false
 has_continuous_callback(cb::ContinuousCallback) = true
+has_continuous_callback(cb::VectorContinuousCallback) = true
 has_continuous_callback(cb::CallbackSet) = !isempty(cb.continuous_callbacks)
 has_continuous_callback(cb::Nothing) = false
 
--- a/test/downstream/ode_event_tests.jl
+++ b/test/downstream/ode_event_tests.jl
@@ -1,4 +1,4 @@
-using OrdinaryDiffEq, RecursiveArrayTools, Test, StaticArrays, DiffEqCallbacks
+using OrdinaryDiffEq, RecursiveArrayTools, Test, StaticArrays, DiffEqCallbacks, ForwardDiff
 
 
 f = function (u,p,t)
@@ -287,3 +287,31 @@ integrator = init(
 )
 
 sol = solve!(integrator)
+
+
+### https://github.com/SciML/DifferentialEquations.jl/issues/662
+
+function prob1!(du, u, p, s)
+  du[1] = u[2]
+  du[2] = zero(eltype(u))
+end
+
+function cb1test(out, u, x)
+  out[1] = u[1] - x
+end
+
+function cb1action!(i)
+  i.u[2] = -i.u[2]
+end
+
+function ode1(x)
+  prob = ODEProblem{true}(prob1!, [zero(typeof(x)), one(typeof(x))], (0.0, 1.0))
+  cb = VectorContinuousCallback(
+    (out, u, t, i) -> cb1test(out, u, x),
+    (i, ndx) -> cb1action!(i),
+    1; rootfind=true)
+  soln = solve(prob, Tsit5(); callback=cb)
+  soln[end][1]
+end
+
+@test ForwardDiff.derivative(ode1, 0.42) ≈ 2


ChrisRackauckas:
When does this show up? It would be good to get a test since it seems to be a blind spot.

mchitre:
My original code is too complicated to post, but I'll make up a MWE shortly.

ChrisRackauckas:
thanks that would be fantastic.

mchitre:
Here's a MWE:
```Julia
function prob1!(du, u, p, s)
  du[1] = u[2]
  du[2] = zero(eltype(u))
end

function cb1test(out, u, x)
  out[1] = u[1] - x
end

function cb1action!(i)
  i.u[2] = -i.u[2]
end

function ode1(x)
  prob = ODEProblem{true}(prob1!, [zero(typeof(x)), one(typeof(x))], (0.0, 1.0))
  cb = VectorContinuousCallback(
    (out, u, t, i) -> cb1test(out, u, x),
    (i, ndx) -> cb1action!(i),
    1; rootfind=true)
  soln = solve(prob, Tsit5(); callback=cb)
  soln[end][1]
end
```

Before fix:
```julia
julia> ode1(0.42)
-0.16000000000000056

julia> ForwardDiff.derivative(ode1, 0.42)
ERROR: MethodError: no method matching has_continuous_callback(::VectorContinuousCallback{var"#11#13"{ForwardDiff.Dual{ForwardDiff.Tag{typeof(ode1),Float64},Float64,1}},var"#12#14",var"#12#14",typeof(DiffEqBase.INITIALIZE_DEFAULT),Float64,Int64,Nothing,Int64})
Closest candidates are:
  has_continuous_callback(::Nothing) at /Users/mandar/.julia/packages/DiffEqBase/m5bAm/src/callbacks.jl:337
  has_continuous_callback(::DiscreteCallback) at /Users/mandar/.julia/packages/DiffEqBase/m5bAm/src/callbacks.jl:334
  has_continuous_callback(::ContinuousCallback) at /Users/mandar/.julia/packages/DiffEqBase/m5bAm/src/callbacks.jl:335
  ...
Stacktrace:
 [1] promote_tspan(::Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(ode1),Float64},Float64,1},1}, ::DiffEqBase.NullParameters, ::Tuple{Float64,Float64}, ::ODEProblem{Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(ode1),Float64},Float64,1},1},Tuple{Float64,Float64},true,DiffEqBase.NullParameters,ODEFunction{true,typeof(prob1!),LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}},DiffEqBase.StandardODEProblem}, ::Base.Iterators.Pairs{Symbol,Any,Tuple{Symbol,Symbol,Symbol},NamedTuple{(:u0, :p, :callback),Tuple{Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(ode1),Float64},Float64,1},1},DiffEqBase.NullParameters,VectorContinuousCallback{var"#11#13"{ForwardDiff.Dual{ForwardDiff.Tag{typeof(ode1),Float64},Float64,1}},var"#12#14",var"#12#14",typeof(DiffEqBase.INITIALIZE_DEFAULT),Float64,Int64,Nothing,Int64}}}}) at /Users/mandar/.julia/packages/DiffEqBase/m5bAm/src/init.jl:40
```

After fix:
```julia
julia> ode1(0.42)
-0.16000000000000056

julia> ForwardDiff.derivative(ode1, 0.42)
1.9999999999999962
```
which is the correct answer (exact answer for this problem is 2).

ChrisRackauckas:
Thanks, I added that to the downstream callback test cases and will merge when it passes.

mchitre:
Do note, however, that if I change the problem definition to:
```
prob = ODEProblem{true}(prob1!, [zero(typeof(x)), one(typeof(x))], (zero(typeof(x)), one(typeof(x))))
```
the code works correctly without the fix. And perhaps that is the right thing to do for this problem.

However, if I use a `ContinuousCallback` instead of a `VectorContinuousCallback`, the original problem definition also works correctly. Hence I feel that the `VectorContinuousCallback` should have the same behavior.

ChrisRackauckas:
Yes, this is a bug and should be fixed. A user shouldn't have to work around this.

PR #579, opened 2020-09-02
Reimplement TableTraits.jl by implementing Tables.jl

quinnj:
This opens up solutions as valid tables in many more operations
(Tables.jl currently has ~100 integrations) and simplifies the code by
relying on Tables.jl utilities to maintain TableTraits.jl compat.


src/tabletraits.jl
```diff
@@ -1,66 +1,48 @@
-struct DESolutionIterator{T,S}
-    sol::S
+Tables.isrowtable(::Type{T}) where {T} = true
```
ChrisRackauckas, line 1: Is this piracy?

quinnj: Oops! Yeah, this is supposed to be
```suggestion
Tables.isrowtable(::Type{<:DESolution}) = true
```

PR #585, opened 2020-09-18
Fix L(u,p,t) call for AffineDiffEqOperator

briochemc:
This PR fixes (I hope) calls of the type `L(u,p,t)` for `L::AffineDiffEqOperator`.

I think there was a number of mistakes in this function (e.g., the spurious double parentheses in `sum((...))`).

Also, I thought that once `tmp` is created with the `A*u` terms, we could add the `B`s to it inplace. (So I renamed `tmp` as `du` to clarify that, but maybe that's not good practice?)

### DIFF
--- a/src/operators/diffeq_operator.jl
+++ b/src/operators/diffeq_operator.jl
@@ -92,9 +92,16 @@ Base.size(L::AffineDiffEqOperator) = size(L.As[1])
 
 
 function (L::AffineDiffEqOperator)(u,p,t::Number)
-    tmp = sum((update_coefficients(A,u,p,t); A*u for A in L.As))
-    tmp2 = sum((typeof(B) <: Union{Number,AbstractArray} ? B : B(t) for B in L.Bs))
-    tmp + tmp2
+    update_coefficients!(L,u,p,t)
+    du = sum(A*u for A in L.As)
+    for B in L.Bs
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Can this get a test?

briochemc:
This adds a few tests but is definitely not exhaustive. I'm still not completely at home with these operators API, so sorry if I'm issing stuff. 

Speaking of which, as I mentioned on Slack, I was trying to add some `DiffEqScalar`s in this test, but I don't think they combine well with the linear operators and various operations. Should I start different issues for these things I (think I) see?

ChrisRackauckas:
>Speaking of which, as I mentioned on Slack, I was trying to add some DiffEqScalars in this test, but I don't think they combine well with the linear operators and various operations. Should I start different issues for these things I (think I) see?

Yes please!

>This adds a few tests but is definitely not exhaustive. I'm still not completely at home with these operators API, so sorry if I'm issing stuff.

It's not you, I think half of the interface is just in my head on "what it should be" and it's not all there. We should probably work together to make all of the operators conform to that and get the specification hammered down.

briochemc:
OK Great, I'll happily contribute! However just FYI, I'm going on a break so I won't do much until end of September :)

ChrisRackauckas:
no worries at all.

PR #587, opened 2020-09-25
Match type params with their corresponding fields

jonniedie:
For `SDEFunction` and `SDDEFunction`, the type parameters `S` and `GG` are out of order with their corresponding fields `ggprime` and `syms`.


ChrisRackauckas:
Thanks! We should probably use that new package to make this more sane!

PR #594, opened 2020-10-17
Remove num_cons field from OptimizationFunction

Vaibhavdixit02:
We don't need a separate field for it since it can just be obtained from the length of the constraint bounds


ChrisRackauckas:
This is a breaking change... but since it's only used by unreleased portions of the downstream libraries (and GalacticOptim in general is essentially unreleased) I'll allow it.

PR #597, opened 2020-10-21
INITIALIZE_DEFAULT / FINALIZE_DEFAULT => initialize! / finalize!

simonbyrne:
We want to be able to define callbacks that have predefined initializer/finalizers, e.g.
```
struct Diagnostics
  # some stuff
end
DiffEqBase.initialize!(d::Diagnostics, u, t, integrator) = # set up stuff
DiffEqBase.finalize!(d::Diagnostics, u, t, integrator) = # finalize stuff

DiscreteCallback(mycond, Diagnostics(details)) # don't need to pass explicit initializer / finalizer functions
```

We could do this now by defining additional methods for `INITIALIZE_DEFAULT` / `FINALIZE_DEFAULT`, but this seems a bit clearer.

If you're on board, I can add some docs.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -1,7 +1,10 @@
 # Necessary to have initialize set u_modified to false if all don't do anything
 # otherwise unnecessary save
-INITIALIZE_DEFAULT(cb,u,t,integrator) = u_modified!(integrator, false)
-FINALIZE_DEFAULT(cb,u,t,integrator) = nothing
+initialize!(cb,u,t,integrator) = u_modified!(integrator, false)
+finalize!(cb,u,t,integrator) = nothing
+
+const INITIALIZE_DEFAULT = initialize!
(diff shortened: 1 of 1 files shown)


simonbyrne:
Okay, I realise this won't quite work as it is passed the `DiscreteCallback`, not the `affect!` argument.

I'll need to think a bit more about this.

simonbyrne:
What I had in mind was something like this:
https://github.com/CliMA/TimeMachine.jl/pull/13/files#diff-f85085614ae2b81b449d2b6235530313731991388cf8d97b3614882bc8b036fd

ChrisRackauckas:
Yeah I see what you're getting at. Usually for this kind of thing I'd make a callable type will all of the function dispatches and pass that type in all spots, but the init and finalize functions would clash so that's a little tricky.

PR #602, opened 2020-11-23
Change to NonlinearSolve.jl

kanav99:
@YingboMa @ChrisRackauckas


ChrisRackauckas:
What makes this WIP?

kanav99:
Because it still doesn't pass `integrator.u` as a parameter to the `zero_func`. I passed this to test the `ad_tests.jl` file

ChrisRackauckas:
Can we just dispatch to Regula Falsi on duals for now, merge, and then handle the other problem?

kanav99:
The problem with Falsi is that - for linear condition functions, it finds the solution in one iteration. But it is a very high probability that the solution it finds is slightly greater and keeps on setting the `right` value, `left` goes on unchanged.

ChrisRackauckas:
But that's no worse than before, just no better in the AD case?

kanav99:
Let's wait for tests to pass, especially downstream

ChrisRackauckas:
We need to switch to github actions

kanav99:
I have rebased and all the tests pass. This can be merged while I bring Falsi in a separate PR.

PR #603, opened 2020-11-26
Revert "Non-allocating isinplace"

kanav99:
Reverts SciML/DiffEqBase.jl#601

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   build:
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -4,25 +4,32 @@ on:
   push:
     branches:
       - master
-    tags: '*'
   pull_request:
 
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        group:
+          - Core
+          - Downstream
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
         with:
-          version: '1.5'
-      - name: Install dependencies
-        run: julia --project=./ -e 'using Pkg; Pkg.instantiate()'
-      - name: Test
+          version: '1'
+      - uses: actions/cache@v1
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
-        run: julia --project=./ -e 'using Pkg; pkg"test";'
-      - name: Test
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.48.3"
+version = "6.48.2"
```
kanav99, line 4: We need to keep it 6.48.3 right?

ChrisRackauckas: yes, we also need to setup Github actions.

kanav99: yeah let me do that in this PR

.github/workflows/Test.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  push:
+    branches:
+      - master
+    tags: '*'
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
devmotion, line 12: Maybe better to run two tests in parallel?
```suggestion
    runs-on: ubuntu-latest
    strategy:
      matrix:
        group:
          - Core
          - Downstream
```

.github/workflows/Test.yml
```diff
@@ -0,0 +1,28 @@
...
+        run: julia --project=./ -e 'using Pkg; Pkg.instantiate()'
+      - name: Test
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
+        run: julia --project=./ -e 'using Pkg; pkg"test";'
+      - name: Test
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
+          GROUP: Downstream
+        run: julia --project=./ -e 'using Pkg; pkg"test";'
```
devmotion, line 28: Just use the default actions and cache artifacts?
```suggestion
      - uses: actions/cache@v1
        env:
          cache-name: cache-artifacts
        with:
          path: ~/.julia/artifacts
          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
          restore-keys: |
            ${{ runner.os }}-test-${{ env.cache-name }}-
            ${{ runner.os }}-test-
            ${{ runner.os }}-
      - uses: julia-actions/julia-buildpkg@latest
      - uses: julia-actions/julia-runtest@latest
        env:
          GROUP: ${{ matrix.group }}
```

devmotion (commented):
I added some suggestions. It would also be good to add some coverage analysis, such as e.g. in https://github.com/TuringLang/DynamicPPL.jl/blob/cf95183bb085eaacaea69d7c91dfc8c9c3b45f17/.github/workflows/DynamicPPL-CI.yml#L58-L68.

kanav99:
I have applied your suggestions, let us see if there is any issue in that.

PR #608, opened 2020-12-02
Prettify MTK generated names

YingboMa:
Ref: https://github.com/SciML/ModelingToolkit.jl/pull/679
Fix https://github.com/SciML/ModelingToolkit.jl/issues/636


YingboMa:
plt

ChrisRackauckas (approved):

PR #609, opened 2020-12-02
Fix plot legends when syms already have dependent variables

YingboMa:
Fix https://github.com/SciML/ModelingToolkit.jl/issues/636

plt1
plt2


isaacsas:
Thanks both!

PR #617, opened 2020-12-10
Fix docstring for deleteat!

imciner2:
There was a stray newline between the docstring and the function which seems to have caused the docstring to not be associated with the function. This means that the docs site can't find the docstring and shows an error instead.


ChrisRackauckas:
Thanks!

PR #621, opened 2020-12-16
Extended recursive_length to operate on Number type

jdeldre:
This ensures that it does not fail for ArrayPartitions of Numbers. Will be helpful for fixing https://github.com/SciML/OrdinaryDiffEq.jl/issues/1308

Also, added ode_default_norm tests to `runtests.jl`. Previously missing.

### DIFF
--- a/src/common_defaults.jl
+++ b/src/common_defaults.jl
@@ -3,6 +3,7 @@
 @inline UNITLESS_ABS2(x::ArrayPartition) = sum(UNITLESS_ABS2, x.x)
 
 @inline recursive_length(u::AbstractArray{<:Number}) = length(u)
+@inline recursive_length(u::Number) = 1
 @inline recursive_length(u::AbstractArray{<:AbstractArray}) = sum(recursive_length, u)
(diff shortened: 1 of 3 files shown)


devmotion:
Seems reasonable and is consistent with `length(::Number)`. Maybe add a separate test for `recursive_length(::Number)`?

Did you check if this fixes all test errors in https://github.com/SciML/OrdinaryDiffEq.jl/issues/1308?

jdeldre:
@devmotion Probably better to use `length(::Number)` explicitly, which I do in this commit. I also added a test. Still testing that it fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1308 but looks good so far.

jdeldre:
Indeed, this does fix SciML/OrdinaryDiffEq.jl#1308. I've now added additional tests in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1309 with a representative case similar to the one that triggered the issue.

devmotion:
Great!

devmotion (approved):

devmotion:
Thanks a lot, I'll make a new release right away.

PR #624, opened 2020-12-27
Implement Base.IteratorSize for integrator iterators

manuel-rhdt:
By default these implementations return `Base.SizeUnknown()` since in general we do not know in advance how many steps there will be before an iterator finishes.

Without these new implementations a call to `Base.collect` e.g. on `IntegratorTuples` fails with `ERROR: MethodError: no method matching length` since by default `collect` assumes that all iterators implement `length` (see https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration ).

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -416,6 +416,7 @@ function Base.iterate(tup::IntegratorTuples, state=0)
   return (tup.integrator.u,tup.integrator.t),state
 end
 
+Base.eltype(::Type{IntegratorTuples{I}}) where {U, T, I<:DEIntegrator{<:Any, <:Any, U, T}} = Tuple{U, T} 
 Base.IteratorSize(::Type{<:IntegratorTuples}) = Base.SizeUnknown()
 
 RecursiveArrayTools.tuples(integrator::DEIntegrator) = IntegratorTuples(integrator)
@@ -435,6 +436,7 @@ function Base.iterate(tup::IntegratorIntervals,state=0)
   return (tup.integrator.uprev,tup.integrator.tprev,tup.integrator.u,tup.integrator.t),state
 end
 
+Base.eltype(::Type{IntegratorIntervals{I}}) where {U, T, I<:DEIntegrator{<:Any, <:Any, U, T}} = Tuple{U, T, U, T} 
 Base.IteratorSize(::Type{<:IntegratorIntervals}) = Base.SizeUnknown()
 
 intervals(integrator::DEIntegrator) = IntegratorIntervals(integrator)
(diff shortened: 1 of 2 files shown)


src/integrator_interface.jl
```diff
@@ -400,6 +400,7 @@ function Base.iterate(integrator::DEIntegrator,state=0)
 end
 
 Base.eltype(integrator::DEIntegrator) = typeof(integrator)
+Base.IteratorSize(::Type{T}) where T<:DEIntegrator = Base.SizeUnknown()
```
devmotion, line 403: It seems the type parameter is not needed:
```suggestion
Base.IteratorSize(::Type{<:DEIntegrator}) = Base.SizeUnknown()
```

src/integrator_interface.jl
```diff
@@ -400,6 +400,7 @@ function Base.iterate(integrator::DEIntegrator,state=0)
 end
 
 Base.eltype(integrator::DEIntegrator) = typeof(integrator)
```
devmotion, line 402: BTW, while you're at it, `eltype` should be defined for types:
```suggestion
Base.eltype(::Type{T}) where {T<:DEIntegrator} = T
```
I'm also wondering how useful (and reasonable) this definition is at all?

src/integrator_interface.jl
```diff
@@ -415,6 +416,8 @@ function Base.iterate(tup::IntegratorTuples, state=0)
   return (tup.integrator.u,tup.integrator.t),state
 end
 
+Base.IteratorSize(::Type{IntegratorTuples{T}}) where T<:DEIntegrator = Base.SizeUnknown()
```
devmotion, line 419: Same here:
```suggestion
Base.IteratorSize(::Type{<:IntegratorTuples}) = Base.SizeUnknown()
```
It seems `eltype` is not defined here?

src/integrator_interface.jl
```diff
@@ -432,6 +435,8 @@ function Base.iterate(tup::IntegratorIntervals,state=0)
   return (tup.integrator.uprev,tup.integrator.tprev,tup.integrator.u,tup.integrator.t),state
 end
 
+Base.IteratorSize(::Type{IntegratorIntervals{T}}) where T<:DEIntegrator = Base.SizeUnknown()
```
devmotion, line 438: Same here:
```suggestion
Base.IteratorSize(::Type{<:IntegratorIntervals}) = Base.SizeUnknown()
```

devmotion:
It would be good to add some tests to ensure that e.g. `collect` works as expected.

manuel-rhdt:
I added a few tests to ensure `tuples` and `intervals` iterators correctly follow the iterator interface.

ChrisRackauckas:
Thanks!

PR #625, opened 2020-12-30
update IterativeSolver bindings for breaking tol change

ChrisRackauckas:
@yingboma should it be abstol or reltol?


YingboMa (commented):
I think either is fine. Since we used `tol=1e-16` before, it won't change the behavior. I think we definitely need to tweak the tolerances later.

PR #626, opened 2021-01-01
Make linear iterative solver more robust

YingboMa:
Fix https://github.com/SciML/OrdinaryDiffEq.jl/issues/1316

I will post a change to disable the non-adaptive forwarding on the OrdinaryDiffEq side later.

### DIFF
--- a/src/linear_nonlinear.jl
+++ b/src/linear_nonlinear.jl
@@ -52,8 +52,14 @@ mutable struct DefaultLinSolve
   iterable
 end
 DefaultLinSolve() = DefaultLinSolve(nothing, nothing)
+@noinline function checkreltol(reltol)
+  if !(reltol isa Real)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

src/linear_nonlinear.jl
```diff
@@ -52,8 +52,15 @@ mutable struct DefaultLinSolve
   iterable
 end
 DefaultLinSolve() = DefaultLinSolve(nothing, nothing)
+@noinline function checkreltol(reltol)
+  if !(reltol isa Real)
+    reltol = real(first(reltol))
+    @warn "Non real valued reltol is not supported by the iterative solvers. `reltol` for the linear solver is set to be `real(reltol[1])`. To customize tolerances for the linear iterable solvers, use `KenCarp3(linsolve=LinSolveGMRES(abstol=1e-16,reltol=1e-16))`."
```
YingboMa, line 58: @ChrisRackauckas this is kind of terrible in the sense that this is printed once in every Newton iteration. Do we want to change it?

ChrisRackauckas: should it just error?

YingboMa: Yeah, error would be better.

PR #631, opened 2021-01-12
Pass tolerance onto nonlinear solve

ChrisRackauckas:


YingboMa (approved):

YingboMa:
It seems that NonlinearSolve.jl doesn't take reltol and abstol kwargs. https://github.com/JuliaComputing/NonlinearSolve.jl/blob/master/src/solve.jl#L9-L13

ChrisRackauckas:
we need to unify the arguments across that package.

PR #632, opened 2021-01-12
Handle callback dt at t=0 separately

ChrisRackauckas:
This is a simple fix to the zero callback in https://discourse.julialang.org/t/diffeq-tolerances-on-continuouscallbacks/53181

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -58,9 +58,9 @@ Contains a single callback whose `condition` is a continuous function. The callb
   negative). For more information on what can
   be done, see the [Integrator Interface](@ref integrator) manual page. Modifications to
   `u` are safe in this function.
-- `rootfind=LeftRootFind`: This is a flag to specify the type of rootfinding to do for finding 
+- `rootfind=LeftRootFind`: This is a flag to specify the type of rootfinding to do for finding
   event location. If this is set to `LeftRootfind`, the solution will be backtracked to the point where
-  `condition==0` and if the solution isn't exact, the left limit of root is used. If set to 
+  `condition==0` and if the solution isn't exact, the left limit of root is used. If set to
   `RightRootFind`, the solution would be set to the right limit of the root. Otherwise the systems and
   the `affect!` will occur at `t+dt`.
 - `interp_points=10`: The number of interpolated points to check the condition. The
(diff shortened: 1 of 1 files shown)


src/callbacks.jl
```diff
@@ -518,7 +518,11 @@ end
     end
 
     # Evaluate condition slightly in future
-    abst = integrator.tprev+integrator.tdir*max(abs(integrator.dt/10000),100*eps(integrator.t))
+    if integrator.t == 0
```
devmotion, line 521: Why test for zero here specifically and not the initial time point? And it seems maybe the problem could also be solved without this branch by specifying a smaller `dt`?

ChrisRackauckas: The problem is due to `eps(integrator.t)` at `t==0` being absurdly small.

devmotion: Oh I got the impression the problem was that `abst` was too large before from the discussion on discourse. Since you removed one part of the `max` call the new value should be less than or equal to the old value - but if it the `eps` part was too small it should actually be the same, shouldn't it?

ChrisRackauckas: The only reason that other part is there is because of you use `eps(integrator.t)` at zero you get:

```julia
julia> eps(0.0)
5.0e-324
```

which isn't helpful. At all other spots it's adapted to the relative size of the values in the problem, so it's a good estimate. So that `max` was there to cover that other case. But it seems like `integrator.dt/10000` ended up dominating the `max` even when it shouldn't, so I've relegated it to exactly the one spot where it matters.

devmotion: Ahh now I understand, thanks!

klaff:
Pardon my ignorance, but is this the right way to pick up the changes in master branch of DiffEqBase?
```
begin # define independent package environment
	import Pkg
	Pkg.activate(mktempdir())
	Pkg.add("Plots"); using Plots
	Pkg.add(name="DiffEqBase",rev="master");
	Pkg.add(name="DifferentialEquations"); using DifferentialEquations
	Pkg.add("PlutoUI"); using PlutoUI
end
```
I'm not seeing a change in the behavior.

ChrisRackauckas:
I just tagged. Get the new version.

klaff:
Much better on the bouncing ball. I still have a similar failure in my "real" simulation which uses VectorContinuousCallback. I haven't found an error in my code yet but it's much more complicated so not a good MWE. Tomorrow I'll make a version of the bouncing ball MWE using VectorContinousCallback and see what happens. If that works fine I'll try to cook my other failing model down.  Thanks!

klaff:
Hi Chris, I have another example which still fails with DiffEqBase = v6.53.5. Could still be my error but I've stared at it too long and need some other eyes on it. I would also like to discuss whether I'm approaching this in a reasonable fashion or am I trying to use a screwdriver as a hammer.

Where would be best to post this example? (Here, new issue, new discourse thread, ?)  Thanks.

ChrisRackauckas:
New issue.

PR #635, opened 2021-01-23
Add get_concrete_problem for NonlinearProblem

utkarsh530:
Add support for declaring NonlinearProblem. Otherwise, `tspan` argument was missing error was showing.
@ChrisRackauckas, please review.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -119,6 +119,12 @@ function get_concrete_problem(prob::AbstractSteadyStateProblem, isadapt; kwargs.
   remake(prob; u0 = u0)
 end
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
You'll likely also need something like:

https://github.com/SciML/DiffEqBase.jl/blob/master/src/diffeqfunction.jl#L18-L36
https://github.com/SciML/DiffEqBase.jl/blob/master/src/diffeqfunction.jl#L312-L395

for https://github.com/SciML/DiffEqBase.jl/blob/master/src/problems/basic_problems.jl#L26-L39

utkarsh530:
Okay, I would declare them under a `NonlinearFunction`. I was currently not including these parameters (color, sparsity, jac_prototype) etc.

Issue #628, opened 2021-01-06
Missing implementation of integrator init for PDEProblem


bgroenks96:
Trying to use the integrator interface via `init` on a `PDEProblem` results in the following error:

```
type PDEProblem has no field p
in top-level scope at heat_conduction.jl:59
in  at DiffEqBase/znIav/src/solve.jl:37
in #init#446 at DiffEqBase/znIav/src/solve.jl:43
in  at DiffEqBase/znIav/src/solve.jl:135
in #get_concrete_problem#457 at DiffEqBase/znIav/src/solve.jl:135
in get_concrete_p at DiffEqBase/znIav/src/solve.jl:208
in getproperty at base/Base.jl:33
```

This can be corrected by adding the following function implementation to `solve.jl`, mirroring that of `solve`:

```julia
function DiffEqBase.init(prob::PDEProblem,alg::DiffEqBase.DEAlgorithm,args...;
                                          kwargs...)
    init(prob.prob,alg,args...;kwargs...)
end
```

Pull request to follow shortly.

ChrisRackauckas:
There is no integrator for a PDEProblem. This won't work. It needs to be `discretize`d

bgroenks96:
I don't know what you mean. The result of `discretize` is a `PDEProblem` with a nested `ODEProblem`. The implementation of `init` should, as far as I can tell, look the same as the one for `solve`.

I tested this on a `PDEProblem` obtained and from `discretize` and it works exactly as expected.

devmotion:
In any case, the `DiffEqBase` qualifications could be removed both in `solve` and a potential implementation of `init`. I assume the code was just copied from an external package at some point.

bgroenks96:
@devmotion I was wondering why that was there!

bgroenks96:
Fixed by #636

PR #636, opened 2021-01-24
Add init wrapper for discretized PDEProblem

bgroenks96:
Adds an `init` wrapper function for a discretized `PDEProblem` identical to what already exists for `solve`.


bgroenks96:
Fixes #628

PR #637, opened 2021-01-25
Add jacobian structures for NonlinearProblem

utkarsh530:
Hi,

I have added parameters for jacobian calculation and helper functions for NonlinearProblem similar to ODEFunction and as pointed out in #635.

@ChrisRackauckas, please review.

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -612,7 +612,7 @@ export remake
 export DEDataArray, DEDataVector, DEDataMatrix
 
 export ODEFunction, DiscreteFunction, SplitFunction, DAEFunction, DDEFunction,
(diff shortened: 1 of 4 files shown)


ChrisRackauckas (approved):

utkarsh530:
@ChrisRackauckas I guess I do need to add a JacobianWrapper too. https://github.com/SciML/DiffEqBase.jl/blob/master/src/linear_nonlinear.jl#L124-L131 this has a dependency on this https://github.com/SciML/DiffEqBase.jl/blob/master/src/diffeqfunction.jl#L1017-L1024. I had currently added a custom jacobian wrapper for `NonlinearFunction` locally.  So I guess we need that too in DiffEqBase too.

PR #638, opened 2021-01-26
Split SciMLBase from DiffEqBase

ChrisRackauckas:
This is the companion PR to https://github.com/SciML/SciMLBase.jl/pull/1


src/init.jl
```diff
@@ -251,7 +251,7 @@ function __init__()
   end
 
   @require GeneralizedGenerated="6b9d7cbe-bcb9-11e9-073f-15a7a543e2eb" begin
-    numargs(::GeneralizedGenerated.RuntimeFn{Args}) where Args = GeneralizedGenerated.from_type(Args) |> length
+    SciMLBase.numargs(::GeneralizedGenerated.RuntimeFn{Args}) where Args = GeneralizedGenerated.from_type(Args) |> length
   end
```
devmotion, line 255: Maybe this could be moved to `SciMLBase`? It seems useful also for non-DE problems.

ChrisRackauckas: yeah, in another iteration. I'm more worried about making sure nothing broke with what was moved already.

PR #657, opened 2021-04-15
Using Setfield to set appropriate jac type

### LINKED ISSUE #656: Update jac_prototype to have same eltype as u0
See comments in https://github.com/SciML/ModelingToolkit.jl/pull/968 related to issue https://github.com/SciML/ModelingToolkit.jl/issues/871

akashkgarg:
Resolves #656


src/solve.jl
```diff
@@ -191,7 +190,15 @@ function get_concrete_problem(prob::DDEProblem, isadapt; kwargs...)
   remake(prob; u0 = u0, tspan = tspan, p=p, constant_lags = constant_lags)
 end
 
-promote_f(f,u0) = f
+promote_f(f,u0) = begin
```
ChrisRackauckas, line 193:
```suggestion
function promote_f(f,u0)
```

src/solve.jl
```diff
@@ -191,7 +190,15 @@ function get_concrete_problem(prob::DDEProblem, isadapt; kwargs...)
   remake(prob; u0 = u0, tspan = tspan, p=p, constant_lags = constant_lags)
 end
 
-promote_f(f,u0) = f
+function promote_f(f,u0)
+    # Ensure our jacobian will be of the same type as u0
+    uElType = isnothing(u0) ? Float64 : eltype(u0)
```
YingboMa, line 195:
```suggestion
    uElType = u0 === nothing ? Float64 : eltype(u0)
```

Project.toml
```diff
@@ -22,6 +22,7 @@ RecursiveFactorization = "f2c3362d-daeb-58d1-803e-2bc74f2840b4"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
```
devmotion, line 25: Can you add a compat bound for it? Otherwise it won't be merged into the registry automatically.

devmotion:
FYI there's an experimental (?) successor to Setfield: https://github.com/JuliaObjects/Accessors.jl Setfield is in maintenance only mode but probably it is safer to use it since, according to the README, in Accessors the developers "may introduce breaking changes without clear deprecation path".

ChrisRackauckas:
Looks like a real CI failure.

ChrisRackauckas:
>FYI there's an experimental (?) successor to Setfield: https://github.com/JuliaObjects/Accessors.jl Setfield is in maintenance only mode but probably it is safer to use it since, according to the README, in Accessors the developers "may introduce breaking changes without clear deprecation path".

Okay, let's give it a try. I didn't know that existed.

YingboMa:
This would need https://github.com/SciML/SciMLBase.jl/pull/54. Also, Julia is case-sensitive, so `uElType` is not `uEltype`.

src/solve.jl
```diff
@@ -191,7 +190,15 @@ function get_concrete_problem(prob::DDEProblem, isadapt; kwargs...)
   remake(prob; u0 = u0, tspan = tspan, p=p, constant_lags = constant_lags)
 end
 
-promote_f(f,u0) = f
+function promote_f(f,u0)
+    # Ensure our jacobian will be of the same type as u0
+    uElType = u0 === nothing ? Float64 : eltype(u0)
+    if isdefined(f, :jac_prototype) && f.jac_prototype isa AbstractArray
+        f = @set f.jac_prototype = similar(f.jac_prototype, uEltype)
```
ChrisRackauckas, line 197:
```suggestion
        f = @set f.jac_prototype = similar(f.jac_prototype, uElType)
```

akashkgarg:
@ChrisRackauckas  did you want me to replace `Setfield` with the new `Accessors.jl`?

ChrisRackauckas:
I think it's fine either way. Let's get an issue open to track that, but I don't think it's the highest priority thing. Though I do wonder if @tkf wants to fully replace the `remake` machinery with Accessors.jl soon.

YingboMa:
How is Accessors better than Setfield?

PR #667, opened 2021-05-23
Add asymmetric fastpow overloads

ChrisRackauckas:
Technically not needed for anything, but noticed while debugging https://github.com/SciML/OrdinaryDiffEq.jl/issues/1414 as potentially useful.


ChrisRackauckas:
Not having this has proved to be pretty useful for debugging too though.

YingboMa:
If those method should not be invoked, maybe don't add them? `fastpow` is internal and only used in the time step control.

ChrisRackauckas:
Yes, the plan now is to not merge.

PR #672, opened 2021-06-17
Use a nudge factor to remove repeats

ChrisRackauckas:
Sometimes simple is better. Instead of using a derivative estimate, now we just check at t + dt * repeat_nudge to avoid repeated zeros, with a default to 1//100 which seems to be a nice balance between numerical stability and keeping most of the interval still an eventful zone. 

Fixes https://github.com/SciML/DifferentialEquations.jl/issues/758
Fixes SciML/DifferentialEquations.jl#647
Fixes SciML/DifferentialEquations.jl#724
Fixes SciML/DifferentialEquations.jl#601
Fixes SciML/DifferentialEquations.jl#642
Fixes https://github.com/SciML/DiffEqBase.jl/issues/599

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -528,7 +528,7 @@ end
     abst = integrator.tprev+integrator.dt * callback.repeat_nudge
     tmp_condition = get_condition(integrator, callback, abst)
     @. prev_sign = sign(previous_condition)
-    prev_sign[ivec] = tmp_condition[ivec]
+    prev_sign[ivec] = sign(tmp_condition[ivec])
   else
       @. prev_sign = sign(previous_condition)
   end
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -528,7 +528,7 @@ end
     abst = integrator.tprev+integrator.tdir*integrator.dt * callback.repeat_nudge
     tmp_condition = get_condition(integrator, callback, abst)
     @. prev_sign = sign(previous_condition)
-    prev_sign[ivec] .= tmp_condition[ivec]
+    prev_sign[ivec] = tmp_condition[ivec]
   else
       @. prev_sign = sign(previous_condition)
   end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/647

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/724

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/601

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/642

YingboMa (approved):
It's impressive that this works so well.

src/callbacks.jl
```diff
@@ -518,20 +525,10 @@ end
...
     tmp_condition = get_condition(integrator, callback, abst)
-
-    # Sometimes users may "switch off" the condition after crossing
-    # This is necessary to ensure proper non-detection of a root
-    # == is for exact floating point equality!
     @. prev_sign = sign(previous_condition)
-    prev_sign[ivec] = tmp_condition[ivec] > previous_condition[ivec] ? 1.0 :
-                  (tmp_condition[ivec] == previous_condition[ivec] ?
-                  (prev_sign[ivec] = sign(previous_condition[ivec])) : -1.0)
+    prev_sign[ivec] = tmp_condition[ivec]
```
devmotion, line 531: Should this be

```suggestion
    prev_sign[ivec] = sign(tmp_condition[ivec])
```
?

devmotion:
BTW the GPU tiemouts are caused by CUDA 3.3 but fixed on master: https://julialang.slack.com/archives/C689Y34LE/p1623939645188500 AFAIK a new release will be available soon (possibly today).

ChrisRackauckas:
Awesome, those were confusing me but I at least figured out it was unrelated to this (because those don't test callbacks)

PR #674, opened 2021-06-21
chainrulez

ChrisRackauckas:


ChrisRackauckas:
Does another piece of https://github.com/SciML/SciMLBase.jl/issues/69.

@oxinabox what's the right translation for something like `ZygoteRules.literal_getproperty(sol::AbstractTimeseriesSolution, ::Val{:u})`?

PR #679, opened 2021-07-03
New approach for speeding up sensefun

YingboMa:


YingboMa:
Don't merge these PRs just yet. I am still testing them locally.

ChrisRackauckas:
This code is stale since it's not in here anymore. It would need to be rewritten as overloads to LinearSolve.jl

PR #689, opened 2021-07-31
remove custom exp2 implimentation

oscardssmith:
1.7 has a `@fastmath exp2` that's 20% faster and more accurate than the version that was defined here. It is a little worse in the case of overflow, so this should get some careful testing.


YingboMa:
Let's hold on this until 1.7 release, since there's no way to testing this current.

oscardssmith:
Lets try this now that 1.7 is out!

oscardssmith:
@YingboMa any idea why CI is failing?

YingboMa:
I think it's just because some tests are too stringent. Do you mind to dev it locally and relax the tests?

ChrisRackauckas:
First, we need to finishing updating to v1.7. Remaining issues:

https://github.com/SciML/StochasticDiffEq.jl/pull/445
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1531
https://github.com/SciML/StochasticDelayDiffEq.jl/pull/47
https://github.com/SciML/DiffEqNoiseProcess.jl/pull/97

oscardssmith:
Oh, can one of you ping me once those are resolved?

oscardssmith:
the first test fail is
```
  Expression: sol(30) == solap(30)
   Evaluated: [13.324178410828372, 13.734216629173758, 32.956664457899734] == RecursiveArrayTools.ArrayPartition{Float64, Tuple{StaticArrays.SVector{2, Float64}, StaticArrays.SVector{1, Float64}}}(([13.324178410828374, 13.734216629173758], [32.956664457899734]))
``` should this be using approx?

ChrisRackauckas:
Yeah, I just did a patch which changes it to approx. It was working before because I guess broadcast just happened to do the math exactly the same in both versions, but some change in v1.7 changes some summation order so now it's different by 1e-15. Your PR has kind of turned into a test of all things happening in v1.7 + the breaking OrdinaryDiffEq cleanup changes 😅

oscardssmith:
Thanks for fixing all the tests!

PR #693, opened 2021-08-05
Update openblas check for v1.7

ChrisRackauckas:
@staticfloat @yingboma

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -58,7 +58,7 @@ SciMLBase = "1.13"
 Setfield = "0.7"
 StaticArrays = "0.11, 0.12, 1.0"
 ZygoteRules = "0.2"
-julia = "1.6"
+julia = "1.7"
 
 [extras]
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
--- a/src/linear_nonlinear.jl
+++ b/src/linear_nonlinear.jl
@@ -71,13 +71,23 @@ Base.resize!(p::LinSolveGPUFactorize,i) = p.A = nothing
 mutable struct DefaultLinSolve
   A
   iterable
+  openblas::Union{Bool, Nothing}
 end
-DefaultLinSolve() = DefaultLinSolve(nothing, nothing)
+DefaultLinSolve() = DefaultLinSolve(nothing, nothing, nothing)
 @noinline function checkreltol(reltol)
   if !(reltol isa Real)
     error("Non real valued reltol is not supported by the linear iterative solvers. To customize tolerances for the linear iterative solvers, use the syntax like `KenCarp3(linsolve=LinSolveGMRES(abstol=1e-16,reltol=1e-16))`.")
   end
   return reltol
+end
+
+function isopenblas()
+    @static if VERSION < v"1.7"
+      blas = BLAS.vendor()
+      blas == :openblas64 || blas == :openblas
+    else
+      occursin("openblas", BLAS.get_config().loaded_libs[1].libname)
+    end
 end
 
 function (p::DefaultLinSolve)(x,A,b,update_matrix=false;reltol=nothing, kwargs...)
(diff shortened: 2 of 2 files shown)


src/linear_nonlinear.jl
```diff
@@ -71,15 +71,22 @@ Base.resize!(p::LinSolveGPUFactorize,i) = p.A = nothing
...
   if !(reltol isa Real)
     error("Non real valued reltol is not supported by the linear iterative solvers. To customize tolerances for the linear iterative solvers, use the syntax like `KenCarp3(linsolve=LinSolveGMRES(abstol=1e-16,reltol=1e-16))`.")
   end
   return reltol
 end
 
+function isopenblas()
+    splitter = Base.Sys.iswindows() ? "\\" : "/"
+    str = split(BLAS.get_config().loaded_libs[1].libname,splitter)[end]
+    str == "libopenblas64_.dll" || str == "libopenblas_.dll"
```
YingboMa, line 87:
```suggestion
    occursin("openblas", str)
```

src/linear_nonlinear.jl
```diff
@@ -71,15 +71,22 @@ Base.resize!(p::LinSolveGPUFactorize,i) = p.A = nothing
 mutable struct DefaultLinSolve
   A
   iterable
+  openblas
```
YingboMa, line 74:
```suggestion
  openblas::Bool
```

ChrisRackauckas: the initialization couldn't be nothing then. It should be Union{Nothing,Bool}

src/linear_nonlinear.jl
```diff
@@ -71,15 +71,22 @@ Base.resize!(p::LinSolveGPUFactorize,i) = p.A = nothing
 mutable struct DefaultLinSolve
   A
   iterable
+  openblas::Bool
```
YingboMa, line 74:
```suggestion
  openblas::Union{Bool, Nothing}
```

src/linear_nonlinear.jl
ChrisRackauckas: wait until someone make a BLAS called "atleastitsbetterthanopenblas".

YingboMa:
Remember to squash when merging it :p

ChrisRackauckas:
This can't merge until v1.7 release, so we should wait a bit, but at least it's ready.

YingboMa:
Or, we could just add a version check so this is backward compatible.

ChrisRackauckas:
true, I'll do that.

src/linear_nonlinear.jl
```diff
@@ -82,9 +82,14 @@ DefaultLinSolve() = DefaultLinSolve(nothing, nothing, nothing)
 end
 
 function isopenblas()
-    splitter = Base.Sys.iswindows() ? "\\" : "/"
-    str = split(BLAS.get_config().loaded_libs[1].libname,splitter)[end]
-    occursin("openblas", str)
+    if VERSION < v"1.7"
```
YingboMa, line 85:
```suggestion
    @static if VERSION < v"1.7"
```

src/linear_nonlinear.jl
```diff
@@ -71,15 +71,27 @@ Base.resize!(p::LinSolveGPUFactorize,i) = p.A = nothing
...
   return reltol
 end
 
+function isopenblas()
+    if VERSION < v"1.7"
+      blas = BLAS.vendor()
+      blas == :openblas64 || blas == :openblas
+    else
+      splitter = Base.Sys.iswindows() ? "\\" : "/"
+      str = split(BLAS.get_config().loaded_libs[1].libname,splitter)[end]
```
YingboMa, line 90: With `occursin`, you won't need to use `split`

ChrisRackauckas: oops

YingboMa (approved):

Project.toml
```diff
@@ -58,7 +58,7 @@ SciMLBase = "1.13"
 Setfield = "0.7"
 StaticArrays = "0.11, 0.12, 1.0"
 ZygoteRules = "0.2"
-julia = "1.6"
+julia = "1.7"
```
YingboMa, line 61: This change probably isn't needed anymore

ChrisRackauckas: true

YingboMa:
Let's wait the release of 1.7 so that the CI can properly test it.

ChrisRackauckas:
ehh I've been testing it locally in its various forms for days on a few different systems. It should at least be fine enough to release to users who are on the beta.

PR #694, opened 2021-08-05
update for RecursiveFactorization v0.2

ChrisRackauckas:


src/linear_nonlinear.jl
```diff
@@ -92,9 +92,8 @@ function (p::DefaultLinSolve)(x,A,b,update_matrix=false;reltol=nothing, kwargs..
       # if the user doesn't use OpenBLAS, we assume that is a better BLAS
       # implementation like MKL
       #
-      # RecursiveFactorization seems to be consistantly winning below 100
-      # https://discourse.julialang.org/t/ann-recursivefactorization-jl/39213
-      if ArrayInterface.can_setindex(x) && (size(A,1) <= 100 || ((blasvendor === :openblas || blasvendor === :openblas64) && size(A,1) <= 500))
+      # OpenBLAS sucks
+      if ArrayInterface.can_setindex(x) && (size(A,1) <= 100 || blasvendor === :openblas || blasvendor === :openblas64)
```
YingboMa, line 96: It only sucks for newer hardwares. This is a bold move :p Are you sure to do this?

YingboMa (approved):
LGTM, if you are comfortable with this config

src/linear_nonlinear.jl
ChrisRackauckas: I think so, because even when it's okay we seem to at least match it. This has at least a lot more stable performance, whereas OpenBLAS seems to just fall apart and doesn't even set the default threads correctly.

PR #712, opened 2021-10-29
Type stable NamedTuple creation

chriselrod:
Current version allocates memory by creating a vector instead of a tuple so that the subsequent named tuple can be type unstable. I switched to creating a tuple, instead.


ChrisRackauckas:
This one too? https://github.com/SciML/DiffEqBase.jl/pull/712/files#diff-ca1c45d4f0bec64e8340981f6b27f4037bf6ac13c8f1af132656c485c0a8e941R19

chriselrod:
I think I got them all.

PR #719, opened 2021-12-08
Use TriangularSolve.jl for the ldiv!

ChrisRackauckas:
@yingboma @chriselrod is this a good idea?

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -27,6 +27,7 @@ import Distributions
 import ChainRulesCore
 import LabelledArrays
 import RecursiveArrayTools
+import TriangularSolve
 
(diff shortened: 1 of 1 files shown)


YingboMa:
Less I am missing something, this PR doesn't seem to change anything. `import` doesn't bring anything to the scope except the module name.

chriselrod:
`TriangularSolve.jl` also only currently supports 2/4 `trsm!` permutations.
Which permutations are used here?

YingboMa:
I think the only `ldiv!` usage here is `trsv!`

chriselrod:
> I think the only `ldiv!` usage here is `trsv!`, so there is probably not much improvement.

I haven't implemented `trsv!` in TriangularSolve.jl.

PR #723, opened 2021-12-22
Remove mapreduce_empty

ChrisRackauckas:
Suggestion from https://github.com/SciML/DiffEqBase.jl/issues/664 . Requires v1.6 and we're in the future.

### DIFF
--- a/src/common_defaults.jl
+++ b/src/common_defaults.jl
@@ -1,7 +1,6 @@
 @inline UNITLESS_ABS2(x::Number) = abs2(x)
-@inline UNITLESS_ABS2(x::AbstractArray) = sum(UNITLESS_ABS2, x)
-@inline UNITLESS_ABS2(x::RecursiveArrayTools.ArrayPartition) = sum(UNITLESS_ABS2, x.x)
-Base.mapreduce_empty(::typeof(UNITLESS_ABS2), op, T) = abs2(Base.reduce_empty(op, T))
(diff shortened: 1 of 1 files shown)


src/common_defaults.jl
```diff
@@ -1,7 +1,8 @@
+abs2_and_sum(x,y) = reduce(Base.add_sum,x,init=zero(real(value(eltype(x))))) +
+                    reduce(Base.add_sum,y,init=zero(real(value(eltype(y)))))
 @inline UNITLESS_ABS2(x::Number) = abs2(x)
-@inline UNITLESS_ABS2(x::AbstractArray) = sum(UNITLESS_ABS2, x)
-@inline UNITLESS_ABS2(x::RecursiveArrayTools.ArrayPartition) = sum(UNITLESS_ABS2, x.x)
-Base.mapreduce_empty(::typeof(UNITLESS_ABS2), op, T) = abs2(Base.reduce_empty(op, T))
+@inline UNITLESS_ABS2(x::AbstractArray) = mapreduce(UNITLESS_ABS2,abs2_and_sum, x, init=zero(real(value(eltype(x)))))
+@inline UNITLESS_ABS2(x::RecursiveArrayTools.ArrayPartition) = mapreduce(UNITLESS_ABS2, abs2_and_sum, x.x, init=zero(real(value(eltype(x)))))
```
devmotion, line 5: Something to be aware of: `mapreduce` with `Base.add_sum` is specialized and uses pairwise summation whereas for general operators (and even `+` IIRC) it doesn't.

Issue #730, opened 2022-01-30
UNITLESS_ABS2 not accepting mixed types after #723


hustf:
Updating to after #723 introduces this bug in my context:

```
julia> rv1 = ArrayPartition([1.0kg, 2.0N, 3.0m/s, 4.0m/s], [1.0kg/s, 2.0N/s, 3.0m/s^2, 4m/s^2])
([1.0kg, 2.0N, 3.0m∙s⁻¹, 4.0m∙s⁻¹], [1.0kg∙s⁻¹, 2.0N∙s⁻¹, 3.0m∙s⁻², 4.0m∙s⁻²])

julia> UNITLESS_ABS2(rv1)
ERROR: ArgumentError: zero(Quantity{Float64}) not defined.
Stacktrace:
 [1] zero(x::Type{Quantity{Float64}})
   @ Unitfu C:\Users\frohu_h4g8g6y\.julia\packages\Unitfu\GkUxI\src\quantities.jl:391
 [2] real(T::Type)
   @ Base .\complex.jl:120
 [3] UNITLESS_ABS2(x::ArrayPartition{Quantity{Float64}, Tuple{Vector{Quantity{Float64,  ᴹ, Unitfu.FreeUnits{(kg,),  ᴹ, nothing}}}, Vector{Quantity{Float64,  ᴹ∙ ᵀ⁻¹, Unitfu.FreeUnits{(kg, s⁻¹),  ᴹ∙ ᵀ⁻¹, nothing}}}}})
   @ DiffEqBase C:\Users\frohu_h4g8g6y\.julia\packages\DiffEqBase\1V2xg\src\common_defaults.jl:5
 [4] top-level scope
   @ REPL[51]:1
```

Note that I have not actually tested this with 'Unitful.jl', as I'm using a fork. I believe there are good reasons to avoid defining a zero for quantities where the only common type parameter is Float64.  A catch-all for all the possible types would trigger a lot of specialized compilation and general slow-down, as well as secondary effects. 

I'm submitting a PR which avoids finding the most general zero quantity by using the first element type only. If the numeric representation of the mixed types can not actually be summed, that is the real problem and would trigger the appropriate error.

hustf:
As per review comment, `init=zero(real(value(eltype(x))))` won't work. For example, first(x) will fail on CuArrays.
[...]

PR #731, opened 2022-01-30
Base mapreduce initial zero on first element type, not all

hustf:
Ref. issue #730 and #723.

### DIFF
--- a/src/common_defaults.jl
+++ b/src/common_defaults.jl
@@ -1,5 +1,5 @@
-abs2_and_sum(x,y) = reduce(Base.add_sum,x,init=zero(real(value(eltype(x))))) +
-                    reduce(Base.add_sum,y,init=zero(real(value(eltype(y)))))
+abs2_and_sum(x,y) = reduce(Base.add_sum,x,init=zero(real(value(eltype(first(x)))))) +
+                    reduce(Base.add_sum,y,init=zero(real(value(eltype(first(y))))))
 @inline UNITLESS_ABS2(x::Number) = abs2(x)
 @inline UNITLESS_ABS2(x::AbstractArray) = mapreduce(UNITLESS_ABS2,abs2_and_sum, x, init=zero(real(value(eltype(first(x))))))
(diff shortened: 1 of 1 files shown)


src/common_defaults.jl
```diff
@@ -1,8 +1,8 @@
-abs2_and_sum(x,y) = reduce(Base.add_sum,x,init=zero(real(value(eltype(x))))) +
-                    reduce(Base.add_sum,y,init=zero(real(value(eltype(y)))))
+abs2_and_sum(x,y) = reduce(Base.add_sum,x,init=zero(real(value(eltype(first(x)))))) +
```
ChrisRackauckas, line 1: This is not possible in general. For example, `first(x)` will fail on CuArrays.

src/common_defaults.jl
```diff
@@ -1,8 +1,8 @@
-abs2_and_sum(x,y) = reduce(Base.add_sum,x,init=zero(real(value(eltype(x))))) +
-                    reduce(Base.add_sum,y,init=zero(real(value(eltype(y)))))
+abs2_and_sum(x,y) = reduce(Base.add_sum,x,init=zero(real(value(eltype(first(x)))))) +
+                    reduce(Base.add_sum,y,init=zero(real(value(eltype(first(y))))))
 @inline UNITLESS_ABS2(x::Number) = abs2(x)
-@inline UNITLESS_ABS2(x::AbstractArray) = mapreduce(UNITLESS_ABS2,abs2_and_sum, x, init=zero(real(value(eltype(x)))))
-@inline UNITLESS_ABS2(x::RecursiveArrayTools.ArrayPartition) = mapreduce(UNITLESS_ABS2, abs2_and_sum, x.x, init=zero(real(value(eltype(x)))))
+@inline UNITLESS_ABS2(x::AbstractArray) = mapreduce(UNITLESS_ABS2,abs2_and_sum, x, init=zero(real(value(eltype(first(x))))))
+@inline UNITLESS_ABS2(x::RecursiveArrayTools.ArrayPartition) = mapreduce(UNITLESS_ABS2, abs2_and_sum, x.x)
```
ChrisRackauckas, line 5: I think your issue needs a solution which works by specializing this function a bit more?

hustf: Yes, ideally a base function exists which extracts the numeric type. Will come back with an update.

PR #733, opened 2022-02-14
Generalizing NaN check for sparse matrices ...

goerch:
... and adding tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.jl.*.mem
 Manifest.toml
 .DS_Store
+.vscode/settings.json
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
It only seems to have fixed some downstream tests haha, nice.

rory-humphries:
Release [v6.81.4](https://github.com/SciML/DiffEqBase.jl/releases/tag/v6.81.4) says this was merged (I assume that implies into main). However, it doesn't look like it was.

ChrisRackauckas:
Interesting. I don't know what happened here.

ChrisRackauckas:
https://github.com/SciML/DiffEqBase.jl/pull/741

PR #735, opened 2022-03-14
Loosen main prob2dtmin dispatch

ChrisRackauckas:
Since Dual <: AbstractFloat == false, which was causing it to hit the rational fallback of `1//10^10` which didn't make much sense. This looser form will set a more sensible default to more cases. There is the fact that in theory `eps` requires AbstractFloat, so maybe expanding to `Real` isn't quite right as this can now try to call `eps` on some number types for which it doesn't exist, but in a practical sense this is more likely to be correct than the other way around. Careful checking of downstream tests will be required on this one to make sure there aren't any of these edge cases (example: TrackedReal) that rely on the current default to not error.


ChrisRackauckas:
Downstream failure in DelayDiffEq is just tests that are now passing. That's always nice!

PR #738, opened 2022-03-23
add NoInit to import from SciML

jd-lara:
Follow up to https://github.com/SciML/SciMLBase.jl/pull/145 to import `NoInit()`. Requires a tag in SciMLBase for tests to pass to set a minimum in the project toml


ChrisRackauckas:
Why is this one needed?

jd-lara:
Maybe is not, I was trying a similar pattern which seemed to me that the re-exported structs are always called in. Maybe I am wrong and it is not needed

ChrisRackauckas:
Just do it from SciMLBase.

PR #740, opened 2022-03-26
simplify and speedup NaNCheck

ChrisRackauckas:
From https://discourse.julialang.org/t/fastest-way-to-check-for-inf-or-nan-in-an-array/76954/22 @oscardssmith

### DIFF
--- a/src/common_defaults.jl
+++ b/src/common_defaults.jl
@@ -1,7 +1,7 @@
-abs2_and_sum(x,y) = reduce(Base.add_sum,x,init=zero(real(value(eltype(x))))) +
-                    reduce(Base.add_sum,y,init=zero(real(value(eltype(y)))))
+abs2_and_sum(x, y) = reduce(Base.add_sum, x, init=zero(real(value(eltype(x))))) +
+                     reduce(Base.add_sum, y, init=zero(real(value(eltype(y)))))
 @inline UNITLESS_ABS2(x::Number) = abs2(x)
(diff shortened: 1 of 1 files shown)


src/common_defaults.jl
```diff
@@ -10,38 +10,35 @@ abs2_and_sum(x,y) = reduce(Base.add_sum,x,init=zero(real(value(eltype(x))))) +
...
-@inline NAN_CHECK(x::Number) = isnan(x)
-@inline NAN_CHECK(x::Float64) = isnan(x) || (x>1e50)
-@inline NAN_CHECK(x::Enum) = false
-@inline NAN_CHECK(x::AbstractArray) = any(NAN_CHECK, x)
-@inline NAN_CHECK(x::RecursiveArrayTools.ArrayPartition) = any(NAN_CHECK, x.x)
 
-@inline ODE_DEFAULT_UNSTABLE_CHECK(dt,u,p,t) = false
-@inline ODE_DEFAULT_UNSTABLE_CHECK(dt,u::Union{Number,AbstractArray},p,t) = NAN_CHECK(u)
+@inline NAN_CHECK(x::Number) = !isfinite(x)
+@inline NAN_CHECK(x::AbstractArray) = !isfinite(sum(x))
```
oscardssmith, line 42: isn't this wrong? (sum can overflow)

ChrisRackauckas: Truth

YingboMa (approved):
LGTM. But maybe someone’s solution is actually that big?

YingboMa:
Is it actually faster?

PR #743, opened 2022-04-11
Add a common kwarg warning system

ChrisRackauckas:
This has been a long time coming. The DiffEq, and more generally, SciML interfaces have a set of common keyword arguments, but it's been fairly difficult to truly enforce. The main reason is because the keyword arguments are `kwargs...` splatted down to the solvers, and then it's the solver's job to handle it. However, with so many different problem types and everything, there has been a changing list of keyword arguments over time, and some solvers re-splat the keyword arguments because of how many different kwargs they can end up getting if, for example, one wasn't using controls for parameter estimation or something.

In pre-1.0 Julia, there wasn't a good fix for downstream users as if you would error on some keyword arguments, they would have to sanitize their list before dropping to solve, but that would make downstream `simulate` functions much harder to write since half of their list would be checking for DiffEq common arguments (a list that the library authors probably only know of a subset). However, this was solved in v1.0 with NamedTuples, i.e. libraries could allow a keyword argument and let users pass `diffeq_arguments = (reltol=1e-8,abstol=1e-6)` and then do `solve(prob,alg;diffeq_arguments)`. That would then pre-sanitize. So libraries have evolved but... DiffEq had kept back for compatibility.

This introduces a system for checking keyword arguments but for the first year we should keep it conservative. I know there are some common errors out there that have arisen from this, such as `solver`, `rel_tol`, `save_at`, etc. that have mostly been swept under the rug. This will now expose them, but to keep backwards compatbility it will default to throwing an error. A new keyword argument `kwargshandle` was added to allow for silencing the warning (i.e. the old behavior), or to change it to error. The goal is to throw enough errors at everyone that it will be deprecated and be a true error in about a year's time.

That would also give us enough time to improve the documentation. Right now this lives here in https://diffeq.sciml.ai/stable/basics/common_solver_opts/, but we need to finish unifying the documentation and then have a true global set of common keyword arguments that all of the interfaces agree on which are specified at https://scimlbase.sciml.ai/dev/.

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -28,7 +28,7 @@ import ChainRulesCore
 import LabelledArrays
 import RecursiveArrayTools
 
-import ChainRulesCore: NoTangent, @ignore_derivatives
+import ChainRulesCore: NoTangent, @non_differentiable
 import ZygoteRules
 
 import DEDataArrays: DEDataArray, DEDataVector, DEDataMatrix, copy_fields!
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -28,7 +28,7 @@ import ChainRulesCore
 import LabelledArrays
 import RecursiveArrayTools
 
-import ChainRulesCore: NoTangent
+import ChainRulesCore: NoTangent, @ignore_derivatives
 import ZygoteRules
 
 import DEDataArrays: DEDataArray, DEDataVector, DEDataMatrix, copy_fields!
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -151,4 +151,6 @@ export NLNewton, NLFunctional, NLAnderson
 
 export SensitivityADPassThrough
 
+export KeywordArgError, KeywordArgWarn, KeywordArgSilent
+
 end # module
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -61,6 +61,8 @@ const allowedkeywords = (
   :seed,
   :alg_hints,
   :kwargshandle,
+  :trajectories,
+  :batch_size,
 )
 
 const KWARGWARN_MESSAGE = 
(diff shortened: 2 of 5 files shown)


ChrisRackauckas:
Tests found that we should do https://github.com/SciML/SciMLBase.jl/pull/146 which was renamed sometime back in Julia v0.6.

YingboMa (approved):

ChrisRackauckas:
`@non_differentiable` threw 

```
ERROR: LoadError: Invalid use of `@non_differentiable`
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:33
  [2] var"@non_differentiable"(__source__::LineNumberNode, __module__::Module, sig_expr::Any)
    @ ChainRulesCore C:\Users\accou\.julia\packages\ChainRulesCore\RbX5a\src\rule_definition_tools.jl:363
  [3] include(mod::Module, _path::String)
    @ Base .\Base.jl:418
  [4] include(x::String)
    @ DiffEqBase C:\Users\accou\.julia\dev\DiffEqBase\src\DiffEqBase.jl:1
  [5] top-level scope
    @ C:\Users\accou\.julia\dev\DiffEqBase\src\DiffEqBase.jl:123
  [6] include
    @ .\Base.jl:418 [inlined]
  [7] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::String)
    @ Base .\loading.jl:1318
  [8] top-level scope
    @ none:1
  [9] eval
    @ .\boot.jl:373 [inlined]
 [10] eval(x::Expr)
    @ Base.MainInclude .\client.jl:453
 [11] top-level scope
    @ none:1
in expression starting at C:\Users\accou\.julia\dev\DiffEqBase\src\solve.jl:203
in expression starting at C:\Users\accou\.julia\dev\DiffEqBase\src\solve.jl:203
in expression starting at C:\Users\accou\.julia\dev\DiffEqBase\src\DiffEqBase.jl:1
```

which I guess is related to the use of keyword arguments? So for now I just am ignoring the spots where we use the checkkwargs function.

devmotion:
`@non_differentiable f(x)` is the correct syntax, without keyword arguments. It is no function annotation but has to be done separately from the function definition.

PR #746, opened 2022-04-26
Check and throw on incompatible initial conditions

ChrisRackauckas:
Fixes the issue discussed in https://discourse.julialang.org/t/blog-post-about-my-experiences-with-julia/79976/84. A better fix is for this to be addressed at the Julia level, i.e. https://github.com/JuliaLang/julia/issues/45086, though we might as well catch and throw before the solve to keep the stacktrace small.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -158,7 +158,7 @@ const NON_SOLVER_MESSAGE =
 """
 Error: The arguments to solve are incorrect.
 The second argument must be a solver choice, `solve(prob,alg)`
-where `alg` is is a `<: DEAlgorithm`, i.e. `Tsti5()`.
+where `alg` is a `<: DEAlgorithm`, i.e. `Tsti5()`.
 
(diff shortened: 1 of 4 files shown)


src/solve.jl
```diff
@@ -103,6 +103,72 @@ end
...
+
+function Base.showerror(io::IO, e::NoTspanError)
+  print(io, NO_TSPAN_MESSAGE)
+end
+
+const NON_SOLVER_MESSAGE = 
+"""
+Error: The arguments to solve are incorrect.
+The second argument must be a solver choice, `solve(prob,alg)`
+where `alg` is is a `<: DEAlgorithm`, i.e. `Tsti5()`.
```
isaacsas, line 161:
```suggestion
where `alg` is a `<: DEAlgorithm`, i.e. `Tsti5()`.
```

isaacsas:
Thus looks great for helping less experienced users.

PR #751, opened 2022-05-18
check if have empty kwargs

isaacsas:
Fixes https://github.com/SciML/DiffEqBase.jl/issues/750

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -274,7 +274,7 @@ function solve(prob::AbstractJumpProblem, args...; kwargs...)
 end
 
 function checkkwargs(kwargshandle; kwargs...)
-  if any(x -> x ∉ allowedkeywords, keys(kwargs))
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

isaacsas:
Probably would be good to make a bug fix release if/when this gets merged as this is likely a pretty confusing error to some users.

isaacsas:
Sure

ChrisRackauckas:
yeah it'll be a quick patch.

isaacsas:
Strangely I'm not able to get the error when I directly build a `f(du,u,p,t)` or `f(u,p,t)`. In this case I guess a kwarg gets passed?

isaacsas:
Ahh, I must have tested on the wrong version as it seems like the issue is actually from ModelingToolkit which is forwarding an invalid kwarg:
```
keys(kwargs) = (:check_length,)
```

I'll move this over to ModelingToolkit.

PR #766, opened 2022-06-08
Fast no broadcast

chriselrod:
Should hopefully improve latency by reducing the amount of compiled code. In theory, should also improve runtime performance by a negligible amount due to removing a branch and reducing code size.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.89.1"
+version = "6.89.2"
 
 [deps]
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Isn't that the default?

And OrdinaryDiffEq would have a lot more of these.

chriselrod:
> Isn't that the default?
> 
> And OrdinaryDiffEq would have a lot more of these.

It's the default in 2.0, but not 1.
I figured we may as well keep support for 1 and specify `broadcast=false`.

chriselrod:
OrdinaryDiffEq seems to already be setting `broadcast=false` everywhere, but may as well bump the compat to allow v0.2 support.

YingboMa (approved):

chriselrod:
OrdinaryDiffEq does not explicitly depend on FastBroadcast, so no need to change anything there:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/83c16d1864ad7a51f7111d72585475a4a8b3ed09/src/OrdinaryDiffEq.jl#L58

PR #769, opened 2022-06-17
Fix UndefVarError for nonzeros

david-pl:
`nonzeros` is not imported so the method currently throws an error.


ChrisRackauckas:
Needs a test.

david-pl:
@ChrisRackauckas turns out there already are tests, they were just not included in `runtests.jl`. Should be fixed now.

ChrisRackauckas (approved):

ChrisRackauckas:
Good find, great thanks!

PR #770, opened 2022-06-18
save_first should be an allowed kwarg

chriselrod:


ChrisRackauckas (approved):

devmotion:
Is that different from `save_start`? OrdinaryDiffEq, eg., currently only supports `save_start` but not `save_first` it seems.

ChrisRackauckas:
oh oops 😅 . Yeah this should be reverted.

PR #775, opened 2022-06-19
Document anyeltypedual, handle empty tuple case and higher order AD

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqBase.jl/issues/774

### DIFF
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -26,7 +26,8 @@ promote_dual(::Type{T},::Type{T2}) where {
 # Untyped dispatch: catch composite types, check all of their fields
 
 """
-anyeltypedual(x)
+    anyeltypedual(x)
+
 
 Searches through a type to see if any of its values are parameters. This is used to
 then promote other values to match the dual type. For example, if a user pass a parameter
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -1,9 +1,49 @@
+"""
+promote_dual(::Type{T},::Type{T2})
+
+Is like the number promotion system, but always prefers a dual number type above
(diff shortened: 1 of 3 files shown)


src/forwarddiff.jl
```diff
@@ -1,9 +1,49 @@
+"""
+promote_dual(::Type{T},::Type{T2})
```
ranocha, line 2:
```suggestion
    promote_dual(::Type{T},::Type{T2})

```
For correct highlighting as code

src/forwarddiff.jl
```diff
@@ -1,9 +1,49 @@
...
+             T2<:ForwardDiff.Dual{T4,V2,N2}} = T
+promote_dual(::Type{T},::Type{T2}) where {
+             T3,V<:ForwardDiff.Dual,V2<:ForwardDiff.Dual,N,
+             T<:ForwardDiff.Dual{T3,V,N},
+             T2<:ForwardDiff.Dual{T3,V2,N}} = ForwardDiff.Dual{T3,promote_dual(V,V2),N}
+
+# Untyped dispatch: catch composite types, check all of their fields
+
+"""
+anyeltypedual(x)
```
ranocha, line 28:
```suggestion
    anyeltypedual(x)

```

src/forwarddiff.jl
```diff
@@ -1,9 +1,49 @@
...
+
+Searches through a type to see if any of its values are parameters. This is used to
+then promote other values to match the dual type. For example, if a user pass a parameter
```
ranocha, line 31:
```suggestion
then promote other values to match the dual type. For example, if a user passes a parameter

```

ranocha (commented):
Thanks for the quick fix!

PR #778, opened 2022-06-19
Even more ForwardDiff dual detection robustness

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqBase.jl/issues/776 and fixes https://github.com/SciML/DiffEqBase.jl/issues/779

### DIFF
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -51,16 +51,23 @@ function anyeltypedual(x)
   if propertynames(x) === ()
     Any
   else
-    mapreduce(y->anyeltypedual(getproperty(x,y)),promote_dual,propertynames(x))
+    mapreduce(y-> !isdefined(x,y) ? Any : anyeltypedual(getproperty(x,y)),promote_dual,propertynames(x))
   end
 end
-anyeltypedual(::Type{T}) where T = T
+
+Base.@pure anyeltypedual(::Type{T}) where T = hasproperty(T,:parameters) ? mapreduce(anyeltypedual,promote_dual,T.parameters;init=Any) : T
+anyeltypedual(::Type{T}) where T<:Union{ForwardDiff.Dual} = T
+anyeltypedual(::Type{T}) where T<:Union{AbstractArray,Set,NTuple} = anyeltypedual(eltype(T))
 
 # Any in this context just means not Dual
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
@ranocha this covers every case I can think of with just falling back to not converting when it cannot find a Dual type, and that fallback is always inferred (since the isempty case does eltype so when inference is failing it will still return the same type and so it should just infer fine. It will lose inference if it can find a dual though as that then would not match the eltype in some cases). Let me know after this patch is out whether there's anything still noticeable downstream. The tests cover a ton of cases here and every mapreduce catches the isempty case so it should be at least safe and robust (there there are probably a few ways to sneak a dual in, but this should catch 99.99% of the cases where someone is auto-diffing).

I can make the opt-out as well, but as the opt-out would make code unsafe for cases like autodiff (so i.e. make someone have to think more when using say Turing.jl), I'd hope the opt-out is rarely ever used.

ranocha:
Thanks a lot! Sure, I will test it with Trixi.jl when it's merged and released. Maybe we're just extreme users - we pass our whole semidiscretization as parameters to the right-hand side, including meshes, buffers etc.

ChrisRackauckas:
Okay cool, I think I handled all of the cases. There's a bunch of tests including structs with undefs, vectors with undefs, sets with structs with undefs, etc. There are a few cases where nesting an undef deep inside of there makes it not catch that there's a ForwardDiff, but this will make it just fallback to the previous behavior (of not catching the ForwardDiff Dual in that case and thus fail to automatically promote), so it shouldn't be "bad" and should be robust to not dying for heavily nested parameter structs. I think the real test will be releasing this as a patch and having @ranocha and @charleskawczynski rerun tests.

Sorry about this. I've been to clamp down on the set of "manual" behaviors and throwing a lot better error messages recently. If anything else got clamped that is usable behavior, just let me know.

ranocha:
Sure, I will do that. Thanks for the quick fix! I will come back to you later when a new version is released.

charleskawczynski:
Looks like this fixed it for us, thanks!

ranocha:
Looks like this fixed it for us, too - or at least it's hidden by other [CI failures using Julia v1.8](https://github.com/trixi-framework/Trixi.jl/pull/1074)...

PR #788, opened 2022-07-01
Fix ForwardDiff Dual detection for circular references

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqBase.jl/issues/784

### DIFF
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -47,7 +47,7 @@ then promote other values to match the dual type. For example, if a user passes 
 
 which is a `Dual` and a `u0` which is a `Float64`, after the first time step, `f(u,p,t) = p*u`
 will change `u0` from `Float64` to `Dual`. Thus the state variable always needs to be converted
-to a dual number before the solve. Worse still, this needs to be done in the case of 
+to a dual number before the solve. Worse still, this needs to be done in the case of
 `f(du,u,p,t) = du[1] = p*u[1]`, and thus running `f` and taking the return value is not a valid
 way to calculate the required state type.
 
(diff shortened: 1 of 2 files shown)


src/forwarddiff.jl
```diff
@@ -91,44 +96,47 @@ Base.@pure function anyeltypedual(::Type{T}) where {T <: Union{NTuple}}
...
-                                                                                  AbstractArray{
-                                                                                                <:Number
-                                                                                                },
-                                                                                  Set{
-                                                                                      <:Number
-                                                                                      },
-                                                                                  NTuple{N,
-                                                                                         <:Number
-                                                                                         }}}
+function anyeltypedual(x::Union{Array{T}, AbstractArray{T}, Set{T}},
```
YingboMa, line 110: That's impressive formatting.

src/forwarddiff.jl
```diff
@@ -57,29 +57,34 @@ and then attempt to upconvert `u0` to match that dual-ness. Because this changes
...
 end
 
 # Opt out since these are using for preallocation, not differentiation
-anyeltypedual(x::Union{ForwardDiff.AbstractConfig, Module}) = Any
-anyeltypedual(x::Type{T}) where {T <: Union{ForwardDiff.AbstractConfig}} = Any
+anyeltypedual(x::Union{ForwardDiff.AbstractConfig, Module}, counter = 0) = Any
+anyeltypedual(x::Type{T}, counter = 0) where {T <: Union{ForwardDiff.AbstractConfig}} = Any
 
-Base.@pure function anyeltypedual(::Type{T}) where {T}
+Base.@pure function anyeltypedual(::Type{T}, counter = 0) where {T}
```
YingboMa, line 79: I think `@pure` annotation on generic function is not legal.

ChrisRackauckas: This should be safe unless someone mutates `T.parameters` though?

ChrisRackauckas: Do a follow up if you're interested. I'm not sure how to test that so I'll leave it to you.

PR #792, opened 2022-07-07
Typo in recommended algorithm Tsti5

DanielVandH:
Changes `Tsti5()` to `Tsit5()`. I also added a link to the available solvers to be consistent with `NO_DEFAULT_ALGORITHM_MESSAGE`, in case you think that can be useful.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -158,7 +158,7 @@ end
 const NON_SOLVER_MESSAGE = """
                            The arguments to solve are incorrect.
                            The second argument must be a solver choice, `solve(prob,alg)`
-                           where `alg` is a `<: DEAlgorithm`, e.g.. `Tsit5()`. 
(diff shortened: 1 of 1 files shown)


src/solve.jl
```diff
@@ -158,9 +158,12 @@ end
 const NON_SOLVER_MESSAGE = """
                            The arguments to solve are incorrect.
                            The second argument must be a solver choice, `solve(prob,alg)`
-                           where `alg` is a `<: DEAlgorithm`, i.e. `Tsti5()`.
+                           where `alg` is a `<: DEAlgorithm`, e.g.. `Tsit5()`.
```
ChrisRackauckas, line 161:
```suggestion
                           where `alg` is a `<: DEAlgorithm`, e.g. `Tsit5()`. 
```

ChrisRackauckas:
Thanks!

PR #803, opened 2022-08-17
diffeqmapreduce for faster anyeltypedual in some situations

chriselrod:
```julia
julia> @time using DiffEqBase, StaticArrays
  7.564427 seconds (10.75 M allocations: 849.099 MiB, 6.27% gc time, 25.67% compilation time: 54% of which was recompilation)

(p1) pkg> st -m DiffEqBase
Status `~/Documents/progwork/julia/env/p1/Manifest.toml`
  [2b5f629d] DiffEqBase v6.95.2

julia> @code_typed DiffEqBase.promote_u0(SA[1.0,2.0], Returns(((Ka = 1.0,CL = 1, Vc = 1.0,))), 1.2)
CodeInfo(
1 ─       goto #4 if not true
2 ─ %2  = %new(DiffEqBase.var"#85#86"{Returns{NamedTuple{(:Ka, :CL, :Vc), Tuple{Float64, Int64, Float64}}}, Int64}, p, 1)::DiffEqBase.var"#85#86"{Returns{NamedTuple{(:Ka, :CL, :Vc), Tuple{Float64, Int64, Float64}}}, Int64}
│   %3  = DiffEqBase.mapreduce::typeof(mapreduce)
│   %4  = DiffEqBase.promote_dual::typeof(DiffEqBase.promote_dual)
│   %5  = invoke Base.:(var"#mapreduce#263")($(QuoteNode(Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}()))::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, %3::typeof(mapreduce), %2::Function, %4::Function, (:value,)::Tuple{Symbol})::Any
│   %6  = ForwardDiff.Dual::Type{ForwardDiff.Dual}
│   %7  = (%5 <: %6)::Bool
└──       goto #4 if not %7
3 ─ %9  = Base.broadcasted(%5, u0)::Base.Broadcast.Broadcasted{Style, Nothing} where Style<:Union{Nothing, Base.Broadcast.BroadcastStyle}
│   %10 = Base.Broadcast.instantiate(%9)::Any
│   %11 = Base.Broadcast.copy(%10)::Any
└──       return %11
4 ┄       return u0
) => Any

julia> @benchmark DiffEqBase.promote_u0(SA[1.0,2.0], Returns(((Ka = 1.0,CL = 1, Vc = 1.0,))), 1.2)
BenchmarkTools.Trial: 10000 samples with 9 evaluations.
 Range (min … max):  3.272 μs …   4.913 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.310 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.351 μs ± 112.216 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▁▅███▇▇▅▃▂▁                 ▁▃▃▄▃▃▂▂          ▁             ▃
  █████████████▆▆▅▆▁▄▅▅▁▁▃▁▁▅▇█████████▇▇▇▇▇███████▇▇▇▇████▇▇ █
  3.27 μs      Histogram: log(frequency) by time      3.75 μs <

 Memory estimate: 336 bytes, allocs estimate: 11.
```
PR:
```julia
julia> @time using DiffEqBase, StaticArrays
[ Info: Precompiling DiffEqBase [2b5f629d-d688-5b77-993f-72d75c75574e]
 20.040235 seconds (10.61 M allocations: 845.558 MiB, 1.28% gc time, 9.50% compilation time: 67% of which was recompilation)

julia> @code_typed DiffEqBase.promote_u0(SA[1.0,2.0], Returns(((Ka = 1.0,CL = 1, Vc = 1.0,))), 1.2)
CodeInfo(
1 ─     return u0
) => SVector{2, Float64}

julia> @benchmark DiffEqBase.promote_u0(SA[1.0,2.0], Returns(((Ka = 1.0,CL = 1, Vc = 1.0,))), 1.2)
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  1.745 ns … 27.479 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     8.910 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.932 ns ±  0.352 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                                           █
  ▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▁▂▁▁▁▂▂▁▁▁█ ▂
  1.74 ns        Histogram: frequency by time        8.93 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.95.2"
+version = "6.95.3"
 
 [deps]
 ArrayInterfaceCore = "30b0a656-2188-435a-8636-2ec0e6a096e2"
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -36,8 +36,21 @@ function promote_dual(::Type{T},
     ForwardDiff.Dual{T3, promote_dual(V, V2), N}
 end
 
-# Untyped dispatch: catch composite types, check all of their fields
+# `reduce` and `map` are specialized on tuples to be unrolled (via recursion)
+# Therefore, they can be type stable even with heterogenous input types.
+# We also don't care about allocating any temporaries with them, as it should
+# all be unrolled and optimized away.
+# Being unrolled also means const prop can work for things like
+# `mapreduce(f, op, propertynames(x))`
+# where `f` may call `getproperty` and thus have return type dependent
+# on the particular symbol.
+# `mapreduce` hasn't received any such specialization.
+@inline diffeqmapreduce(f::F, op::OP, x::Tuple) where{F,OP} = reduce(op, map(f, x))
+@inline diffeqmapreduce(f::F, op::OP, x::NamedTuple) where{F,OP} = reduce(op, map(f, x))
+# For other container types, we probably just want to call `mapreduce`
+@inline diffeqmapreduce(f::F, op::OP, x) where{F,OP} = mapreduce(f, op, x)
 
+# Untyped dispatch: catch composite types, check all of their fields
 """
     anyeltypedual(x)
 
@@ -64,8 +77,7 @@ function anyeltypedual(x, counter = 0)
     if propertynames(x) === ()
         Any
     elseif counter < 100
-        counter += 1
-        mapreduce(y -> !isdefined(x, y) ? Any : anyeltypedual(getproperty(x, y), counter),
+        diffeqmapreduce(y -> !isdefined(x, y) ? Any : anyeltypedual(getproperty(x, y), counter+1),
                   promote_dual, propertynames(x))
     else
         Any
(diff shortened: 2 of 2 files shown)


ChrisRackauckas:
format

chriselrod:
There are inference test failures.
I'll use the better/more type stable `Static.reduce_tup`.

chriselrod:
On Core, it is now the new tests I added that are failing. They passed for me on 1.7.
I'll try 1.8 (is "1" now 1.8?) and master...

chriselrod:
The failures were because of code cov adding a bunch of `Expr(:code_coverage_effect)`: https://github.com/SciML/DiffEqBase.jl/runs/7889809012?check_suite_focus=true

chriselrod:
Is there any reason to think the failures are related?
```
 ERROR: LoadError: Package StochasticDiffEq errored during testing (received signal: KILL)
Stacktrace:
 [1] pkgerror(msg::String)
```
The failures certainly look bad.

ChrisRackauckas:
Sounds like something odd has happened with the v1.8 update that we now have to look into. Fantastic. The allocator looks different: the OrdinaryDiffEq.jl code makes it looks like `similar` or `zero` on some array type has changed its output type.

PR #806, opened 2022-08-22
@fastmath sqrt for ODE_DEFAULT_NORM

chriselrod:
This PR is purely motivated by aesthetics, not benchmarks.

When looking at assembly of an experimental perform_step! function, I thought it'd be prettier without the error check and call to complex domain error.

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -17,7 +17,9 @@ jobs:
         julia-version: [1,1.6]
         os: [ubuntu-latest]
         package:
-          - {user: SciML, repo: DelayDiffEq.jl, group: All}
+          - {user: SciML, repo: DelayDiffEq.jl, group: Interface}
+          - {user: SciML, repo: DelayDiffEq.jl, group: Integrators}
+          - {user: SciML, repo: DelayDiffEq.jl, group: Regression}
           - {user: SciML, repo: DiffEqFlux.jl, group: DiffEqFlux}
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
What does it do instead of error?

chriselrod:
> What does it do instead of error?

Return `NaN`.
But, are you expecting negative inputs?

chriselrod:
```julia
julia> @fastmath sqrt(-3.0)
NaN

julia> @fastmath sqrt(NaN)
NaN

julia> @fastmath sqrt(Inf)
Inf

julia> sqrt(Inf)
Inf

julia> sqrt(NaN)
NaN

julia> sqrt(-3.0)
ERROR: DomainError with -3.0:
sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
Stacktrace:
 [1] throw_complex_domainerror(f::Symbol, x::Float64)
   @ Base.Math ./math.jl:33
 [2] sqrt(x::Float64)
   @ Base.Math ./math.jl:675
 [3] top-level scope
   @ REPL[11]:1
```
It only makes a difference for negative inputs, which we shouldn't have because of `abs2` anyway.

ChrisRackauckas:
And NaN is probably better anyways, since we don't want to throw an error and instead catch, warn, and exit with a retcode anytime inside of the solver.

ChrisRackauckas (approved):

PR #812, opened 2022-08-31
Late wrapping in auto-specialized form

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/894


src/solve.jl
```diff
@@ -388,30 +388,49 @@ function init_call(_prob, args...; merge_callbacks = true, kwargshandle = Keywor
...
+        sensealg = prob.kwargs[:sensealg]
+    end
+
+    u0 = u0 !== nothing ? u0 : prob.u0
+    p = p !== nothing ? p : prob.p
+
+    init_up(prob, sensealg, u0, p, args...; kwargs...)
+end
+
+function init(prob::AbstractJumpProblem, args...; kwargs...)
```
isaacsas, line 403: @ChrisRackauckas by bypassing `init_call` this seems to have broken passing `kwargs` to `solve` via `JumpProblem`s.

ChrisRackauckas: What's throwing the error? It should just be a quick fix to bypass the bypass (like is done on the `solve` call)

PR #818, opened 2022-09-06
fix JumpProblem kwarg passing

isaacsas:
I think this should fix the issue I mentioned with kwargs not getting passed to solve from JumpProblem.


isaacsas:
If tests pass on JumpProcesses then this should be good.

ChrisRackauckas (approved):

PR #819, opened 2022-09-07
fix deprecated call

thevolatilebit:
Since https://github.com/JuliaLang/julia/pull/39448, `getproperty` has been deprecated in favor of `getfield`.

### DIFF
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -63,10 +63,12 @@ function (dec::DualEltypeChecker)(::Val{Y}) where {Y}
     anyeltypedual(getproperty(dec.x, Y), dec.counter)
 end
 
-# use `getfield` for `Pairs`, see https://github.com/JuliaLang/julia/pull/39448
-function (dec::DualEltypeChecker{<:Base.Pairs})(::Val{Y}) where {Y}
-    isdefined(dec.x, Y) || return Any
-    anyeltypedual(getfield(dec.x, Y), dec.counter)
+# use `getfield` on `Pairs`, see https://github.com/JuliaLang/julia/pull/39448
+if VERSION >= v"1.7"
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
That PR is only about `Pair`s, I don't see how it applies here?

thevolatilebit:
I keep getting
```julia
┌ Warning: use values(kwargs) and keys(kwargs) instead of kwargs.data and kwargs.itr
│   caller = DualEltypeChecker at forwarddiff.jl:62 [inlined]
└ @ Core ~/.julia/packages/DiffEqBase/iK5G7/src/forwarddiff.jl:62
```

The proposed change seems to fix this. [will find out more details]

ChrisRackauckas:
The solution would then be to create a separate dispatch on `function (dec::DualEltypeChecker{<:Pairs})(::Val{Y}) where {Y}` which uses getfield (and put a comment about that PR above it. Most of the time, you want to use `getproperty`: `getfield` is usually the incorrect thing to do. But it sounds like on Pairs you do want to use `getfield`.

thevolatilebit:
Thanks, updated the PR.

ChrisRackauckas:
Can you add a test with `@test_nowarn`?

ChrisRackauckas:
Test failure looks real

thevolatilebit:
Thanks, sure, the issue is that `Pairs` seems to be available only in v1.7+. Will update that.

ChrisRackauckas (approved):

ChrisRackauckas:
Beautiful. Don't worry about the formatter, I'll merge and run it. Thanks!

thevolatilebit:
> Beautiful. Don't worry about the formatter, I'll merge and run it. Thanks!

Great, many thanks!

PR #825, opened 2022-09-09
Solution Wrapping

xtalax:
Adds a solution wrapping stage for `ODEProblem` and `NonlinearProblem`, currently dispatching on `prob.f.sys.metadata`, This will be changed in `SciMLBase` in future to dispatch on `prob.problem_type` when `NonlinearProblem` has this field.

Important: must be merged AFTER https://github.com/SciML/SciMLBase.jl/pull/246


src/solve.jl
```diff
@@ -731,6 +731,9 @@ explanations of the timestepping algorithms, see the
   Defaults to true.
 * `merge_callbacks`: Toggles whether to merge `prob.callback` with the `solve` keyword
   argument `callback`. Defaults to `true`.
+* `nowrap`: Toggles whether to wrap the solution if `prob.problem_type` has a preferred
+  alternate wrapper type for the solution. Useful when speed, but not shape of solution
+  is important. Defaults to `false`.
```
ChrisRackauckas, line 736: It changes the return type, so it needs to be a value type for type stability.

src/solve.jl
```diff
@@ -733,7 +733,7 @@ explanations of the timestepping algorithms, see the
   argument `callback`. Defaults to `true`.
 * `nowrap`: Toggles whether to wrap the solution if `prob.problem_type` has a preferred
   alternate wrapper type for the solution. Useful when speed, but not shape of solution
-  is important. Defaults to `Val(false)`. `Val(true)` will not wrap the solution.
+  is important. Defaults to `Val(false)`. `Val(true)` will cancel wrapping the solution.
```
ChrisRackauckas, line 736: Flip this: `wrap`, defaults to `Val(true)`.

ChrisRackauckas:
Bump the SciMLBase lower bound?

PR #833, opened 2022-09-16
slightly improve `prob2dtmin`

oscardssmith:
companion to https://github.com/SciML/OrdinaryDiffEq.jl/pull/1762. This is less necessary, but a minor cleanup.


oscardssmith:
How did codecov go down? I added a test and didn't add lines?

src/utils.jl
```diff
@@ -18,18 +18,16 @@ end
...
+        return eps(t2 - t1)
+    else
+        return max(eps(typeof(t1)), eps(t1))
     end
-    return dtmin
 end
 prob2dtmin(tspan, ::Integer, ::Any) = 0
 # Multiplication is for putting the right units on the constant!
-prob2dtmin(tspan, onet, ::Any) = onet * 1 // 10^(10)
+prob2dtmin(tspan, onet, ::Any) = onet * 1 // Int64(2)^33 # roughly 10^10 but more likely to turn into a multiplication.
```
devmotion, line 30: The `Int64` seems problematic on 32bit systems?

oscardssmith: no, the lack of `Int64` was a problem on 32 bit systems that wasn't caught by tests because you don't have 32 bit CI. On a 32 bit system, this previously would be `1//1410065408` which is 10x higher than intended. (it's actually pretty lucky that it's only that bad).

devmotion: We test 32bit e.g. in DelayDiffEq since it is even more sensitive regarding regressions in eg OrdinaryDiffEq. At least previously the calculations were not infected with `Int64`. Such mix ups of Int32 and Int64 actually caused test failures and problems in DelayDiffEq some years ago.

oscardssmith: what do you want to do on 32 bit systems? Overflowing and picking a randomly different dtmin seems like a pretty bad state. The other option I can see is changing this to `onet * exp2(-33f0)` but that will cause promotion.

ChrisRackauckas: Preventing overflow is a good idea here. This seems like a reasonable fix.

ChrisRackauckas (approved):

src/utils.jl
devmotion: I'm curious how it will affect DelayDiffEq tests on 32bit, but I guess we'll see.

ChrisRackauckas:
CI believes so too.

PR #835, opened 2022-10-05
promote `tspan` with `dt`

ranocha:
This follows the suggestion of @devmotion in https://github.com/SciML/OrdinaryDiffEq.jl/issues/1776#issuecomment-1268395677

Closes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1776

### DIFF
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -219,7 +219,7 @@ end
 function promote_tspan(u0::AbstractArray{<:ForwardDiff.Dual}, p, tspan, prob, kwargs)
     if (haskey(kwargs, :callback) && has_continuous_callback(kwargs[:callback])) ||
        (haskey(prob.kwargs, :callback) && has_continuous_callback(prob.kwargs[:callback]))
-        return eltype(u0).(tspan)
+        return _promote_tspan(eltype(u0).(tspan), kwargs)
     else
         return _promote_tspan(tspan, kwargs)
     end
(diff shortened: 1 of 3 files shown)


src/forwarddiff.jl
```diff
@@ -213,15 +213,15 @@ end
...
 function promote_tspan(u0::AbstractArray{<:ForwardDiff.Dual}, p,
                        tspan::Tuple{<:ForwardDiff.Dual, <:ForwardDiff.Dual}, prob, kwargs)
-    return tspan
+    return _promote_tspan(tspan, kwargs)
 end
 
 function promote_tspan(u0::AbstractArray{<:ForwardDiff.Dual}, p, tspan, prob, kwargs)
     if (haskey(kwargs, :callback) && has_continuous_callback(kwargs[:callback])) ||
        (haskey(prob.kwargs, :callback) && has_continuous_callback(prob.kwargs[:callback]))
         return eltype(u0).(tspan)
```
ChrisRackauckas, line 222: handle it here too?

src/init.jl
```diff
@@ -1,5 +1,13 @@
 value(x) = x
-promote_tspan(u0, p, tspan, prob, kwargs) = tspan
+promote_tspan(u0, p, tspan, prob, kwargs) = _promote_tspan(tspan, kwargs)
+function _promote_tspan(tspan, kwargs)
+    if haskey(kwargs, :dt)
+        tspan1, tspan2, _ = promote(tspan..., kwargs[:dt])
```
devmotion, line 5: Probably not worth it since it operates on `Pairs`/`NamedTuple` anyway but one could avoid one lookup:
```suggestion
    if (dt = get(kwargs, :dt, nothing)) !== nothing
        tspan1, tspan2, _ = promote(tspan..., dt)
```

PR #839, opened 2022-11-09
Modify NLNewton type to allow for arbitrary relax types

bgroenks96:
Modifies the constructor for `NLNewton` to permit non-numeric relax types. This is used to specify line search algorithms from `LineSearches.jl`; e.g:

`NLNewton(relax=BackTracking())` for Newton-Armijo iteration.


YingboMa (changes requested):
format

bgroenks96:
Could you be more specific?

YingboMa:
Please follow https://github.com/SciML/ModelingToolkit.jl/blob/master/CONTRIBUTING.md

devmotion:
I ran JuliaFormatter, I thought that might be faster and speed up the review/merge cycle. Basically, I navigated to `.julia/dev/DiffEqBase` and ran
```julia
julia> ] add JuliaFormatter

julia> using JuliaFormatter

julia> format("."; verbose=true)
```
in a temporary environment. Alternatively, I sometimes use VSCode and just run the formatter there as it uses JuliaFormatter nowadays (called something like "Format document").

src/nlsolve/type.jl
```diff
@@ -73,11 +73,13 @@ end
 function NLNewton(; κ = 1 // 100, max_iter = 10, fast_convergence_cutoff = 1 // 5,
                   new_W_dt_cutoff = 1 // 5, always_new = false, check_div = true,
                   relax = 0 // 1)
-    if !(0 <= relax < 1)
+    function _check_relax(r::Number)
+        (0 <= relax < 1) ? r :
```
YingboMa, line 77: Defining functions like this creates an unnecessary closure.

ChrisRackauckas:
Alright, formatted and removed the closure. Also rebased. Let's close this one out.

devmotion:
It seems with the latest commits the purpose of the PR was lost (and I assume it won't work anymore since `_check_relax` is not defined)? Maybe an alternative would be to add a keyword argument `check_relax::Bool = true` to the constructor (name is up for discussion) and only check if `relax` is between 0 and 1 if it is set to `true`? It seems at least the motivation for the PR was that the check fails for some non-numeric types, so in that way it could be disabled manually. Alternatively, one could try to be a bit smart and use the default `check_relax::Bool = relax isa Number`.

ChrisRackauckas:
oh wait, I see what you mean. We might as well just start this one off fresh then.

PR #845, opened 2022-11-29
fix ODE_DEFAULT_NORM for empty vectors

oscardssmith:
fixes problem introduced in https://github.com/SciML/DiffEqBase.jl/pull/697.


ChrisRackauckas (approved):
why is the empty case important?

oscardssmith:
the empty case matters because automatic simplification of problems (e.g. MTK) can sometimes manage to remove all variables from simple problems.

ChrisRackauckas:
But it shouldn't even enter the solver in those cases, it should just bypass and build the solution?

oscardssmith:
you certainly would think that, wouldn't you.

ChrisRackauckas:
but... it does do that.

PR #847, opened 2022-11-30
Remove DiffEqBase nlsolve

ChrisRackauckas:


YingboMa (approved):

YingboMa (commented):
I think we still need `NLNewton, NLFunctional, NLAnderson` though. That's the alg types.

ChrisRackauckas:
They are in OrdinaryDiffEq. They just weren't being exported.

PR #848, opened 2022-12-02
support fixedsizediffcache

ChrisRackauckas:


ChrisRackauckas:
Odd failure, but also on master https://github.com/SciML/DiffEqBase.jl/pull/850

PR #852, opened 2022-12-05
fix solving problems with empty u0

oscardssmith:
previously `build_null_solution` was only called for `u0==nothing` but it also is needed for `empty` `u0`s.


ChrisRackauckas:
A test downstream needs to get updated: https://github.com/SciML/DiffEqBase.jl/actions/runs/3624564051/jobs/6111681280#step:6:596

oscardssmith:
After trying this, I don't think this is the right approach.

PR #854, opened 2022-12-15
make Distributions a weak dependency

oscardssmith:
On 1.9, this saves roughly .5 seconds. We also should do this for chainrules, but I'll leave that for another PR.

Before
```

julia> @time @eval using DiffEqBase
  1.914418 seconds (6.02 M allocations: 432.638 MiB, 6.05% gc time, 16.83% compilation time: <1% of which was recompilation)
```
After
```
julia> @time @eval using DiffEqBase
  1.516961 seconds (4.86 M allocations: 365.564 MiB, 5.58% gc time, 18.51% compilation time: <1% of which was recompilation)
```

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -1082,8 +1082,6 @@ function get_concrete_p(prob, kwargs)
 end
 
 handle_distribution_u0(_u0) = _u0
-handle_distribution_u0(_u0::Distributions.Sampleable) = rand(_u0)
-isdistribution(_u0::Distributions.Sampleable) = true
 
 eval_u0(u0::Function) = true
 eval_u0(u0) = false
--- a/Project.toml
+++ b/Project.toml
@@ -32,6 +32,12 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Tricks = "410a4b4d-49e4-4fbc-ab6d-cb71b17b3775"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
+[weakdeps]
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+
+[extensions]
+DistributionsExt = "Distributions"
+
 [compat]
 ArrayInterfaceCore = "0.1.26"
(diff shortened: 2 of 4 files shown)


src/solve.jl
```diff
@@ -1082,8 +1082,6 @@ function get_concrete_p(prob, kwargs)
 end
 
 handle_distribution_u0(_u0) = _u0
-handle_distribution_u0(_u0::Distributions.Sampleable) = rand(_u0)
-isdistribution(_u0::Distributions.Sampleable) = true
```
ChrisRackauckas, line 1086: this should be in the `else` of the `isdefined`?

oscardssmith: what do you mean?

ChrisRackauckas: good point

ChrisRackauckas: wait no, isn't this breaking for v1.7?

oscardssmith: no. As described in https://pkgdocs.julialang.org/dev/creating-packages/#Transition-from-normal-dependency-to-extension this PR will make Distributions act as a normal dependency in 1.8 or below (and will load these 2 lines)

ChrisRackauckas: but the `isdefined` will be false, so it will not include the file?

oscardssmith: It's `if !isdefined(Base, :get_extension)` (note the not)

ChrisRackauckas: Then the file is not included in v1.9? In what case is the file included and not included? For it to not break, it needs to be included in all versions, so I don't get how an `if` can even work.

oscardssmith: On 1.9 it gets included by the
```
[extensions]
DistributionsExt = "Distributions"
```
in the Project.toml which will load `DistributionsExt.jl` if (and only if) `Distributions` is in the user's `Manifest.toml`

anandijain: where's the docs to extension packages?

ChrisRackauckas:
We should do Measurements and stuff too

PR #855, opened 2022-12-19
allow Static.jl v0.7

ranocha:
The previous compat entry like this was changed in https://github.com/SciML/DiffEqBase.jl/pull/847. Because of this, we cannot use versions of OrdinaryDiffEq.jl newer than v6.33.3, see also https://github.com/JuliaSIMD/StrideArrays.jl/issues/65. If there is no strong reason to keep only Static.jl v0.8, I would like to be able to use the newest versions of both Trixi.jl and OrdinaryDiffEq.jl.


ranocha:
Thanks, @ChrisRackauckas !

PR #856, opened 2022-12-24
Change requires to weak deps

ChrisRackauckas:


ChrisRackauckas:
@oscardssmith does this look correct?

src/init.jl
```diff
@@ -15,100 +15,25 @@ function SciMLBase.tmap(args...)
 end
 
 function __init__()
-    @require Measurements="eff96d63-e80a-5855-80a2-b1b0885c5ab7" begin
-        function promote_u0(u0::AbstractArray{<:Measurements.Measurement},
-                            p::AbstractArray{<:Measurements.Measurement}, t0)
-            u0
+    if VERSION <= v"1.9"
```
oscardssmith, line 18: this should be `if !isdefined(Base, :get_extension)` (small detail but important as otherwise the code will be broken on 1.9 release candidates and betas which you don't want.

oscardssmith:
Looks great! What's the TTFX impact?

src/init.jl
```diff
@@ -15,100 +15,25 @@ function SciMLBase.tmap(args...)
...
-            sqrt(sum(x -> ODE_DEFAULT_NORM(x[1], x[2]),
-                     zip((value(x) for x in u), Iterators.repeated(t))) / length(u))
-        end
-        @inline function ODE_DEFAULT_NORM(u::Array{<:Measurements.Measurement, N},
-                                          t) where {N}
-            sqrt(sum(x -> ODE_DEFAULT_NORM(x[1], x[2]),
-                     zip((value(x) for x in u), Iterators.repeated(t))) / length(u))
+        @require MonteCarloMeasurements="0987c9cc-fe09-11e8-30f0-b96dd679fdca" begin
+            include("../ext/MonteCarloMeasurementsExt.jl")
+            using MonteCarloMeasurementsExt
```
oscardssmith, line 26: I don't think the `using` is needed here (but I'm not 100% sure)

src/init.jl
ChrisRackauckas: can you do the suggest? I'm not sure what you're saying

oscardssmith, line 18:
```suggestion
    if !isdefined(Base, :get_extension)
```

ChrisRackauckas:
I don't know if there will be a major one. Only if these packages are used I guess. But it's good to just keep up the maintenance

Project.toml
```diff
@@ -34,9 +34,17 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [weakdeps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+GeneralizedGenerated = "6b9d7cbe-bcb9-11e9-073f-15a7a543e2eb"
+Measurements = "eff96d63-e80a-5855-80a2-b1b0885c5ab7"
+MonteCarloMeasurements = "0987c9cc-fe09-11e8-30f0-b96dd679fdca"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
```
oscardssmith, line 40: you need to add these as test dependencies.

ChrisRackauckas: Why? That seems weird if they aren't in the tests.

Project.toml
```diff
@@ -88,5 +88,11 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
+
+GeneralizedGenerated = "6b9d7cbe-bcb9-11e9-073f-15a7a543e2eb"
```
oscardssmith, line 92: these lines aren't needed. You need to give the package names in the test target but you don't need to add the pkg to extras.

Project.toml
```diff
@@ -67,14 +75,24 @@ julia = "1.6"
...
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
+
+
+GeneralizedGenerated = "6b9d7cbe-bcb9-11e9-073f-15a7a543e2eb"
+Measurements = "eff96d63-e80a-5855-80a2-b1b0885c5ab7"
+MonteCarloMeasurements = "0987c9cc-fe09-11e8-30f0-b96dd679fdca"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
```
ChrisRackauckas, line 95:
```suggestion
```

ext/GeneralizedGeneratedExt.jl
```diff
@@ -0,0 +1,9 @@
+module GeneralizedGeneratedExt
+
+using GeneralizedGenerated, DiffEqBase
```
oscardssmith, line 3: The reason this is failing is that for requres compatability you need
```suggestion
using DiffEqBase
isdefined(Base, :get_extension) ? (using GeneralizedGenerated) : (using ..GeneralizedGenerated)
```
as described in https://pkgdocs.julialang.org/dev/creating-packages/#Requires.jl (because if `GeneralizedGenerated` is loaded by Requires then it will be in the parent rather than the extension.

oscardssmith: (the same applies to all of these)

ext/MeasurementsExt.jl
```diff
@@ -0,0 +1,32 @@
+module MeasurementsExt
+
+using Measurements, DiffEqBase
```
ChrisRackauckas, line 3:
```suggestion
using DiffEqBase
isdefined(Base, :get_extension) ? (using Measurements) : (using ..Measurements)
```

ext/MonteCarloMeasurementsExt.jl
```diff
@@ -0,0 +1,45 @@
+module MonteCarloMeasurementsExt
+
+using MonteCarloMeasurements, DiffEqBase
```
ChrisRackauckas, line 3:
```suggestion
using DiffEqBase
isdefined(Base, :get_extension) ? (using MonteCarloMeasurements) : (using ..MonteCarloMeasurements)
```

ext/UnitfulExt.jl
```diff
@@ -0,0 +1,24 @@
+module UnitfulExt
+
+using Unitful, DiffEqBase
```
ChrisRackauckas, line 3:
```suggestion
using DiffEqBase
isdefined(Base, :get_extension) ? (using Unitful) : (using ..Unitful)
```

src/init.jl
```diff
@@ -15,100 +15,25 @@ function SciMLBase.tmap(args...)
...
 
 function __init__()
-    @require Measurements="eff96d63-e80a-5855-80a2-b1b0885c5ab7" begin
-        function promote_u0(u0::AbstractArray{<:Measurements.Measurement},
-                            p::AbstractArray{<:Measurements.Measurement}, t0)
-            u0
+    if !isdefined(Base, :get_extension)
+        @require Measurements="eff96d63-e80a-5855-80a2-b1b0885c5ab7" begin
+            include("../ext/MeasurementsExt.jl")
+            using MeasurementsExt
```
oscardssmith, line 21: I think this `using` is also wrong.

ChrisRackauckas: I sure wish this was documenter somewhere... anywhere...

src/DiffEqBase.jl
```diff
@@ -150,5 +150,9 @@ export KeywordArgError, KeywordArgWarn, KeywordArgSilent
 
 if !isdefined(Base, :get_extension)
     include("../ext/DistributionsExt.jl")
+    include("../ext/MeasurementsExt.jl")
```
oscardssmith, line 153: these shouldn't be here. These will load unconditionally on `Version<=1.9` which will break since you only want them to load if someone else has loaded the weakdep. The reason `Distributions` is different is because that was being transferred from a `dep` to a `weakdep` instead of a `@requires` to a `weakdep`.

src/init.jl
oscardssmith: it kindof is. https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions) has all the steps that need to be taken.

ext/UnitfulExt.jl
```diff
@@ -0,0 +1,25 @@
+module UnitfulExt
+
+using DiffEqBase
+isdefined(Base, :get_extension) ? (using Unitful) : (using ..Unitful)
+
+# Support adaptive errors should be errorless for exponentiation
+DiffEqBase.value(x::Type{Unitful.AbstractQuantity{T, D, U}}) where {T, D, U} = T
```
oscardssmith, line 7: I think you might need to import diffeqbase. value since ci says they like 18 doesn't know what value is.

oscardssmith:
this looks like it's getting really close!

ext/UnitfulExt.jl
ChrisRackauckas: That's done on line 3?

oscardssmith: is `value` exported?

ChrisRackauckas: no, but it's directly namespacing into DiffEqBase.

oscardssmith: but https://github.com/SciML/DiffEqBase.jl/actions/runs/3919228340/jobs/6700098045 is showing
```
inplace = false: Error During Test at /home/runner/work/DiffEqBase.jl/DiffEqBase.jl/downstream/test/interface/units.jl:16
  Got exception outside of a @test
  UndefVarError: value not defined
  Stacktrace:
    [1] ODE_DEFAULT_NORM
      @ ~/work/DiffEqBase.jl/DiffEqBase.jl/ext/UnitfulExt.jl:20 [inlined]
  ```

Comment (ChrisRackauckas) [reply ext/UnitfulExt.jl:7]: I know what it's saying, but that doesn't mean that normal semantics of Julia have changed. If you `using X`, you can use unexported values via `X.value`. That's what it's doing here, for a value defined at https://github.com/SciML/DiffEqBase.jl/blob/weakdeps/src/init.jl#L1 . Unless extension packages have different rules.

File: ext/UnitfulExt.jl [hunk 13]
```diff
@@ -0,0 +1,25 @@
...
+DiffEqBase.value(x::Unitful.AbstractQuantity) = x.val
+@inline function DiffEqBase.ODE_DEFAULT_NORM(u::AbstractArray{<:Unitful.AbstractQuantity, N
+                                                              },
+                                             t) where {N}
+    sqrt(sum(x -> DiffEqBase.ODE_DEFAULT_NORM(x[1], x[2]),
+             zip((value(x) for x in u), Iterators.repeated(t))) / length(u))
```
ChrisRackauckas, line 13: oh, it's due to right here

ext/UnitfulExt.jl
```diff
@@ -0,0 +1,25 @@
...
+end
+@inline function DiffEqBase.ODE_DEFAULT_NORM(u::Array{<:Unitful.AbstractQuantity, N},
+                                             t) where {N}
+    sqrt(sum(x -> DiffEqBase.ODE_DEFAULT_NORM(x[1], x[2]),
+             zip((value(x) for x in u), Iterators.repeated(t))) / length(u))
```
ChrisRackauckas, line 18:
```suggestion
             zip((DiffEqBase.value(x) for x in u), Iterators.repeated(t))) / length(u))
end
@inline function DiffEqBase.ODE_DEFAULT_NORM(u::Array{<:Unitful.AbstractQuantity, N},
                                             t) where {N}
    sqrt(sum(x -> DiffEqBase.ODE_DEFAULT_NORM(x[1], x[2]),
             zip((DiffEEqBase.value(x) for x in u), Iterators.repeated(t))) / length(u))
```

ext/UnitfulExt.jl
```diff
@@ -0,0 +1,25 @@
...
+                                             t) where {N}
+    sqrt(sum(x -> DiffEqBase.ODE_DEFAULT_NORM(x[1], x[2]),
+             zip((DiffEqBase.value(x) for x in u), Iterators.repeated(t))) / length(u))
+end
+@inline function DiffEqBase.ODE_DEFAULT_NORM(u::Array{<:Unitful.AbstractQuantity, N},
+                                             t) where {N}
+    sqrt(sum(x -> DiffEqBase.ODE_DEFAULT_NORM(x[1], x[2]),
+             zip((DiffEEqBase.value(x) for x in u), Iterators.repeated(t))) / length(u))
+end
+@inline DiffEqBase.ODE_DEFAULT_NORM(u::Unitful.AbstractQuantity, t) = abs(value(u))
```
ChrisRackauckas, line 20:
```suggestion
@inline DiffEqBase.ODE_DEFAULT_NORM(u::Unitful.AbstractQuantity, t) = abs(DiffEqBase.value(u))
```

PR #857, opened 2022-12-27
allow initing and solving with u0=nothing

oscardssmith:
Take 2 of https://github.com/SciML/OrdinaryDiffEq.jl/pull/1818 This time it works with external solvers. @ChrisRackauckas how should I add a test for this? I can't `init` a problem within DiffEqBase since none of the algs are defined.

### DIFF
--- a/test/downstream/null_de.jl
+++ b/test/downstream/null_de.jl
@@ -20,8 +20,11 @@ for kwargs in [
 ]
     sol = solve(prob, kwargs...)
     init_sol = solve!(init(prob, kwargs...))
+    step_sol = step!(init(prob, kwargs...), prob.tspan[end]-prob.tspan[begin])
     @test sol.u == init_sol.u
     @test sol.t == init_sol.t
+    @test sol.u == step_sol.u
+    @test sol.t == step_sol.t
 end
 
 @variables t x y
--- a/test/downstream/null_de.jl
+++ b/test/downstream/null_de.jl
@@ -22,11 +22,11 @@ for kwargs in [
     init_integ = init(prob, kwargs...)
     solve!(init_integ)
     step_integ = init(prob, kwargs...)
-    step_sol = step!(step_integ, prob.tspan[end] - prob.tspan[begin])
-    @test sol.u == init_integ.u
-    @test sol.t == init_integ.t
-    @test sol.u == step_integ.u
-    @test sol.t == step_integ.t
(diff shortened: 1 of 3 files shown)


oscardssmith:
hmm, the invalidation is the new `solve!` method I added (I think). I'm not sure what the best fix is.

ChrisRackauckas:
The null integrator should probably interpolate to the analytical solution?

test/downstream/null_de.jl
```diff
@@ -13,10 +13,16 @@ sol = solve(prob, Tsit5())
...
-sol = solve!(integ)
-@test sol[x] == [0.0, 0.0]
-@test sol[y] == [0.0, 0.0]
+for kwargs in [
+    Dict(:saveat => 0:0.1:1),
+    Dict(:save_start => false),
+    Dict(:save_end => false),
+]
+    sol = solve(prob, kwargs...)
+    init_sol = solve!(init(prob, kwargs...))
```
ChrisRackauckas, line 22: and test the iteration?

oscardssmith:
I think this is finally done. In the process I discovered a bug in `build_null_solution` where
```
f(u,p,t) = -u
prob = ODEProblem(nothing, (0,1.0))
solve(prob, tspan=0.1)
```
didn't expand the `0.1` to `0.0:0.1:1.0`

ChrisRackauckas:
`solve(prob, tspan=0.1)` should be a warning about an incorrect kwarg?

ChrisRackauckas:
The test failures look real. I don't think the if statement is only catching the right cases.

ChrisRackauckas:
The test case looks fine if downstream passes.

ChrisRackauckas:
https://github.com/SciML/DiffEqBase.jl/actions/runs/3875552613/jobs/6608254597#step:6:618 way too much is caught in this branch.

oscardssmith:
@ChrisRackauckas I think this is finally ready to go.

ChrisRackauckas:
Except tests are failing.

oscardssmith:
The failing test is from sundials, and happens when constructing `JacVec(Lotka, ones(2))` which never inits or solves anything so it's hard to see how it could possibly related. The failure looks like it's `static_hasmethod` related.

ChrisRackauckas (approved):

oscardssmith:
thanks!

PR #859, opened 2022-12-30
Add alg_order method

charleskawczynski:
This PR copies the `alg_order` method definition from OrdinaryDiffEq.jl, so that it can be extended from DiffEqBase. cf: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1825.

I realized when making this PR, that we could move this all the way down to SciMLBase if that's preferred. I'm not super familiar with what sort of things are being separated

### DIFF
--- a/src/DiffEqBase.jl
+++ b/src/DiffEqBase.jl
@@ -151,4 +151,9 @@ export KeywordArgError, KeywordArgWarn, KeywordArgSilent
 if !isdefined(Base, :get_extension)
     include("../ext/DistributionsExt.jl")
 end
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Yeah, I guess it would fit in better with https://github.com/SciML/SciMLBase.jl/blob/master/src/alg_traits.jl, though it really only has a good meaning for differential equation algorithms.

ChrisRackauckas:
was this ever opened in SciMLBase?

charleskawczynski:
> was this ever opened in SciMLBase?

No, not yet. Sorry, been busy lately (just had twins! 👶🏻👶🏻)

PR #864, opened 2023-01-16
Fix ODE_DEFAULT_NORM for Static Arrays

utkarsh530:
Using Static Arrays in `ODE_DEFAULT_NORM` wasn't hitting the correct dispatch:

```
using StaticArrays, DiffEqBase, Cthulhu
tmp = @SArray rand(10)
@descend DiffEqBase.ODE_DEFAULT_NORM(tmp, 1.0)
```
```
ODE_DEFAULT_NORM(u::AbstractArray, t) in DiffEqBase at /home/gridsan/utkarsh/.julia/dev/DiffEqBase/src/common_defaults.jl:37
   ∘ ─ %0 = invoke ODE_DEFAULT_NORM(::SVector{100, Float64},::Float64)::Float64
38 1 ─ %1 = invoke Base.var"#mapreduce##kw"()($(QuoteNode((init = 0.0,)))::NamedTuple{(:init,), Tuple{Float64}}, DiffEqBase.mapreduce::typeof(mapreduce), DiffEqBase.UNITLESS_ABS2::Function, DiffEqBase.abs2_and_sum::Function, u::SVector{100, Float64})::Float64
```

Apparently, this caused problems with using the method in https://github.com/SciML/DiffEqGPU.jl

```
using DiffEqBase, CUDA, StaticArrays
function test(u)
  @cushow DiffEqBase.ODE_DEFAULT_NORM(u, 1.0f0)
  return
end

tmp = @SArray rand(10)
@cuda test(tmp) # works

tmp = @SArray rand(100)
@cuda test(tmp) #fails, causes some dynamic fn invocation
```

### DIFF
--- a/ext/GeneralizedGeneratedExt.jl
+++ b/ext/GeneralizedGeneratedExt.jl
@@ -1,7 +1,8 @@
 module GeneralizedGeneratedExt
 
 using DiffEqBase
-isdefined(Base, :get_extension) ? (using GeneralizedGenerated) : (using ..GeneralizedGenerated)
+isdefined(Base, :get_extension) ? (using GeneralizedGenerated) :
+(using ..GeneralizedGenerated)
 
 function SciMLBase.numargs(::GeneralizedGenerated.RuntimeFn{Args}) where {Args}
     GeneralizedGenerated.from_type(Args) |> length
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
Add a test for this?

PR #874, opened 2023-02-10
Add `stability_size`, `fsal`, and dense output fields

YingboMa:


ChrisRackauckas (approved):

src/tableaus.jl
```diff
@@ -3,20 +3,26 @@ $(TYPEDEF)
 
 Holds a tableau which defines an explicit Runge-Kutta method.
 """
-mutable struct ExplicitRKTableau{MType <: AbstractMatrix, VType <: AbstractVector, fsal} <:
+mutable struct ExplicitRKTableau{MType <: AbstractMatrix, VType <: AbstractVector, S} <:
```
YingboMa, line 6: This is unfortunately breaking, but maybe it's fine?

ChrisRackauckas: yeah it's honestly not used right now, so let's just go for it. Line up the OrdinaryDiffEq PR and we merge one right after the other.

PR #886, opened 2023-03-01
Replace destats with stats

avik-pal:


oscardssmith:
this change is at least somewhat breaking.

src/destats.jl
```diff
@@ -24,7 +24,7 @@ Statistics from the differential equation solver about the solution process.
 - maxeig: Maximum eigenvalue over the solution. This is only computed if the
   method is an auto-switching algorithm.
 """
-mutable struct DEStats
+mutable struct Stats
```
ChrisRackauckas, line 27: missing a deprecation `const`

avik-pal: Do we keep the names for internal structs the same https://github.com/SciML/DelayDiffEq.jl/pull/258#discussion_r1122316765? Or change them

avik-pal:
Only thing that fails is ForwardDiff Dual detection with https://github.com/SciML/SciMLBase.jl/pull/391

Project.toml
```diff
@@ -73,7 +73,7 @@ PreallocationTools = "0.4"
 RecursiveArrayTools = "2"
 Reexport = "1.0"
 Requires = "1.0"
-SciMLBase = "1.84"
+SciMLBase = "1.89"
```
ChrisRackauckas, line 76:
```suggestion
SciMLBase = "1.90"
```

PR #902, opened 2023-04-04
minor cleanup

oscardssmith:
`NonlinearProblem <: DEProblem` so the `Union` was redundant. As a side note, it probably would be a good idea to rename `DEProblem` to `SCIMLProblem` as some of the problems don't really have anything to do with differential equations.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -426,7 +426,7 @@ function init_call(_prob, args...; merge_callbacks = true, kwargshandle = Keywor
     end
 end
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
There already is SciMLProblem.

oscardssmith:
Are you sure? It looks like there's a `SciMLSolution` and `SciMLAlgorithm` but no `SciMLProblem`

ChrisRackauckas:
AbstractSciMLProblem.

PR #903, opened 2023-04-06
Use a mutable struct instead of Dict for Safe termination

avik-pal:


src/termination_conditions.jl
```diff
@@ -28,6 +28,22 @@ struct NLSolveSafeTerminationOptions{T1, T2, T3}
     min_max_factor::T3
 end
 
+TruncatedStacktraces.@truncate_stacktrace NLSolveSafeTerminationOptions
+
+Base.@kwdef mutable struct NLSolveSafeTerminationResult
+    best_objective_value::Number = Inf64
```
ChrisRackauckas, line 34: parameterize this?

PR #904, opened 2023-04-20
Don't return NaN norm when only partials contain NaNs

YingboMa:
It's somewhat common to have `NaN`s show up in the partials from the RHS function, we should not poison the internal norm because of that.

### DIFF
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -252,7 +252,11 @@ value(x::ForwardDiff.Dual) = value(ForwardDiff.value(x))
 @inline fastpow(x::ForwardDiff.Dual, y::ForwardDiff.Dual) = x^y
 
 sse(x::Number) = x^2
-sse(x::ForwardDiff.Dual) = sse(ForwardDiff.value(x)) + sum(sse, ForwardDiff.partials(x))
+function sse(x::ForwardDiff.Dual)
+    primal_norm = sse(ForwardDiff.value(x))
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Why would we not want to exit here? It would mean that the derivative calculation has gone awry so we'd want to exit?

YingboMa:
If the user passes `Dual` to compute the primal and partials simultaneously to compute the cost and the gradient in one evaluation, early existing would be incorrect.

ChrisRackauckas:
Why would early exiting be incorrect? The step either needs to reject or from this point on the gradient will all be NaN, so you might as well return early and tell them the gradient calculation has failed?

YingboMa:
It won't even evaluate the cost.

ChrisRackauckas:
That's true that it won't evaluate the cost, but if the user is using dual numbers then they almost certainly want the derivative, and if the derivative is already diverged then why would there be any reason to keep on computing? I would think we just stop there and say the derivative is not a number, as if you're doing an optimization and hit this then the optimizer will want to give up here anyways.

YingboMa:
Debugging this problem is pretty not obvious so maybe https://github.com/SciML/SciMLBase.jl/pull/435 can help a bit. Also, what if the minima is exactly when the derivative becomes undefined? I guess it's a rather rare case.

ChrisRackauckas:
I think we're not doing this? Comment if you think otherwise.

PR #913, opened 2023-06-09
Add a result with state stored for better handling of non linear solvers

avik-pal:
The initial implementation was based on the version in DEQs.jl. But DEQs.jl stored the solver history, which means returning just the best index was good enough. But to generically handle NonlinearSolve.jl, we will also have to store the state.

See https://github.com/avik-pal/BatchedNonlinearSolve.jl/blob/5ff8656e82348e587e2a7645b76c9cf6608e3712/src/raphson.jl#L36 where I am testing out some aspects before moving stuff to NonlinearSolve.

### DIFF
--- a/src/termination_conditions.jl
+++ b/src/termination_conditions.jl
@@ -47,27 +47,24 @@ Base.@kwdef struct NLSolveSafeTerminationResultWithState{uType, T}
     return_code::Base.RefValue{NLSolveSafeTerminationReturnCode.T} = Ref(NLSolveSafeTerminationReturnCode.Failure)
 end
 
-# Remove once support for AbstractDict has been dropped
 __get(n::NLSolveSafeTerminationResultWithState, ::Val{:u}) = n.u
-__get(n::NLSolveSafeTerminationResultWithState, ::Val{prop}) where {prop} = getproperty(n, prop)[]
+function __get(n::NLSolveSafeTerminationResultWithState, ::Val{prop}) where {prop}
+    getproperty(n, prop)[]
+end
 __get(n::NLSolveSafeTerminationResult, ::Val{prop}) where {prop} = getproperty(n, prop)
-__get(d::AbstractDict, ::Val{prop}) where {prop} = d[prop]
 
 function __setproperty!(n::NLSolveSafeTerminationResultWithState, ::Val{:u}, value)
     n.u .= value
 end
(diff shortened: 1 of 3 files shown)


src/termination_conditions.jl
```diff
@@ -36,11 +36,37 @@ Base.@kwdef mutable struct NLSolveSafeTerminationResult{T}
...
+#       this by making `u` untyped, but that would lead to type instability in the final
+#       return type.
+Base.@kwdef struct NLSolveSafeTerminationResultWithState{uType, T}
+    u::uType
+    best_objective_value::Base.RefValue{T} = Ref(Inf64)
+    best_objective_value_iteration::Base.RefValue{Int64} = Ref(0)
+    return_code::Base.RefValue{NLSolveSafeTerminationReturnCode.T} = Ref(NLSolveSafeTerminationReturnCode.Failure)
+end
+
 # Remove once support for AbstractDict has been dropped
```
ChrisRackauckas, line 50: how quickly can we phase that out? I think we can do that non-breakingly since it's internal API, it just needs steps so that the downstream packages don't just break.

avik-pal: The dict interface is already removed from downstream. I can remove it in this PR itself.

ChrisRackauckas: Yes that would be clean.

avik-pal: I have removed Dict Support

ChrisRackauckas:
https://github.com/SciML/DiffEqBase.jl/actions/runs/5336499162/jobs/9671275200?pr=913 looks like there's some downstream failures.

avik-pal:
https://github.com/SciML/SteadyStateDiffEq.jl/pull/61 should fix that

avik-pal:
@ChrisRackauckas I checked in the Manifest in https://github.com/SciML/SteadyStateDiffEq.jl/pull/61 and tests are passing

ChrisRackauckas:
@vpuri3 looks like you had a breaking change in one of the updates? https://github.com/SciML/DiffEqBase.jl/actions/runs/5337907412/jobs/9674567530?pr=913#step:6:875

vpuri3:
thanks for pointing that out. looks like i forgot to define a `cache_internals(::FunctionOp, ::AbstractArray)`. fix is https://github.com/SciML/SciMLOperators.jl/pull/207

PR #920, opened 2023-08-22
Switch find_first_continuous_callback to use a generated implementation.

meson800:
# Summary
As mentioned in SciML/DifferentialEquations.jl#971, the current recursive method for identifying the first continuous callback can cause the compiler to give up on type inference, especially when there are many callbacks. The fallback then allocates.

This switches this function to using a generated function (along with an inline function that takes splatted tuples). Because this generated function explicitly unrolls the tuple, there are no type inference problems.

# Testing
I added a test that allocates using the old implementation (about 19kb allocations!) but does not with the new function. If I remove one of the callbacks, the allocations no longer happen with the recursive function, so it might be dependent on e.g. the specific Julia version to determine the point where the compiler gives up on type inference.

It passes all of the other DiffEqBase.jl, but I was wondering if there was a better test suite I could do to ensure this is working as expected. It seems to work in my ODE code, but I'm a little wary since I touched some pretty core code.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -116,45 +116,42 @@ function get_condition(integrator::DEIntegrator, callback, abst)
     end
 end
 
-# Use Recursion to find the first callback for type-stability
-
-# Base Case: Only one callback
-function find_first_continuous_callback(integrator, callback::AbstractContinuousCallback)
-    (find_callback_time(integrator, callback, 1)..., 1, 1)
+# Use a generated function for type stability even when many callbacks are given
+@inline function find_first_continuous_callback(integrator,
+                                                callbacks::Vararg{
+                                                                  AbstractContinuousCallback,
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
> the allocations no longer happen with the recursive function, so it might be dependent on e.g. the specific Julia version to determine the point where the compiler gives up on type inference.

yes, 16 is the compiler cutoff for tuple recursion.

> It passes all of the other DiffEqBase.jl, but I was wondering if there was a better test suite I could do to ensure this is working as expected. It seems to work in my ODE code, but I'm a little wary since I touched some pretty core code.

The downstream OrdinaryDiffEq tests should be sufficient.

ChrisRackauckas:
Looks good!

PR #926, opened 2023-09-13
Add solve kwargs from BVP

avik-pal:


src/solve.jl
```diff
@@ -80,7 +80,12 @@ const allowedkeywords = (:dense,
     # This is for copying/deepcopying noise in StochasticDiffEq
     :alias_noise,
     # This is for SimpleNonlinearSolve handling for batched Nonlinear Solves
-    :batch)
+    :batch,
+    # `problem_type` is for BVPs. needed in Ensemble Solve
+    :problem_type,
```
ChrisRackauckas, line 85: Isn't that only defined in the problem as a trait?

avik-pal: Yeah I will look into why this is causing a problem downstream. Dropping this for now.

PR #927, opened 2023-09-16
Add Enzyme extension

wsmoses:
Requires https://github.com/SciML/SciMLBase.jl/pull/496 and https://github.com/EnzymeAD/Enzyme.jl/pull/1056

### DIFF
--- a/ext/DiffEqBaseEnzymeExt.jl
+++ b/ext/DiffEqBaseEnzymeExt.jl
@@ -7,7 +7,7 @@ isdefined(Base, :get_extension) ? (import Enzyme) : (import ..Enzyme)
 using ChainRulesCore
 using EnzymeCore
 
-function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{Duplicated{RT}}, prob, sensealg::Union{Const{Nothing}, Const{<:AbstractSensitivityAlgorithm}}, u0, p, args...; kwargs...) where RT
+function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{Duplicated{RT}}, prob, sensealg::Union{Const{Nothing}, Const{<:DiffEqBase.AbstractSensitivityAlgorithm}}, u0, p, args...; kwargs...) where RT
     @inline function copy_or_reuse(val, idx)
         if EnzymeCore.EnzymeRules.overwritten(config)[idx] && ismutable(val)
             return deepcopy(val)
@@ -31,7 +31,7 @@ function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.
     return EnzymeCore.EnzymeRules.AugmentedReturn{RT, RT, Any}(res[1], dres, tup::Any)
 end
 
-function EnzymeCore.EnzymeRules.reverse(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{<:Duplicated{RT}}, tape, prob, sensealg::Union{Const{Nothing}, Const{<:AbstractSensitivityAlgorithm}}, u0, p, args...; kwargs...) where RT
+function EnzymeCore.EnzymeRules.reverse(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{<:Duplicated{RT}}, tape, prob, sensealg::Union{Const{Nothing}, Const{<:DiffEqBase.AbstractSensitivityAlgorithm}}, u0, p, args...; kwargs...) where RT
 	dres, clos = tape
     dres = dres::RT
 	dargs = clos(dres)
--- a/ext/DiffEqBaseEnzymeExt.jl
+++ b/ext/DiffEqBaseEnzymeExt.jl
@@ -1,0 +1,53 @@
+module DiffEqBaseEnzymeExt
+
+using DiffEqBase
+import DiffEqBase: value
+isdefined(Base, :get_extension) ? (import Enzyme) : (import ..Enzyme)
+
+using ChainRulesCore
+using EnzymeCore
+
+function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{Duplicated{RT}}, prob, sensealg, u0, p, args...; kwargs...) where RT
+    @inline function copy_or_reuse(val, idx)
+        if EnzymeCore.EnzymeRules.overwritten(config)[idx] && ismutable(val)
+            return deepcopy(val)
+        else
+            return val
+        end
+    end
+
+    @inline function arg_copy(i)
+        copy_or_reuse(args[i].val, i+5)
+    end
+ 
+    res = DiffEqBase._solve_adjoint(copy_or_reuse(prob.val, 2), copy_or_reuse(sensealg.val, 3), copy_or_reuse(u0.val, 4), copy_or_reuse(p.val, 5), SciMLBase.ChainRulesOriginator(), ntuple(arg_copy, Val(length(args)))...;
+        kwargs...)
+
+    dres = deepcopy(res[1])::RT
+    for v in dres.u
+        v.= 0
+    end
+    tup = (dres, res[2])
+    return EnzymeCore.EnzymeRules.AugmentedReturn{RT, RT, Any}(res[1], dres, tup::Any)
+end
+
+function EnzymeCore.EnzymeRules.reverse(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{<:Duplicated{RT}}, tape, prob, sensealg, u0, p, args...; kwargs...) where RT
+	dres, clos = tape
+    dres = dres::RT
+	dargs = clos(dres)
+    for (darg, ptr) in zip(dargs, (func, prob, sensealg, u0, p, args...))
+        if ptr isa EnzymeCore.Const
+            continue
+        end
+        if darg == ChainRulesCore.NoTangent()
+            continue
+        end
+        ptr.dval .+= darg
+    end
+    for v in dres.u
+        v.= 0
+    end
+    return ntuple(_ -> nothing, Val(length(args)+4))
+end
+
+end
--- a/ext/DiffEqBaseEnzymeExt.jl
+++ b/ext/DiffEqBaseEnzymeExt.jl
@@ -31,7 +31,7 @@ function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.
     return EnzymeCore.EnzymeRules.AugmentedReturn{RT, RT, Any}(res[1], dres, tup::Any)
 end
 
-function EnzymeCore.EnzymeRules.reverse(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{<:Duplicated{RT}}, tape, prob, sensealg, u0, p, args...; kwargs...) where RT
+function EnzymeCore.EnzymeRules.reverse(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{<:Duplicated{RT}}, tape, prob, sensealg::Union{Const{Nothing}, Const{<:AbstractSensitivityAlgorithm}}, u0, p, args...; kwargs...) where RT
 	dres, clos = tape
     dres = dres::RT
 	dargs = clos(dres)
--- a/ext/DiffEqBaseEnzymeExt.jl
+++ b/ext/DiffEqBaseEnzymeExt.jl
@@ -7,7 +7,7 @@ isdefined(Base, :get_extension) ? (import Enzyme) : (import ..Enzyme)
 using ChainRulesCore
 using EnzymeCore
 
-function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{Duplicated{RT}}, prob, sensealg, u0, p, args...; kwargs...) where RT
+function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{Duplicated{RT}}, prob, sensealg::Union{Const{Nothing}, Const{<:AbstractSensitivityAlgorithm}}, u0, p, args...; kwargs...) where RT
     @inline function copy_or_reuse(val, idx)
         if EnzymeCore.EnzymeRules.overwritten(config)[idx] && ismutable(val)
             return deepcopy(val)
--- a/Project.toml
+++ b/Project.toml
@@ -10,6 +10,7 @@ DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 EnumX = "4e289a0a-7415-4d19-859d-a7e5c4648b56"
+EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
 FastBroadcast = "7034ab61-46d4-4ed7-9d0f-46aef9175898"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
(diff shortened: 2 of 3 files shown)


ext/DiffEqBaseEnzymeExt.jl
```diff
@@ -0,0 +1,46 @@
...
+
+    @inline function arg_copy(i)
+        copy_or_reuse(args[i].val, i+5)
+    end
+ 
+    res = DiffEqBase._solve_adjoint(copy_or_reuse(prob.val, 2), copy_or_reuse(sensealg.val, 3), copy_or_reuse(u0.val, 4), copy_or_reuse(p.val, 5), SciMLBase.ChainRulesOriginator(), ntuple(arg_copy, Val(length(args)))...;
+        kwargs...)
+
+    dres = Enzyme.Compiler.make_zero(RT, IdDict(), res[1])
+    return EnzymeCore.EnzymeRules.AugmentedReturn{RT, RT, Any}(res[1], dres, (dres, res[2])::Any)
```
ChrisRackauckas, line 27: Why `Any`?

ext/DiffEqBaseEnzymeExt.jl
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    @inline function arg_copy(i)
+        copy_or_reuse(args[i].val, i+5)
+    end
+ 
+    res = DiffEqBase._solve_adjoint(copy_or_reuse(prob.val, 2), copy_or_reuse(sensealg.val, 3), copy_or_reuse(u0.val, 4), copy_or_reuse(p.val, 5), SciMLBase.ChainRulesOriginator(), ntuple(arg_copy, Val(length(args)))...;
+        kwargs...)
+
+    dres = Enzyme.Compiler.make_zero(RT, IdDict(), res[1])
```
ChrisRackauckas, line 26: What's this command doing? That's not documented.

wsmoses: It's an internal function that is effectively like deepcopy, but zero'ing memory (to be accumulated into). The better solution here is to construct a zero'd res[1], if you have an API for that.

ext/DiffEqBaseEnzymeExt.jl
wsmoses: The return type needs to be a concrete type, without the cast, the return is `AugmentedReturn{RT, RT, A} where A`.

ext/DiffEqBaseEnzymeExt.jl
ChrisRackauckas: And is this why you needed https://github.com/SciML/SciMLBase.jl/pull/496 ? That's the only place where I could see mutating the prob. The solution though is only captured in a subset of the fields, specifically `sol.u` (kind of`sol.t`, maybe `sol.k` but we can't actually differentiate w.r.t. those so we just mark them as not differentiable so using that would error, but the other pieces it's definitely not differentiable w.r.t.).

ext/DiffEqBaseEnzymeExt.jl
ChrisRackauckas: why not use `typeof((dres, res[2]))`?

ext/DiffEqBaseEnzymeExt.jl
wsmoses: No the reason it is needed is that otherwise it is not legal to write a rule for the return type ODEProblem.

Alternatively to make_zero, if a good copy function was defined (with say a zero operation), that would suffice. However,
```
ERROR: LoadError: type ODESolution has no field sc
Stacktrace:
  [1] getproperty(x::ODESolution{…}, s::Symbol)
    @ SciMLBase ~/.julia/dev/SciMLBase/src/solutions/ode_solutions.jl:52 [inlined]
  [2] copy(VA::ODESolution{…})
    @ RecursiveArrayTools ~/.julia/packages/RecursiveArrayTools/X30HP/src/vector_of_array.jl:351
  [3] #augmented_primal#1
    @ ~/.julia/dev/DiffEqBase/ext/DiffEqBaseEnzymeExt.jl:27
  [4] augmented_primal
    @ ~/.julia/dev/DiffEqBase/ext/DiffEqBaseEnzymeExt.jl:10 [inlined]
    ```

**Comment by wsmoses** [comment]
Collectively the changes outlined here (and the test below) enables that SciML example to run successfully.

```julia
using Enzyme

Enzyme.API.typeWarning!(false)

# SciML Tools
using OrdinaryDiffEq,  SciMLSensitivity
using Optimization, OptimizationOptimisers

# Standard Libraries
using LinearAlgebra, Statistics

# External Libraries
using ComponentArrays, Lux, Zygote, Plots, StableRNGs
gr()

# Set a random seed for reproducible behaviour
const rng = StableRNG(1111)

begin

	function lotka!(du, u, p, t)
	    α, β, γ, δ = p
	    du[1] = α * u[1] - β * u[2] * u[1]
	    du[2] = γ * u[1] * u[2] - δ * u[2]
	end

	# Define the experimental parameter
	tspan = (0.0, 5.0)
	u0 = 5.0f0 * rand(rng, 2)
	p_ = [1.3, 0.9, 0.8, 1.8]
	prob = ODEProblem(lotka!, u0, tspan, p_)
	solution = solve(prob, Vern7(), abstol = 1e-12, reltol = 1e-12, saveat = 0.25)

	# Add noise in terms of the mean
	const X = Array(solution)
	const t = solution.t

	x̄ = mean(X, dims = 2)
	noise_magnitude = 5e-3
	const Xₙ = X .+ (noise_magnitude * x̄) .* randn(rng, eltype(X), size(X))

	plot(solution, alpha = 0.75, color = :black, label = ["True Data" nothing])
	scatter!(t, transpose(Xₙ), color = :red, label = ["Noisy Data" nothing])

	rbf(x) = exp.(-(x .^ 2))

	# Multilayer FeedForward
	U = Lux.Chain(Lux.Dense(2, 5, rbf), Lux.Dense(5, 5, rbf), Lux.Dense(5, 5, rbf),
	              Lux.Dense(5, 2))
	# Get the initial parameters and state variables of the model
	p, st = Lux.setup(rng, U)

	# Define the hybrid model
	function ude_dynamics!(du, u, p, t, p_true)
	    û = U(u, p, st)[1] # Network prediction
	    du[1] = p_true[1] * u[1] + û[1]
	    du[2] = -p_true[4] * u[2] + û[2]
	end

	# Closure with the known parameter
	nn_dynamics!(du, u, p, t) = ude_dynamics!(du, u, p, t, p_)
	# Define the problem

	const prob_nn = ODEProblem(nn_dynamics!, Xₙ[:, 1], tspan, p)

	function predict(θ, X, T)
	    _prob = remake(prob_nn, u0 = X, tspan = (T[1], T[end]), p = θ)
	    Array(solve(_prob, Vern7(), saveat = T,
	                abstol = 1e-6, reltol = 1e-6))
	end

	function loss(θ)
	    X̂ = predict(θ, Xₙ[:, 1], t)
		mean(abs2, Xₙ .- X̂)
	end
end

losses = Float64[]

callback = function (p, l)
    push!(losses, l)
    if length(losses) % 5 == 0
        println("Current loss after $(length(losses)) iterations: $(losses[end])")
    end
    return false
end

adtype = Optimization.AutoEnzyme()
optf = Optimization.OptimizationFunction((x, p) -> loss(x), adtype)
optprob = Optimization.OptimizationProblem(optf, ComponentVector{Float64}(p))

res1 = Optimization.solve(optprob, ADAM(), callback = callback, maxiters = 5000)
println("Training loss after $(length(losses)) iterations: $(losses[end])")
```

ext/DiffEqBaseEnzymeExt.jl
ChrisRackauckas: Could you just use:

```julia
function zerocopy(sol::ODESolution)
    ODESolution{T, N}(zero.(sol.u),
        nothing,
        nothing,
        zero(sol.t),
        nothing,
        nothing,
        nothing,
        nothing,
        false,
        0,
        nothing,
        nothing,
        sol.retcode)
end
```

? Basically just remove any of the things that are non output values and zero the values?

wsmoses: The type needs to be equivalent to the primal.

ChrisRackauckas: How about:

```julia
function zerocopy(sol::ODESolution)
    ODESolution{T, N}(zero.(sol.u),
        sol.u_analytic,
        sol.errors,
        zero(sol.t),
        sol.k,
        sol.prob,
        sol.alg,
        sol.interp,
        sol.dense,
        sol.tslocation,
        sol.stats,
        sol.alg_choice,
        retcode)
end
```

ChrisRackauckas:
Is https://github.com/SciML/SciMLBase.jl/pull/496 still necessary with the deepcopy?

wsmoses:
That is necessary to register a rule here at all, independent of the content of the rule.

ChrisRackauckas:
Why would that matter? ODEProblem isn't being differentiated?

wsmoses:
The problem is that the type is illegal to be used as an argument or return of a reverse-mode custom rule otherwise.

ChrisRackauckas:
Even as a const argument?

wsmoses:
Const (or duplicated in forward mode) are both fine.

ChrisRackauckas:
The purpose of the `solve_up` function is so that `u0` and `p` are removed from the `prob` since those are the two arguments being differentiated, and those are then passed as separate arguments. That is done to make `prob` a const w.r.t. differentiation of `u0` and `p`, since the functions then just use the two other pieces. This was required for ChainRules to work, so we could just make use of that.

ext/DiffEqBaseEnzymeExt.jl
```diff
@@ -0,0 +1,53 @@
+module DiffEqBaseEnzymeExt
+
+using DiffEqBase
+import DiffEqBase: value
+isdefined(Base, :get_extension) ? (import Enzyme) : (import ..Enzyme)
+
+using ChainRulesCore
+using EnzymeCore
+
+function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{Duplicated{RT}}, prob, sensealg, u0, p, args...; kwargs...) where RT
```
ChrisRackauckas, line 10:
```suggestion
function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{Duplicated{RT}}, prob::Const, sensealg::Const, u0, p, args...; kwargs...) where RT
```

is that sufficient?

wsmoses: Nope, the types there do not assert the types that are passed in. The results of Enzyme's activity analysis deduced that the prob object contained differentiable data, and therefore correctly requested a Duplicated first arg. 

Instead the argument needs to be marked inactive in ActivityAnalysis, similar to EnzymeRules.inactive, but a variant for only particular args.

wsmoses: @vchuravy you and someone else I forget had a use for an "this arg is inactive" that I made ActivityAnalysis parse some metadata for.

Did you have thoughts on what the API design for that should be?

ChrisRackauckas:
```julia
using Enzyme

Enzyme.API.typeWarning!(false)

# SciML Tools
using OrdinaryDiffEq,  SciMLSensitivity
using Optimization, OptimizationOptimisers

# Standard Libraries
using LinearAlgebra, Statistics

# External Libraries
using ComponentArrays, Lux, Zygote, StableRNGs, Plots

# Set a random seed for reproducible behaviour
const rng = StableRNG(1111)

rbf(x) = exp.(-(x .^ 2))

begin

	function lotka!(du, u, p, t)
	    α, β, γ, δ = p
	    du[1] = α * u[1] - β * u[2] * u[1]
	    du[2] = γ * u[1] * u[2] - δ * u[2]
	end

	# Define the experimental parameter
	tspan = (0.0, 5.0)
	u0 = 5.0f0 * rand(rng, 2)
	p_ = [1.3, 0.9, 0.8, 1.8]
	prob = ODEProblem(lotka!, u0, tspan, p_)
	solution = solve(prob, Vern7(), abstol = 1e-12, reltol = 1e-12, saveat = 0.25)

	# Add noise in terms of the mean
	const X = Array(solution)
	const t = solution.t

	x̄ = mean(X, dims = 2)
	noise_magnitude = 5e-3
	const Xₙ = X .+ (noise_magnitude * x̄) .* randn(rng, eltype(X), size(X))

	plot(solution, alpha = 0.75, color = :black, label = ["True Data" nothing])
	scatter!(t, transpose(Xₙ), color = :red, label = ["Noisy Data" nothing])

	# Multilayer FeedForward
	U = Lux.Chain(Lux.Dense(2, 5, rbf), Lux.Dense(5, 5, rbf), Lux.Dense(5, 5, rbf),
	              Lux.Dense(5, 2))
	# Get the initial parameters and state variables of the model
	p, st = Lux.setup(rng, U)

	# Define the hybrid model
	function ude_dynamics!(du, u, p, t, p_true)
	    û = U(u, p, st)[1] # Network prediction
	    du[1] = p_true[1] * u[1] + û[1]
	    du[2] = -p_true[4] * u[2] + û[2]
	end

	# Closure with the known parameter
	nn_dynamics!(du, u, p, t) = ude_dynamics!(du, u, p, t, p_)
	# Define the problem

	const prob_nn = ODEProblem(nn_dynamics!, Xₙ[:, 1], tspan, p)

	function predict(θ, X, T)
	    _prob = remake(prob_nn, u0 = X, tspan = (T[1], T[end]), p = θ)
	    Array(solve(_prob, Vern7(), saveat = T,
	                abstol = 1e-6, reltol = 1e-6))
	end

	function loss(θ)
	    X̂ = predict(θ, Xₙ[:, 1], t)
		mean(abs2, Xₙ .- X̂)
	end
end

losses = Float64[]

callback = function (p, l)
    push!(losses, l)
    if length(losses) % 5 == 0
        println("Current loss after $(length(losses)) iterations: $(losses[end])")
    end
    return false
end

adtype = Optimization.AutoEnzyme()
optf = Optimization.OptimizationFunction((x, p) -> loss(x), adtype)

# Multilayer FeedForward
U = Lux.Chain(Lux.Dense(2, 5, rbf), Lux.Dense(5, 5, rbf), Lux.Dense(5, 5, rbf),
                Lux.Dense(5, 2))
# Get the initial parameters and state variables of the model
p, st = Lux.setup(rng, U)

optprob = Optimization.OptimizationProblem(optf, ComponentVector{Float64}(p))

res1 = Optimization.solve(optprob, ADAM(), callback = callback, maxiters = 5000)
println("Training loss after $(length(losses)) iterations: $(losses[end])")

using Optim, OptimizationOptimJL

optprob2 = Optimization.OptimizationProblem(optf, res1.u)
res2 = Optimization.solve(optprob2, Optim.LBFGS(), callback = callback, maxiters = 1000)
println("Final training loss after $(length(losses)) iterations: $(losses[end])")

# Rename the best candidate
p_trained = res2.u

ts = first(solution.t):(mean(diff(solution.t)) / 2):last(solution.t)
X̂ = predict(p_trained, Xₙ[:, 1], ts)
# Trained on noisy data vs real solution
pl_trajectory = plot(ts, transpose(X̂), xlabel = "t", ylabel = "x(t), y(t)", color = :red,
                     label = ["UDE Approximation" nothing])
scatter!(solution.t, transpose(Xₙ), color = :black, label = ["Measurements" nothing])
savefig("Plot.png")
```

Plot

ext/DiffEqBaseEnzymeExt.jl
ChrisRackauckas, line 10:
```suggestion
function EnzymeCore.EnzymeRules.augmented_primal(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{Duplicated{RT}}, prob, sensealg::Union{Const{Nothing}, Const{<:AbstractSensitivityAlgorithm}}, u0, p, args...; kwargs...) where RT
```

this is important because the adjoint method specifically wants to pass through if `sensealg` is something different. This allows us to implement `sensealg=EnzymeAdjoint()` which works (or is supposed to work, will need to check) by differentiating the solver itself, rather than using the adjoint, which has different numerical and performance properties.

wsmoses: that's fine, just also make sure to add it to the reverse as well!

ext/DiffEqBaseEnzymeExt.jl
```diff
@@ -0,0 +1,53 @@
...
+
+    dres = deepcopy(res[1])::RT
+    for v in dres.u
+        v.= 0
+    end
+    tup = (dres, res[2])
+    return EnzymeCore.EnzymeRules.AugmentedReturn{RT, RT, Any}(res[1], dres, tup::Any)
+end
+
+function EnzymeCore.EnzymeRules.reverse(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{<:Duplicated{RT}}, tape, prob, sensealg, u0, p, args...; kwargs...) where RT
```
ChrisRackauckas, line 34:
```suggestion
function EnzymeCore.EnzymeRules.reverse(config::EnzymeCore.EnzymeRules.ConfigWidth{1}, func::Const{typeof(DiffEqBase.solve_up)}, ::Type{<:Duplicated{RT}}, tape, prob, sensealg::Union{Const{Nothing}, Const{<:AbstractSensitivityAlgorithm}}, u0, p, args...; kwargs...) where RT
```

ChrisRackauckas (approved):

PR #937, opened 2023-09-25
use correct `alg`

prbzrg:
The correct `alg` is decided in 
https://github.com/SciML/DiffEqBase.jl/blob/2dd4fafdf4857902a4401f11ae7e22934113a9f3/src/solve.jl#L1426-L1431
we should use it.

Fixes https://github.com/SciML/SciMLSensitivity.jl/issues/907

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -190,7 +190,6 @@ function Base.showerror(io::IO, e::NaNTspanError)
     println(io, TruncatedStacktraces.VERBOSE_MSG)
 end
 
-
 const NON_SOLVER_MESSAGE = """
                            The arguments to solve are incorrect.
                            The second argument must be a solver choice, `solve(prob,alg)`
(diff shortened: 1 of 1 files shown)


src/solve.jl
```diff
@@ -1450,9 +1449,9 @@ function _solve_adjoint(prob, sensealg, u0, p, originator, args...; merge_callba
     end
 
     if isempty(args)
-        _concrete_solve_adjoint(_prob, nothing, sensealg, u0, p, originator; kwargs...)
+        _concrete_solve_adjoint(_prob, alg, sensealg, u0, p, originator; kwargs...)
     else
-        _concrete_solve_adjoint(_prob, args[1], sensealg, u0, p, originator,
+        _concrete_solve_adjoint(_prob, alg, sensealg, u0, p, originator,
```
ChrisRackauckas, line 1454: `alg` isn't always defined here? in the two branches above `alg` needs to be defined

ChrisRackauckas (approved):

prbzrg:
Why are we checking `isempty(args)` when `else` part is the same?
https://github.com/SciML/DiffEqBase.jl/blob/2dd4fafdf4857902a4401f11ae7e22934113a9f3/src/solve.jl#L1468-L1474

ChrisRackauckas:
That should be splatting the rest of the args.

prbzrg:
But `get_concrete_problem` doesn't take any more arguments:
https://github.com/SciML/DiffEqBase.jl/blob/2dd4fafdf4857902a4401f11ae7e22934113a9f3/src/solve.jl#L1114

ChrisRackauckas:
Then we don't need that branch 😅

ChrisRackauckas (approved):

prbzrg:
CI failed, but it seems unrelated, and format check needs formatting for files that didn't change here.

ChrisRackauckas:
It is unrelated, but I want to wait for @ErikQQY 's SDE fixes before merging more.

ChrisRackauckas:
@ErikQQY see the failing remake tests here

PR #938, opened 2023-09-29
merge stats objects

pepijndevos:
This allows something like an EnsembleSolution to easily provide a combined stats object, like https://github.com/SciML/SciMLBase.jl/pull/512


ChrisRackauckas:
Can we just complete the move of stats and the change to DEStats?

PR #940, opened 2023-10-01
A few test fixes for SciMLBase v2.0

ChrisRackauckas:


ErikQQY:
I am working on the BVP remake test failings, I think it was introduced in here: https://github.com/SciML/SciMLBase.jl/pull/477

ChrisRackauckas:
Awesome, thanks for looking into it. Yes, BVProblem remake is pretty much busted from the v2 update so someone will need to fix that up. I was just trying to get tests green again so that downstream testing can continue as normal. If you can follow this up with a test fix there that would be great.

PR #942, opened 2023-10-04
Translate Particle problem to EnsembleProblem

baggepinnen:
This PR adds WIP functionality that automatically translates an `ODEPRoblem{<:Particles}` into an `EnsembleProblem`, solves the ensemble problem, and finally repackages the solution into an ODESolution with Particles.

This is an attempt at making MCM work with DAEs and other things that might be thrown off by the uncertain numbers.

The PR is WIP, I am quite unsure about how to create the final `ODESolution{Particles}`. The current state of the PR kind-of works in some cases but not in others. There might also be problems in the interaction between the plotting recipe for `Particles` and that for `ODESolution`, if I plot a single state-variable only, it works fine, but when plotting multiple `idxs` at the same time I get a garbage plot.

Any initial comments on what I might be doing wrong when I create the solution below is appreciated.

### DIFF
--- a/ext/DiffEqBaseMonteCarloMeasurementsExt.jl
+++ b/ext/DiffEqBaseMonteCarloMeasurementsExt.jl
@@ -51,4 +51,65 @@ end
     abs(value(u))
 end
 
+# To make DAE problems work, we translate the MCM problem into an ensemble problem,
+# solve it, and then repackage the solution into an ODESolution with Particles
+
+const ParticleODEProblem = ODEProblem{<:Any, <:Any, <:Any, <:MonteCarloMeasurements.SomeKindOfParticles}
+
+using MonteCarloMeasurements: nparticles, vecindex
+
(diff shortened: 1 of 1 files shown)


ext/DiffEqBaseMonteCarloMeasurementsExt.jl
```diff
@@ -51,4 +51,65 @@ end
...
+end
+
+function postprocess(esol)
+    soli = esol[1]
+    t = soli.t
+    nx = length(soli[1])
+
+    # [state_index][mc_index]
+    utraj = map(t) do t
+        data = reduce(hcat, OrdinaryDiffEq.EnsembleAnalysis.componentwise_vectors_timepoint(esol,t)) # nmc × nx
```
ChrisRackauckas, line 96:
```suggestion
        data = reduce(hcat, SciMLBase.EnsembleAnalysis.componentwise_vectors_timepoint(esol,t)) # nmc × nx
```

ChrisRackauckas:
In principle this looks fine, though it should probably go to SciMLBase.

PR #945, opened 2023-10-07
Simplify `:alg` extraction

prbzrg:


prbzrg:
There was a CI failing related to `DynamicSS`. Based on
https://github.com/SciML/SciMLBase.jl/blob/master/src/deprecated.jl
I replaced types, maybe this fix it.

prbzrg:
Why doesn't this test have a `alg`?
https://github.com/SciML/SciMLSensitivity.jl/blob/282c24ef282d50d431913bfac2a83b025d34ae76/test/steady_state.jl#L411

ChrisRackauckas:
It's just using the default.

prbzrg:
> It's just using the default.

how does it access to defaults when `DifferentialEquations` isn't imported?

https://github.com/SciML/SciMLSensitivity.jl/blob/master/test/steady_state.jl

ChrisRackauckas (approved):

avik-pal:
I fixed it in https://github.com/SciML/DiffEqBase.jl/commit/2f0a9a43a04f7a0594ea7152cbf1fc6efa21c0ca

avik-pal:
Are nonlinear solvers AbstractDEAlgorithm? https://github.com/SciML/DiffEqBase.jl/blob/dc0e5c1a05ea63e4627e424fa492c8f391fcded4/src/solve.jl#L1460

prbzrg:
```julia
julia> supertypes(DynamicSS)
(DynamicSS, SteadyStateDiffEq.SteadyStateDiffEqAlgorithm, SciMLBase.AbstractSteadyStateAlgorithm, SciMLBase.AbstractDEAlgorithm, SciMLBase.AbstractSciMLAlgorithm, Any)
```

avik-pal:
Yeah but here the dispatch is failing for the discrete solvers.

```julia
julia> supertypes(NewtonRaphson)
(NewtonRaphson, NonlinearSolve.AbstractNewtonAlgorithm, NonlinearSolve.AbstractNonlinearSolveAlgorithm, SciMLBase.AbstractNonlinearAlgorithm, SciMLBase.AbstractSciMLAlgorithm, Any)
```

Dispatch on `AbstractSciMLAlgorithm`?

prbzrg:
Is it intended to ignore types of `EnsembleAlgorithm` and use `nothing` in 
https://github.com/SciML/DiffEqBase.jl/blob/301dd29b06d88cd4bc3e265e704a3e3a99f8495e/src/solve.jl#L1050-L1051
?

ChrisRackauckas:
That looks to me like a bug . It would be good to try and find a code that hits that branch.

src/solve.jl
```diff
@@ -1047,10 +1015,11 @@ function solve_call(prob::SteadyStateProblem,
...
 
 function solve(prob::EnsembleProblem, args...; kwargs...)
-    if isempty(args) || length(args) == 1 && typeof(args[1]) <: EnsembleAlgorithm
-        __solve(prob, nothing, args...; kwargs...)
+    alg = extract_alg(args, kwargs, kwargs)
+    if length(args) > 1
+        __solve(prob, alg, Base.tail(args)...; kwargs...)
     else
-        __solve(prob, args...; kwargs...)
+        __solve(prob, alg; kwargs...)
```
ChrisRackauckas, line 1022: this is dropping the rest of the args?

prbzrg: No, there isn't any other `args`, the length check assures us.

ChrisRackauckas: oh I see

prbzrg:
Most of the errors were caused by simplifying the `alg` extractions, and some of them needed changes elsewhere. like adding `kwargs` to `EnsembleProblem` or having a default `isadaptive` for non-`AbstractDEAlgorithm`. Here, I just imitated previous behaviour but IMO, the needed tests should happen elsewhere. like in `SciMLBase`.

PR #948, opened 2023-10-09
deprecate stats

pepijndevos:
This is getting moved to SciMLBase. I'm not sure if this is the correct way to mark it deprecated. It seems `@deprecate_moved` seems close but makes the current binding error. `@deprecate_binding` actually requires the new symbol to exist.  Hum.

### DIFF
--- a/src/stats.jl
+++ b/src/stats.jl
@@ -39,6 +39,7 @@ mutable struct Stats
 end
 
 Base.@deprecate_binding DEStats Stats false
+Base.deprecate(@__MODULE__, :Stats)
 
 Stats(x::Int = -1) = Stats(x, x, x, x, x, x, x, x, x, x, 0.0)
 


ChrisRackauckas:
We can just forward it. Two steps:

1. Move Stats in DiffEqBase -> DEStats in SciMLBase
2. Add `const Stats = SciMLBase.DEStats` and bump the lower bound to the point where it exists in SciMLBase.

That is non-breaking and will let it adjust over time, since there's 10 other solvers that will also need to update.

pepijndevos:
I dismissed that idea because I thought it'd upend the dependency hierarchy but sure I'll do that after https://github.com/SciML/SciMLBase.jl/pull/512 lands

ChrisRackauckas:
Supersceded by https://github.com/SciML/DiffEqBase.jl/pull/949

PR #954, opened 2023-10-30
Rework the Termination Condition API to be type stable

avik-pal:
Leads to quite a bit of type instability in the cache of the NonlinearSolve Algorithms. I will prototype a version there as well to see if this fixes things!


avik-pal:
This should be good to go once the NonlinearSolve PR tests pass

PR #955, opened 2023-10-31
Add compats for StdLibs

avik-pal:


avik-pal:
https://github.com/JuliaRegistries/General/pull/94465

PR #958, opened 2023-11-03
`dualT4` should not be promoted with `T` since `T` might already be a dual

oscardssmith:
this is the second part of the fix for https://github.com/SciML/DifferentialEquations.jl/issues/994

### DIFF
--- a/src/norecompile.jl
+++ b/src/norecompile.jl
@@ -57,13 +57,11 @@ function wrapfun_iip(ff,
     dualT = dualgen(T)
     dualT1 = ArrayInterface.promote_eltype(T1, dualT)
     dualT2 = ArrayInterface.promote_eltype(T2, dualT)
-    dualT4 = dualgen(T4)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
https://github.com/SciML/DiffEqBase.jl/actions/runs/6749257536/job/18349252125?pr=958 this failure looks very real.

oscardssmith:
CI is required to confirm, but I think this might finally work. The problem we were running into previously is that `tgrad` actually has 2 different function signatures  depending on whether we are inside a `ForwardDiff` call that was differentiating wrt the initial conditions (since doing so adds an extra ForwardDiff tag to the time type.

ChrisRackauckas:
https://github.com/SciML/DiffEqBase.jl/actions/runs/6852222169/job/18630260564?pr=958#step:6:1416 that's a real test failure.

oscardssmith:
yeah this pr I believe only works once my ordinarydiffeq pr is merged

PR #963, opened 2023-11-14
add `_rate_prototype` function for getting the type of `du`

oscardssmith:
from the type of `u` and `t`
This is designed to simplify handling in `OrdinaryDiffEq` which reimpliments this logic several times (and gets the `unitful` version wrong).

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -62,4 +62,5 @@ function default_logger(logger)
     LoggingExtras.TeeLogger(logger1, logger2)
 end
 
-_rate_prototype(u, t::T, onet::T) where {T} = u / oneunit(t)
+# for the non-unitful case the correct type is just u
+_rate_prototype(u, t::T, onet::T) where {T} = u
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -61,3 +61,5 @@ function default_logger(logger)
 
     LoggingExtras.TeeLogger(logger1, logger2)
 end
+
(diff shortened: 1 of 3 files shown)


src/utils.jl
```diff
@@ -61,3 +61,6 @@ function default_logger(logger)
 
     LoggingExtras.TeeLogger(logger1, logger2)
 end
+
+# for the non-unitful case the correct type is just u
+_rate_prototype(u, t::T, onet::T) where {T} = u
```
ChrisRackauckas, line 66: what's the purpose of the `onet`?

ChrisRackauckas: oh I see, to see if `typeof(one(t)) === typeof(t)`

test/utils.jl
```diff
@@ -1,7 +1,8 @@
 using Test
 
 using DiffEqBase, ForwardDiff
-using DiffEqBase: prob2dtmin, timedepentdtmin
+using DiffEqBase: prob2dtmin, timedepentdtmin, _rate_prototype
+using Unitful
```
ChrisRackauckas, line 6:
```suggestion
using Unitful
using ForwardDiff: Dual, Tag

```

ChrisRackauckas (approved):

oscardssmith:
When we merge this it would be good to have a version bump so I can use this in https://github.com/SciML/OrdinaryDiffEq.jl/pull/2051.

PR #969, opened 2023-11-24
Make the termination conditions non-allocating

avik-pal:


ChrisRackauckas:
Test it?

avik-pal:
Yeah let me club all the changes needed for SimpleNonlinearSolve together

avik-pal:
This should be good to go, the SciMLSensitivity failure is due to the deprecated DynamicSS call which I have fixed in a PR there

ChrisRackauckas (approved):

PR #971, opened 2023-12-06
Change the default norm for NonlinearSolve to norm(_, 2)

avik-pal:
RMS-Norm causes problem with some of the algorithms, especially TR and Bad-Broyden Updates

We could use Inf Norm for size invariance but I haven't tested it out


avik-pal:
The NonlinearSolve CI failure is not a regression rather an improvement

PR #977, opened 2023-12-17
Update documentation of `save_start` and `save_end`

ArneBouillon:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1842

A proposal to clarify the complicated current behaviour of `save_start` and `save_end` in the documentation.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -776,10 +776,14 @@ section at the end of this page for some example usage.
   settings of `dense`, `saveat` and `save_everystep` and is used by some applications
   to manually turn off saving temporarily. Everyday use of the solvers should leave
   this unchanged. Defaults to `true`.
-* `save_start`: Denotes whether the initial condition should be included in
-  the solution type as the first timepoint. Defaults to `true`.
-* `save_end`: Denotes whether the final timepoint is forced to be saved,
-  regardless of the other saving settings. Defaults to `true`.
+* `save_start`: Denotes whether the initial condition should be included in the solution
+  type as the first timepoint. This setting overrides `saveat` when set to `false`.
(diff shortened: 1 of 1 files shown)


src/solve.jl
```diff
@@ -776,10 +776,14 @@ section at the end of this page for some example usage.
...
-  the solution type as the first timepoint. Defaults to `true`.
-* `save_end`: Denotes whether the final timepoint is forced to be saved,
-  regardless of the other saving settings. Defaults to `true`.
+* `save_start`: Denotes whether the initial condition should be included in the solution
+  type as the first timepoint. This setting overrides `saveat` when set to `false`.
+  Defaults to
+  `save_everystep || isempty(saveat) || saveat isa Number || prob.tspan[1] in saveat`.
+* `save_end`: Denotes whether the final condition should be included in the solution type
+  as the final timepoint. This setting is overridden by other saving settings when set to
+  `false`. Defaults to
```
ChrisRackauckas, line 785: That shouldn't be the case. `save_start` and `save_end` should be concrete and overriding in that sense. The solution here isn't to document a current bug.

ArneBouillon: That's on me then; I interpreted https://github.com/SciML/OrdinaryDiffEq.jl/issues/1842#issuecomment-1849911515 to mean that the current behaviour should be documented. I'll mark both PRs for this as WIP.

Should both arguments be handled as `save_start` currently is; that is, overriding `saveat` when explicitly given? If so, I imagine there are a lot of different solvers implementing the DiffEq.jl interface that will need to be changed (or, of course, that are currently buggy).

PR #979, opened 2023-12-19
Missed updating a few to inf norm

avik-pal:


ChrisRackauckas (approved):

avik-pal:
The nonlinearsolve failure is from the Klement tests which are unrelated to the change here.

PR #981, opened 2023-12-22
Fix type stability for duals

avik-pal:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2091

### DIFF
--- a/src/forwarddiff.jl
+++ b/src/forwarddiff.jl
@@ -262,16 +262,17 @@ totallength(x::Number) = 1
 function totallength(x::ForwardDiff.Dual)
     totallength(ForwardDiff.value(x)) + sum(totallength, ForwardDiff.partials(x))
 end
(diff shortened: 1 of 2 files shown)


src/forwarddiff.jl
```diff
@@ -262,15 +262,16 @@ totallength(x::Number) = 1
 function totallength(x::ForwardDiff.Dual)
     totallength(ForwardDiff.value(x)) + sum(totallength, ForwardDiff.partials(x))
 end
-totallength(x::AbstractArray) = sum(totallength, x)
+# sum breaks type inference
+totallength(x::AbstractArray) = mapreduce(totallength, +, x)
```
ChrisRackauckas, line 266: does it help if you provide the zero type?

avik-pal: yes that works

PR #985, opened 2023-12-27
Missing Splat for StaticArrays

avik-pal:


ChrisRackauckas:
I guess this will do it too. I'll change mine to this.

PR #998, opened 2024-02-06
Fix discrete callback evaluation modification

ChrisRackauckas:
Fixes the handling of https://github.com/SciML/DiffEqBase.jl/pull/997

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -603,7 +603,7 @@ end
         end
         integrator.u_modified = true
         callback.affect!(integrator)
-        if (integrator.sol.prob isa DAEProblem || !(integrator.f.mass_matrix isa UniformScaling)) && integrator.u_modified
+        if integrator.u_modified
             reeval_internals_due_to_modification!(integrator)
         end
         @inbounds if callback.save_positions[2]
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -603,8 +603,8 @@ end
         end
         integrator.u_modified = true
         callback.affect!(integrator)
-        if integrator.sol.prob isa DAEProblem || !(integrator.f.mass_matrix isa UniformScaling)
-            DiffEqBase.initialize_dae!(integrator)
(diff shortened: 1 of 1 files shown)


src/callbacks.jl
```diff
@@ -603,8 +603,8 @@ end
         end
         integrator.u_modified = true
         callback.affect!(integrator)
-        if integrator.sol.prob isa DAEProblem || !(integrator.f.mass_matrix isa UniformScaling)
-            DiffEqBase.initialize_dae!(integrator)
+        if (integrator.sol.prob isa DAEProblem || !(integrator.f.mass_matrix isa UniformScaling)) && integrator.u_modified
```
ChrisRackauckas, line 606:
```suggestion
        if integrator.u_modified
```

topolarity:
Does this also resolve https://github.com/SciML/OrdinaryDiffEq.jl/issues/2131?

oscardssmith:
it doesn't, but it really should. I've gone through this in the debugger and it is calling the initialize_dae at the right time, it's just that the call isn't changing anything. I think somehow the call isn't updating Sundials with the new `p`...

src/callbacks.jl
```diff
@@ -603,8 +603,8 @@ end
         end
         integrator.u_modified = true
         callback.affect!(integrator)
-        if integrator.sol.prob isa DAEProblem || !(integrator.f.mass_matrix isa UniformScaling)
-            DiffEqBase.initialize_dae!(integrator)
+        if integrator.u_modified
```
oscardssmith, line 606: isn't this check dead since `integrator.u_modified = true` is 2 lines above this?

ChrisRackauckas: No that's the point, you're supposed to set it in the callback if you want to override it.

oscardssmith: right. That makes sense. Is this ready to merge?

oscardssmith:
closing in favor of https://github.com/SciML/DiffEqBase.jl/pull/1007

PR #1008, opened 2024-02-15
Run JuliaFormatter.format()

LilithHafner:
Using JuliaFormatter v1.0.49

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -612,7 +612,7 @@ end
         integrator.u_modified = true
         callback.affect!(integrator)
         if integrator.u_modified
-            reeval_internals_due_to_modification!(integrator, #=continuous_modification=# false)
+            reeval_internals_due_to_modification!(integrator, false) #=continuous_modification=#
         end
         @inbounds if callback.save_positions[2]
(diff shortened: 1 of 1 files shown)


src/callbacks.jl
```diff
@@ -612,7 +612,7 @@ end
         integrator.u_modified = true
         callback.affect!(integrator)
         if integrator.u_modified
-            reeval_internals_due_to_modification!(integrator, #=continuous_modification=# false)
+            reeval_internals_due_to_modification!(integrator, false) #=continuous_modification=#
```
oscardssmith, line 615: this is incorrect. This comment was specifically to refer to the `false`

LilithHafner: I'll report this upstream. Until then, perhaps we should simply close this PR and leave format-check broken.

ChrisRackauckas: ehh we can just change that comment, we can report upstream but let's just get the green check mark done.

ChrisRackauckas, line 615:
```suggestion
            reeval_internals_due_to_modification!(integrator, false) # continuous_modification=false
```

PR #1030, opened 2024-05-13
Incorrect ordering of promotion

avik-pal:
If `p` is not extracted before, then it tries to promote `u0` to the tracked type.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.149.2"
+version = "6.149.3"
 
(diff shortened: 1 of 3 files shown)


avik-pal:
@ChrisRackauckas the test failure is real https://github.com/SciML/DiffEqBase.jl/actions/runs/9069483730/job/24919132493#step:6:1967. We need to figure out how to do the conversion in a device agnostic way but `convert(AbstractArray, ...)` doesnt seem to work

ChrisRackauckas:
oh oops. Use Adapt?

avik-pal:
does Adapt natively work with RAT? I don't see any code there for that.

We should probably overload `convert(AbstractArray, <RAT>)` instead of https://github.com/SciML/RecursiveArrayTools.jl/blob/39573324b1460a8cefeed7dbb2a85350713544d4/src/vector_of_array.jl#L130

PR #1035, opened 2024-05-16
Don't check for large values in the unstable check

oscardssmith:
Fixes regression for Float32 ODEs caused by https://github.com/SciML/DiffEqBase.jl/pull/1033. This check in general doesn't really do much, and I think we probably will switch it to just be `return false` eventually, but first we need to add something to `OrdinaryDiffEq.jl` that actually checks for divergence in a more useful way (e.g. tracking number of failed steps in a row).

## Checklist


oscardssmith:
We actually did this yesterday (oops)

PR #1040, opened 2024-05-22
remove tests that broke with the existance of a Default ODESolver

oscardssmith:


ChrisRackauckas:
Merge when tests pass

PR #1054, opened 2024-06-06
use Enzyme make_zero

m-bossart:
Uses Enzyme's `make_zero` util instead of manually zeroing out the derivative variable. 
Needed for SciML/NonlinearSolve.jl#439 to make the Enzyme rule compatible with `NonlinearProblem`

### DIFF
--- a/ext/DiffEqBaseEnzymeExt.jl
+++ b/ext/DiffEqBaseEnzymeExt.jl
@@ -28,10 +28,7 @@ function Enzyme.EnzymeRules.augmented_primal(config::Enzyme.EnzymeRules.ConfigWi
         SciMLBase.EnzymeOriginator(), ntuple(arg_copy, Val(length(args)))...;
         kwargs...)
 
-    dres = deepcopy(res[1])::RT
-    for v in dres.u
-        v .= 0
-    end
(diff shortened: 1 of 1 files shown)


ext/DiffEqBaseEnzymeExt.jl
```diff
@@ -52,9 +49,7 @@ function Enzyme.EnzymeRules.reverse(config::Enzyme.EnzymeRules.ConfigWidth{1},
         end
         ptr.dval .+= darg
     end
-    for v in dres.u
-        v .= 0
-    end
+    dres = Enzyme.make_zero(dres)
```
wsmoses, line 52: @ChrisRackauckas this is incorrect.

It will not update the values of dres in place, instead allocating a new variable.

This will lead to correctness issues if the rule is called within a loop.

ChrisRackauckas: is there a `make_zero!(v)` we can use?

wsmoses: not yet, but I'll start making one. We haven't historically seen too much of a need for and always gottena way with the broadcast to zero (like here). But obviously generic structs need generic zeros

PR #1055, opened 2024-06-06
Revert "use Enzyme make_zero"

ChrisRackauckas:
Reverts SciML/DiffEqBase.jl#1054 @wsmoses @m-bossart


wsmoses:
FWIW you can still (and should) keep the change to replace deep-copy and broadcast in the fwd pass with make zero, the reverse pass broadcast tho is the part that would cause issues

PR #1056, opened 2024-06-07
make `ODE_DEFAULT_UNSTABLE_CHECK` just return false

oscardssmith:
Since https://github.com/SciML/SciMLBase.jl/pull/693 the instability check is only triggered on accepted steps, so we can save some time and just never return unstable by default.


oscardssmith:
interesting. The test failure suggests that we do sometimes accept a step of size Inf. not sure how that's possible.

PR #1070, opened 2024-08-06
feat: perform limited DAE initialization for null integrators/solutions

AayushSabharwal:
Requires https://github.com/SciML/SciMLBase.jl/pull/760 for the `constructorof` method

## Checklist


src/solve.jl
```diff
@@ -675,7 +678,12 @@ function build_null_solution(prob::AbstractDEProblem, args...;
     end
 
     timeseries = [Float64[] for i in 1:length(ts)]
-
+    
+    if SciMLBase.has_initializeprob(prob.f) && SciMLBase.has_initializeprobpmap(prob.f)
+        initializeprob = prob.f.initializeprob
+        nlsol = solve(initializeprob)
+        @set! prob.p = prob.f.initializeprobpmap(prob, nlsol)
```
ChrisRackauckas, line 685: isn't the map in-place?

AayushSabharwal: Yeah. I think I wrote this when it was out-of-place, and didn't update it. Will edit and check

AayushSabharwal: No wait it's out of place 😅 with the idea that solving should not change the problem since then there's a chance for stateful errors where solving the same problem multiple times leads to different results

PR #1071, opened 2024-08-12
add integrator accessors for dispatch purposes

isaacsas:
This is needed for https://github.com/SciML/DiffEqCallbacks.jl/pull/229, but shouldn't get merged until we are settled that https://github.com/SciML/DiffEqCallbacks.jl/pull/229 has the correct approach (and doesn't need changes here).

### DIFF
--- a/src/integrator_accessors.jl
+++ b/src/integrator_accessors.jl
@@ -2,8 +2,8 @@
 # StochasticDiffEq and provide dispatch points that JumpProcesses and others can use.
 
 get_tstops(integ::DEIntegrator) =
-    error("get_tstops not implemented for integrators of type $(typeof(integ))")
+    error("get_tstops not implemented for integrators of type $(nameof(typeof(integ)))")
 get_tstops_array(integ::DEIntegrator) =
-    error("get_tstops_array not implemented for integrators of type $(typeof(integ))")
+    error("get_tstops_array not implemented for integrators of type $(nameof(typeof(integ)))")
 get_tstops_max(integ::DEIntegrator) =
-    error("get_tstops_max not implemented for integrators of type $(typeof(integ))")
+    error("get_tstops_max not implemented for integrators of type $(nameof(typeof(integ)))")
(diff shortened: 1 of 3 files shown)


src/integrator_accessors.jl
```diff
@@ -0,0 +1,9 @@
+# the following are setup per how integrators are implemented in OrdinaryDiffEq and
+# StochasticDiffEq and provide dispatch points that JumpProcesses and others can use.
+
+# returns if the given integrator has an integ.iter field
+hasiter(integ) = true
```
ChrisRackauckas, line 5: `hasfield` and `hasproperty` work at compile time though, so I'm not sure if this one is the most essential?

src/integrator_accessors.jl
```diff
@@ -0,0 +1,9 @@
...
+
+get_tstops(integ) = integ.opts.tstops
+get_tstops_array(integ) = get_tstops(integ).valtree
+get_max_tstops(integ) = maximum(get_tstops_array(integ))
```
ChrisRackauckas, line 9: Does this need to get specialized differently for Sundials and ODEInterface? If so, then we should probably leave the default as empty and do per-package implementations.

src/integrator_accessors.jl
isaacsas: Ok, I wasn’t sure about if `hasfield` is compile time, but since it is I can drop this.

src/integrator_accessors.jl
isaacsas: I guess they probably do — I don’t know their internals too well. So then we’ll need versions for OrdinaryDiffEq, StochasticDiffEq, JumpProcesses, and those two you mention?

isaacsas:
Format failures seem to be in other files so can be handled in a separate PR.

src/solve.jl
```diff
@@ -97,7 +97,9 @@ const allowedkeywords = (:dense,
     :trace_level,
     :store_trace,
     # Termination condition for solvers
-    :termination_condition)
+    :termination_condition,
+    # whether to alias the user tstops array (JumpProcesses.SSAStepper only at this time)
+    :alias_tstops)
```
ChrisRackauckas, line 102: Let's not do this for now. I'll add a full aliasing system next week.

isaacsas:
I removed the alias keyword.

PR #1075, opened 2024-08-27
Revert "Fix ordering of the promotion [Take II]"

ChrisRackauckas:
Reverts SciML/DiffEqBase.jl#1032. Sees if that fixes some of the Tracker tests that started failing.

### DIFF
--- a/ext/DiffEqBaseTrackerExt.jl
+++ b/ext/DiffEqBaseTrackerExt.jl
@@ -97,15 +97,9 @@ Tracker.@grad function DiffEqBase.solve_up(prob,
         },
         u0, p, args...;
         kwargs...)
-    sol, pb_f = DiffEqBase._solve_adjoint(
-        prob, sensealg, Tracker.data(u0), Tracker.data(p),
+    out = DiffEqBase._solve_adjoint(prob, sensealg, Tracker.data(u0), Tracker.data(p),
         SciMLBase.TrackerOriginator(), args...; kwargs...)
-
-    if sol isa AbstractArray
-        !hasfield(typeof(sol), :u) && return sol, pb_f # being safe here
-        return sol.u, pb_f # AbstractNoTimeSolution isa AbstractArray
-    end
-    return convert(AbstractArray, sol), pb_f
+    Array(out[1]), out[2]
 end
 
 end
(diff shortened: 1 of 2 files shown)


ext/DiffEqBaseTrackerExt.jl
```diff
@@ -97,15 +97,9 @@ Tracker.@grad function DiffEqBase.solve_up(prob,
...
-    sol, pb_f = DiffEqBase._solve_adjoint(
-        prob, sensealg, Tracker.data(u0), Tracker.data(p),
+    out = DiffEqBase._solve_adjoint(prob, sensealg, Tracker.data(u0), Tracker.data(p),
         SciMLBase.TrackerOriginator(), args...; kwargs...)
-
-    if sol isa AbstractArray
-        !hasfield(typeof(sol), :u) && return sol, pb_f # being safe here
-        return sol.u, pb_f # AbstractNoTimeSolution isa AbstractArray
-    end
-    return convert(AbstractArray, sol), pb_f
```
avik-pal, line 108: probably reverting just this will do

ChrisRackauckas:
Can you make that PR? I'm out and about right now.

ext/DiffEqBaseTrackerExt.jl
avik-pal: Let's wait and see https://github.com/SciML/DiffEqBase.jl/pull/1076

avik-pal:
Source of the errors doesn't seem to be that PR. tests are still failing https://github.com/SciML/DiffEqBase.jl/actions/runs/10584877403/job/29330163818?pr=1075#step:6:1828

avik-pal:
@ChrisRackauckas based on the stacktrace I have a feeling this is an OrdinaryDiffEq issue. I would expect the integrator cache to hold TrackedArrays and not regular arrays when using discrete adjoints.

ChrisRackauckas:
I was starting to fear that it might be related to the OrdinaryDiffEq split PRs. But I don't see how? 

https://github.com/search?q=repo%3ASciML%2FOrdinaryDiffEq.jl%20value&type=code

Everything is directly using DiffEqBase.value, which is the thing that should be de-tracking as appropriate.

PR #1079, opened 2024-09-04
make `promote_u0` use `isequal` rather than `==`

oscardssmith:
otherwise NaN params fail


ChrisRackauckas:
Tests fail

ChrisRackauckas:
@ErikQQY do you know what's going on with the BVP tests?

ErikQQY:
When using `Shooting` solvers, apart from the ODE solvers, we should also specify the nonlinear solvers to the `Shooting` solver, I noticed DifferentialEquations.jl CIs also failed because of this, I can send a patch.

PR #1089, opened 2024-09-28
Deprecate DiffEqBase.fastpow for FastPower.fastpower

ChrisRackauckas:
This requires the registration to finish but will finally make it more accessible to Symbolics and other tooling.


oscardssmith:
what is `FastPower`?

ChrisRackauckas:
Same code https://github.com/SciML/FastPower.jl. Just moved to a package for easier downstream usage.

PR #1093, opened 2024-10-09
Remove `Tricks`

oscardssmith:
Tricks are for kids!

(and as of Julia 1.10, `static_hasmethod` is just `hasmethod` in Base)


Project.toml
```diff
@@ -103,7 +102,6 @@ Static = "1"
 StaticArraysCore = "1.4"
 Statistics = "1"
 Tracker = "0.2"
-Tricks = "0.1.6"
 TruncatedStacktraces = "1"
```
avik-pal, line 105: We should also remove this from packages

oscardssmith: yes.

PR #1096, opened 2024-10-16
remove `hasmethod` from callback

oscardssmith:
possibly fixes https://github.com/SciML/DiffEqBase.jl/pull/1096
I don't think we need to support it since this is an internal feature and OrdinaryDiffEq now uses the better version of this.


oscardssmith:
good to merge?

PR #1101, opened 2024-10-29
feat: use specialization in NonlinearProblems

### LINKED ISSUE #417: The precompilation of this package is likely excessive and not very useful
This package does a more or less exhaustive combination of problems, algorithms and element types

https://github.com/SciML/NonlinearSolve.jl/blob/d434b8d48086fa799ea344694c90a997dae22d48/src/NonlinearSolve.jl#L96-L98

https://github.com/SciML/NonlinearSolve.jl/blob/d434b8d48086fa799ea344694c90a997dae22d48/src/NonlinearSolve.jl#L137-L144

A total of 104 combinations of problems and algorithms are compiled in total.

At the same time, the problem specializes on the input function:

```
julia> typeof(prob)
NonlinearProblem{Float64, false, Float64, NonlinearFunction{false, SciMLBase.FullSpecialize, var"#1#2",
```

(the `var"#1#2"` is an anonymous function defined locally). so the compiled code is unlikely to be valid for users (which have their own functions that they solve).

It is not clear to me that there is any point at all to the precompilation of `solve(prob, alg)` after the very first call has been compiled.

avik-pal:
needs https://github.com/SciML/SciMLBase.jl/pull/836
fixes https://github.com/SciML/NonlinearSolve.jl/issues/417

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.159.0"
+version = "6.160.0"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Did you make sure adjoints unwrap?

Ickaser:
Any progress here?

ChrisRackauckas:
This probably needs to be moved to NonlinearSolveBase.jl, and get a PR to SciMLSensitivity to unwrap where necessary. I could do the latter part if someone does the former.

PR #1104, opened 2024-11-12
docs: document late binding tstops

AayushSabharwal:
## Checklist

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -800,7 +800,8 @@ section at the end of this page for some example usage.
   multistep methods), then `tstops` will use an interpolation,
   matching the behavior of `saveat`. If a method cannot change timesteps and
   also cannot interpolate, then `tstops` must be a multiple of `dt` or else an
-  error will be thrown. Default is `[]`.
+  error will be thrown. `tstops` may also be a function accepting the parameter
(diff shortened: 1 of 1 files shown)


src/solve.jl
```diff
@@ -800,7 +800,8 @@ section at the end of this page for some example usage.
   multistep methods), then `tstops` will use an interpolation,
   matching the behavior of `saveat`. If a method cannot change timesteps and
   also cannot interpolate, then `tstops` must be a multiple of `dt` or else an
-  error will be thrown. Default is `[]`.
+  error will be thrown. `tstops` may also be a function accepting the parameter
```
ChrisRackauckas, line 803:
```suggestion
  error will be thrown. `tstops` may also be a function `tstops(p, tspan)`, accepting the parameter
```

ChrisRackauckas:
If this is not implemented in DiffEqBase, then we need to make a trait for it, and trait check its usage in DiffEqBase so it's only given to appropriate solvers.

AayushSabharwal:
I guess it doesn't matter, since we can't implement the late binding tstops here if we can't guarantee/check if `add_tstop!` is valid.

AayushSabharwal:
Trait defined in https://github.com/SciML/SciMLBase.jl/pull/858

PR #1108, opened 2025-01-05
Add `ODE_DEFAULT_NORM` overload for GTPSA

mattsignorelli:
## Checklist

- [X] Appropriate tests were added
- [X] Any code changes were done in a way that does not break public API
- [X] All documentation related to code changes were updated
- [X] The new code follows the
- [X] Any new documentation only uses public API
  
## Additional context

I added an overload for `ODE_DEFAULT_NORM` with `TPS` types where it calculates the 2-norm of the scalar (0th order) part of the `TPS`, and is necessary because the GTPSA C library does not allow `sqrt(TPS(0))` as the derivatives are undefined. 

The calculation of the norm of the scalar part follows from what I see in the `Tracker` extension. If instead it should calculate a norm including all monomial coefficients/partial derivatives, I can quickly change that to the function `GTPSA.normTPS`.

### DIFF
--- a/ext/DiffEqBaseGTPSAExt.jl
+++ b/ext/DiffEqBaseGTPSAExt.jl
@@ -13,13 +13,21 @@ end
 value(x::TPS) = scalar(x)
 value(::Type{TPS{T}}) where {T} = T
 
-ODE_DEFAULT_NORM(u::TPS, t) = @fastmath abs(value(u))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
>  If instead it should calculate a norm including all monomial coefficients/partial derivatives, I can quickly change that to the function GTPSA.normTPS

That would be better because then it would be guaranteed convergent.

mattsignorelli:
Done.

PR #1112, opened 2025-02-10
Import `totallength` in ForwardDiff extension

devmotion:
I think this is necessary and should fix downstream issues I just ran into:

```julia
  Got exception outside of a @test
  MethodError: no method matching totallength(::Vector{ForwardDiff.Dual{ForwardDiff.Tag{Pumas.Tag, ForwardDiff.Dual{ForwardDiff.Tag{Pumas.Tag, Float64}, Float64, 2}}, ForwardDiff.Dual{ForwardDiff.Tag{Pumas.Tag, Float64}, Float64, 2}, 1}})
  The function `totallength` exists, but no method is defined for this combination of argument types.
  
  Closest candidates are:
    totallength(::ForwardDiff.Dual)
     @ DiffEqBaseForwardDiffExt ~/.julia/packages/DiffEqBase/DnAEG/ext/DiffEqBaseForwardDiffExt.jl:504
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.162.1"
+version = "6.162.2"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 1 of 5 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
fuck I bungled the hotfix sorry. I don't know how this wasn't caught in the first set of tests but I tested this locally on an autodiff solve case and we seem good now.

PR #1120, opened 2025-02-23
Add error_control keyword

ErikQQY:
Add an `error_control` keyword

For https://github.com/SciML/BoundaryValueDiffEq.jl/pull/287


ErikQQY:
Tests failure not related?

ChrisRackauckas:
Should this just use `controller` like?

https://docs.sciml.ai/DiffEqDocs/stable/extras/timestepping/

ErikQQY:
Yeah, agreed, we might just use the same convention to avoid confusion here

PR #1127, opened 2025-03-19
Draft: simplify internal_itp

oscardssmith:
to match https://github.com/SciML/NonlinearSolve.jl/pull/548 (although this also fixes `k2` to `2` since paying for a Floating point pow is expensive. 

Do not merge until https://github.com/SciML/NonlinearSolve.jl/pull/548 is merged.

### DIFF
--- a/src/internal_itp.jl
+++ b/src/internal_itp.jl
@@ -36,78 +36,59 @@ function SciMLBase.solve(prob::IntervalNonlinearProblem{IP, Tuple{T, T}}, alg::I
     ϵ = eps(T)
     if iszero(fl)
         return SciMLBase.build_solution(prob, alg, left, fl;
-            retcode = ReturnCode.ExactSolutionLeft, left = left,
-            right = right)
+            retcode = ReturnCode.ExactSolutionLeft, left, right)
     elseif iszero(fr)
         return SciMLBase.build_solution(prob, alg, right, fr;
-            retcode = ReturnCode.ExactSolutionRight, left = left,
-            right = right)
+            retcode = ReturnCode.ExactSolutionRight, left, right)
     end
-    #defining variables/cache
-    k1 = T(alg.k1)
     k2 = T(alg.k2)
+    span = abs(right - left)
+    k1 = T(alg.alg_k1)
     n0 = T(alg.n0)
-    n_h = ceil(log2(abs(right - left) / (2 * ϵ)))
-    mid = (left + right) / 2
-    x_f = (fr * left - fl * right) / (fr - fl)
-    xt = left
-    xp = left
-    r = zero(left) #minmax radius
-    δ = zero(left) # truncation error
-    σ = 1.0
-    ϵ_s = ϵ * 2^(n_h + n0)
-    i = 0 #iteration
-    while i <= maxiters
-        #mid = (left + right) / 2
+    n_h = exponent(span / (2 * ϵ))
+    ϵ_s = ϵ * exp2(n_h + n0)
+    T0 = zero(fl)
+
+    i = 1
+    while i ≤ maxiters
+        stats.nsteps += 1
         span = abs(right - left)
-        r = ϵ_s - (span / 2)
-        δ = k1 * (span^k2)
-
-        ## Interpolation step ##
-        x_f = left + (right - left) * (fl / (fl - fr))
-
-        ## Truncation step ##
-        σ = sign(mid - x_f)
-        if δ <= abs(mid - x_f)
-            xt = x_f + (σ * δ)
-        else
-            xt = mid
-        end
-
-        ## Projection step ##
-        if abs(xt - mid) <= r
-            xp = xt
-        else
-            xp = mid - (σ * r)
-        end
-
-        ## Update ##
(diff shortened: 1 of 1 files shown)


src/internal_itp.jl
```diff
@@ -36,78 +35,58 @@ function SciMLBase.solve(prob::IntervalNonlinearProblem{IP, Tuple{T, T}}, alg::I
...
         return SciMLBase.build_solution(prob, alg, right, fr;
-            retcode = ReturnCode.ExactSolutionRight, left = left,
-            right = right)
+            retcode = ReturnCode.ExactSolutionRight, left, right)
     end
-    #defining variables/cache
-    k1 = T(alg.k1)
-    k2 = T(alg.k2)
+    span = abs(right - left)
+    k1 = T(alg.scaled_k1)/span
```
oscardssmith, line 44: do you know if the lack of scaling based on `span` was intentional here? The NonlinearSolveBase verison already had it.

oscardssmith:
I think this is ready now (but I have started nuking CI from orbit with ForwardDiff PRs so we'll see if CI here ever completes)

src/internal_itp.jl
```diff
@@ -36,78 +35,56 @@ function SciMLBase.solve(prob::IntervalNonlinearProblem{IP, Tuple{T, T}}, alg::I
...
-            xp = mid - (σ * r)
-        end
+        xt = ifelse(δ ≤ abs(diff), x_f + copysign(δ, diff), mid)  # Truncation Step
 
-        ## Update ##
-        tmin, tmax = minmax(left, right)
-        xp >= tmax && (xp = prevfloat(tmax))
-        xp <= tmin && (xp = nextfloat(tmin))
+        xp = ifelse(abs(xt - mid) ≤ r, xt, mid - copysign(r, diff))  # Projection Step
+        if span < 2ϵ
```
DaniGlez, line 64: Is this correct? `ϵ` here is the machine epsilon `eps(T)` (not `eps(left)`), i.e. a _relative_ difference, compared to an absolute value `span`. If `left` and `right` are >> 1 in absolute value and different this condition cannot be true, while if they are << 1 in absolute value this condition can be met even if the interval is not tight.

oscardssmith: good catch. This was based on https://github.com/SciML/NonlinearSolve.jl/pull/548/ where we have an abstol. I think this branch can just be deleted.

src/internal_itp.jl
```diff
@@ -36,78 +35,56 @@ function SciMLBase.solve(prob::IntervalNonlinearProblem{IP, Tuple{T, T}}, alg::I
...
+        elseif yps < T0
+            left, fl = xp, yp
         else
-            left = prevfloat_tdir(xp, prob.tspan...)
-            right = xp
-            return SciMLBase.build_solution(prob, alg, left, f(left);
-                retcode = ReturnCode.Success, left = left,
-                right = right)
+            return SciMLBase.build_solution(
+                prob, alg, xp, yps; retcode = ReturnCode.Success, left, right
```
DaniGlez, line 77: Another question: here we're no longer contracting the left - right interval for the solution object, does this change behavior in any case? I'm investigating a small outcome variation in a test in some internal code of ours that opts into `rootfind=SciMLBase.RightRootFind` for a callback and it seemingly appears between DiffEqBase v6.167.0 and v6.167.2 (though this is just a tentative assessment as of now).

DaniGlez: Yes, that makes sense; my concern is that while the actual crossing (`xp`) is provided as the "main" value of the solution `sol.u`, the way this solution is used ( https://github.com/SciML/DiffEqBase.jl/blob/master/src/callbacks.jl#L365 ) this value is seemingly not used; instead, the `left` and `right` fields of the solution seem to be the relevant ones (depending on the user-defined rootfinding side). However, in this file, the solution is built without updating the `left` and `right` variables, which could be arbitrarily far from `xp` (which is I guess the value we want) if the iteration hitting this branch is particularly lucky. The previous code basically took `xp` (the value producing the exact zero) as the `right` value and its prevfloat as the `left` value.

DaniGlez: I guess another connected question (that is probably not very relevant practically except for sorting simultaneous events) is what should happen if there are multiple zeros in a row (e.g. for a sequence of adjacent floats for `t` the condition produces something like `-a, 0, 0, 0, a` where `a` is small). It's not as important, because for this to happen you need to have small absolute values of `t` (so that the expected difference between consecutive values is comparable to numerical error) and/or a flat-ish `f`, but maybe it's worth thinking about what's the desirable behavior in that case (it should be cheap to seek the first or last zero with a few refinement iterations, probably with straightforward bisection for smaller overhead).

DaniGlez: Upon further testing it appears that setting `left` and `right` to `xp` here reverts my test results to their pre-6.167.1 values, I will make a PR once I double-check.

DaniGlez: Upon further testing it appears that setting `left` and `right` to `xp` here reverts my test results to their pre-6.167.1 values, I will make a PR once I double-check.

PR #1132, opened 2025-03-26
support ForwardDiff 1.0

oscardssmith:


oscardssmith:
I've run the tests locally with sufficient deved versions to get ForwardDiff1.0 in the tests and this passes locally.

PR #1136, opened 2025-03-31
remove ITP early termination condition

oscardssmith:
Added for similarity to BracketingNonlinearSolve, but not useful here since we always want ULP accurate solves here. Found by https://github.com/SciML/DiffEqBase.jl/pull/1127#discussion_r2021827114


ChrisRackauckas:
What's this related to?

oscardssmith:
oops, the commit had the description, but the PR lost it.

PR #1137, opened 2025-04-01
Tighten the interval returned by InternalITP when there's an exact zero crossing

DaniGlez:
Background is here: https://github.com/SciML/DiffEqBase.jl/pull/1127/files#r2022436123

To summarize: inside the branch where the callback condition is evaluated to an exact zero, this PR fills the `left` and `right` fields of the solution with the location of the zero crossing instead of the values at the start of that ITP iteration.

### DIFF
--- a/src/internal_itp.jl
+++ b/src/internal_itp.jl
@@ -41,7 +41,7 @@ function SciMLBase.solve(prob::IntervalNonlinearProblem{IP, Tuple{T, T}}, alg::I
             retcode = ReturnCode.ExactSolutionRight, left, right)
     end
     span = abs(right - left)
-    k1 = T(alg.scaled_k1)/span
+    k1 = T(alg.scaled_k1) / span
     n0 = T(alg.n0)
     n_h = exponent(span / (2 * ϵ))
     ϵ_s = ϵ * exp2(n_h + n0)
@@ -61,11 +61,6 @@ function SciMLBase.solve(prob::IntervalNonlinearProblem{IP, Tuple{T, T}}, alg::I
         xt = ifelse(δ ≤ abs(diff), x_f + copysign(δ, diff), mid)  # Truncation Step
 
         xp = ifelse(abs(xt - mid) ≤ r, xt, mid - copysign(r, diff))  # Projection Step
-        if span < 2ϵ
-            return SciMLBase.build_solution(
-                prob, alg, xt, f(xt); retcode = ReturnCode.Success, left, right
-            )
-        end
         yp = f(xp)
         yps = yp * sign(fr)
         if yps > T0
--- a/src/internal_itp.jl
+++ b/src/internal_itp.jl
@@ -74,7 +74,7 @@ function SciMLBase.solve(prob::IntervalNonlinearProblem{IP, Tuple{T, T}}, alg::I
             left, fl = xp, yp
         else
             return SciMLBase.build_solution(
-                prob, alg, xp, yps; retcode = ReturnCode.Success, left=xp, right=xp
+                prob, alg, xp, yps; retcode = ReturnCode.Success, left = xp, right = xp
             )
         end
 
--- a/src/internal_itp.jl
+++ b/src/internal_itp.jl
@@ -74,7 +74,7 @@ function SciMLBase.solve(prob::IntervalNonlinearProblem{IP, Tuple{T, T}}, alg::I
             left, fl = xp, yp
         else
             return SciMLBase.build_solution(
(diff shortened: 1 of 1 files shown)


DaniGlez:
BTW, in that branch we assume that `yps` is an exact zero because it's an `else` after `if yps > T0` and `elseif yps < T0`, but a NaN would also fall into that branch; should an appropriate exception be raised in that case?

oscardssmith:
I do think this code is missing a branch for if right and left get within 1 ULP of each other. I think currently for a function like `f(x) = x^2-2` it will run to maxiters since the function never hits 0.

DaniGlez:
Shouldn't that case be covered by the `if` later down in the function?

https://github.com/SciML/DiffEqBase.jl/blob/0874ff4ef8df2c74c843ce534e64ab3cdd40efb4/src/internal_itp.jl#L84

oscardssmith:
This looks good to me. Would you object to either you adding the change from https://github.com/SciML/DiffEqBase.jl/actions/runs/14193284895/job/39762654888?pr=1137 or vice versa? There are some CI failures here that look potentially related (on both branches) so it would be nice to consolidate the changes so I can only investigate once.

DaniGlez:
I've incorporated the change from #1136 here for convenience

oscardssmith (approved):
Looks good to me! Let's see if CI agrees.

DaniGlez:
CI not too happy, it seems

DaniGlez:
In this CI run, the only test failure that I can connect to callback rootfinding with `InternalITP` is this one: https://github.com/SciML/DiffEqBase.jl/actions/runs/14203971489/job/39797246048?pr=1137#step:6:17566

```julia 
LoadError: MethodError: no method matching exponent(::ReverseDiff.TrackedReal{Float64, Float64, Nothing})
```

oscardssmith:
I think the CI results might be bogus let's try again. We really don't want to be autodiffing the ITP so hopefully we aren't...

DaniGlez:
First one looks relevant: https://github.com/SciML/DiffEqBase.jl/actions/runs/14233499341/job/39888542399?pr=1137

DaniGlez:
Yeah, I think I can reproduce it (from https://github.com/SciML/DelayDiffEq.jl/blob/master/test/integrators/events.jl#L8 ). It seems that the callback location is now shifted 1 ulp to the right compared to previous behaviour. So the test failure could be remediated by:

1) When building the solution, shift the `left` field 1 ulp to the left, which I think would restore pre-6.167.1 behavior
2) Accept the new behavior, since the crossing is an exact zero and arguably the behavior is more sensible, and as a consequence amend the relevant `DelayDiffEq` tests to replace the `prevfloat(2.6)` instances in that test set by a plain `2.6`

oscardssmith:
@ChrisRackauckas thoughts on the right approach here?

ChrisRackauckas:
Yeah if we are now better at stopping at the exact zero, that's better. We should accept that and change the test. Our goal has always been to get there, it just wasn't robust before.

oscardssmith:
sounds good. In that case, I think this is ready to merge and we can fix the test.

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/14584553847/job/40907662031?pr=2681#step:5:1336 this failure might be related?

oscardssmith:
Seems highly unlikely. That failure appears to be a case where there are 2 possible zeros and it's picking a different one.

DaniGlez:
It's related, it is the same test failure we were discussing in this very thread 😅 see my earlier comment and the subsequent replies: https://github.com/SciML/DiffEqBase.jl/pull/1137#issuecomment-2775916197

DaniGlez:
The TL;DR is that as of current DiffEqBase.jl#master those tests no longer need the `prevfloat` (i.e. you can just have the test at t = `2.6` instead of `prevfloat(2.6)`). I can PR the fix to the DelayDiffEq repo but I think I'd need a new release of DiffEqBase to have CI grab the current version of DiffEqBase instead of 6.167.2.

ChrisRackauckas:
The release is already out. 

> It's related, it is the same test failure we were discussing in this very thread 😅 see my earlier comment and the subsequent replies: https://github.com/SciML/DiffEqBase.jl/pull/1137#issuecomment-2775916197

Completely forgot 😅 . If you can make that downstream PR that would be great.

DaniGlez:
Sounds great! I see you've already made the PR.

PR #1138, opened 2025-04-01
Use NonlinearAliasSpecifier for NonlinearProblem

jClugstor:
## Checklist

## Additional context
This adds support for using the NonlinearAliasSpecifier to specify if `u0` is aliased or not. 

The reason I want to put it here for now is because with the current way that NonlinearSolve is structured, it would take significant refactoring to do this in a similar way this was done in OrdinaryDiffEq, LinearSolve etc. as well as supporting aliasing other variables. `alias_u0` is an argument to many of the internal NonlinearSolve functions, and it would take major changes to change it to using a `NonlinearAliasSpecifier`, if it even makes sense to do so. Additionally, the `__init` functions are spread out between the different solving algorithms, so there's no single place in NonlinearSolve to put this, like in the other packages. 

This will deprecate the `alias_u0` kwarg, and allow users to use a `NonlinearAliasSpecifier` to specify if `u0` is aliased. 
A `Bool` `alias` kwarg is supported as well.


src/solve.jl
```diff
@@ -1089,13 +1089,38 @@ function solve(prob::NonlinearProblem, args...; sensealg = nothing,
         sensealg = prob.kwargs[:sensealg]
     end
 
+
```
ChrisRackauckas, line 1092: This is now handled in nonlinearsolvebase?

jClugstor: The `sensealg` part?

ChrisRackauckas: No the solve call for NonlinearProblem. https://github.com/SciML/NonlinearSolve.jl/blob/master/lib/NonlinearSolveBase/src/solve.jl#L5

ChrisRackauckas: Actually, double check the stack trace during a nonlinearsolve solve? Does it hit here? I thought it was eliminated.

jClugstor: It does hit here, unless you use the default solver, so with 
```julia
using NonlinearSolve

f(u, p) = u .* u .- p
u0 = [1.0, 1.0]
p = 2.0
prob = NonlinearProblem(f, u0, p)
solve(prob)
```

it hits the `__solve` for the NonlinearSolvePolyAlgorithm here instead: 
https://github.com/SciML/NonlinearSolve.jl/blob/b182ae1b8584763004392b335e094545d9f1be58/lib/NonlinearSolveBase/src/solve.jl#L124-L128

which is why I thought I needed to put this here.

ChrisRackauckas: oh I guess we'll move that later.

ChrisRackauckas (approved):

PR #1140, opened 2025-04-03
chore: pass `sensealg` to `solve_call` kwargs

DhairyaLGandhi:
Seems passing `sensealg` to `solve_call` kwargs got missed

## Checklist


ChrisRackauckas:
Why would this matter? Since the dispatching is on the solve_up part, after that point the sensealg doesn't matter?

DhairyaLGandhi:
Running under `ForwardDiff` and the Debugger, @AayushSabharwal and I found that `sensealg` gets missed inside `solve_call` causing the subsequent steps to get `sensealg = nothing` even when one was provided. Thanks for the clarification! The SciMLSensitivity rules don't get called with ForwardDiff, so perhaps this is unimportant.

PR #1142, opened 2025-04-23
Properly extend unitfulvalue

ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2675


ChrisRackauckas:
Extensively tested locally. Will add the ODE test downstream as well.

PR #1143, opened 2025-04-24
fix: handle `u0` and `p` in `kwargs` of `get_concrete_problem`

AayushSabharwal:
The `u0` and `p` in the `prob` returned from `get_updated_symbolic_problem` were ignored because the ones in `kwargs` take priority.


AayushSabharwal:
Alright something is broken

PR #1144, opened 2025-04-24
Better handle dictionaries in anytypedual

ChrisRackauckas:


ext/DiffEqBaseForwardDiffExt.jl
```diff
@@ -447,6 +447,8 @@ function anyeltypedual(
     Any
 end
 
+
+
```
ChrisRackauckas, line 451:
```suggestion
```

PR #1145, opened 2025-04-24
fix: hotfix usage of `get_updated_symbolic_problem`

AayushSabharwal:
Oops 😅 😅

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -1249,8 +1249,9 @@ function get_concrete_problem(prob::AbstractJumpProblem, isadapt; kwargs...)
 end
 
 function get_concrete_problem(prob::SteadyStateProblem, isadapt; kwargs...)
+    oldprob = prob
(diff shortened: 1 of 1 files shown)


src/solve.jl
```diff
@@ -1271,8 +1273,9 @@ function get_concrete_problem(prob::NonlinearProblem, isadapt; kwargs...)
 end
 
 function get_concrete_problem(prob::NonlinearLeastSquaresProblem, isadapt; kwargs...)
+    oldprob = prob
     prob = get_updated_symbolic_problem(_get_root_indp(prob), prob; kwargs...)
```
ChrisRackauckas, line 1277: oldprob?

AayushSabharwal: prior to calling `get_updated_symbolic_problem`

AayushSabharwal: The check below used to compare `prob !== prob` which is of course always false

PR #1149, opened 2025-05-05
feat: pass `resid_prototype` as `u` to null oop `NonlinearLeastSquaresProblem`

AayushSabharwal:


src/solve.jl
```diff
@@ -776,7 +776,7 @@ function build_null_solution(
         resid = isnothing(prob.f.resid_prototype) ? Float64[] : copy(prob.f.resid_prototype)
         prob.f(resid, prob.u0, prob.p)
     else
-        resid = prob.f(prob.u0, prob.p)
+        resid = prob.f(prob.f.resid_prototype, prob.p)
```
ChrisRackauckas, line 779: This is wrong though?

ChrisRackauckas: Only do this if `u0 === nothing`? Why is this needed though?

AayushSabharwal: `build_null_solution` is only called when `prob.u0 === nothing`

AayushSabharwal: This is necessary because otherwise the OOP form doesn't know what sort of array to create

ChrisRackauckas: I see, makes sense in its own way.

ChrisRackauckas (approved):

PR #1150, opened 2025-05-08
Possible solution to repeated callbacks at dt=0

ChrisRackauckas:
Possible solution to https://github.com/SciML/ModelingToolkit.jl/issues/3327 . I don't like this solution though because the way it works is that it allows it to catch a dt=0 case by eliminating the bottom_t from matching t, and then checking if the nonlinear solver did not converge, and then setting error_occured on that. I think it's possible to construct a bug where in the interval (t0,t1) you have an event `t2` and an "event" found at t0, this will search for the first event so it will narrow the intervals to `t3 < t2` and then do the search on (t0,t3), then fail, and so it would say there is no event.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -363,13 +363,16 @@ function bisection(
         f, tup, t_forward::Bool, rootfind::SciMLBase.RootfindOpt, abstol, reltol;
         maxiters = 1000)
     if rootfind == SciMLBase.LeftRootFind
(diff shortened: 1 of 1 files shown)


oscardssmith:
pinging @BenChung since I know he's thought a lot about this.

ChrisRackauckas:
It ended up even weirder than I thought. It was https://github.com/SciML/OrdinaryDiffEq.jl/pull/2695 😅

PR #1155, opened 2025-05-14
Mark null ODE solutions as dense

ChrisRackauckas:
We actually know the entire time series for the solution, so we should state that we know that via `dense=true`.

```julia
using OrdinaryDiffEqDefault, ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D

@variables x(t) y(t)
eqs = [y ~ x, y ~ sin(t)]
@mtkbuild sys = ODESystem(eqs, t)

tspan = (0.0, 10.0)
prob = ODEProblem(sys, [], tspan)
sol = solve(prob)
using Plots
plot(sol, idxs = [x])
savefig("plot.png")
```

Before and after

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Before:
image

After:
plot

PR #1158, opened 2025-05-24
Setup Mooncake rrule

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqBase.jl/issues/1103


ChrisRackauckas:
I have a hack working. In the rule it thinks it's chainrules so I cannot make it skip the rule when using Mooncake, but it at least kind of works. I need that issue solved to make the discrete adjoint actually skip things

ext/DiffEqBaseMooncakeExt.jl
```diff
@@ -0,0 +1,22 @@
...
+        Union{Nothing, DiffEqBase.AbstractSensitivityAlgorithm},
+        Any,
+        Any,
+        Any,
+    },
+    true,
+    )
+
+Mooncake.@zero_adjoint Mooncake.MinimalCtx Tuple{typeof(DiffEqBase.numargs), Any}
+Mooncake.@mooncake_overlay DiffEqBase.set_mooncakeoriginator_if_mooncake(x::ADOriginator) = MooncakeOriginator
```
yebai, line 20: @AstitvaAggarwal, can you replace `Mooncake.@mooncake_overlay` with a handwritten rule?

PR #1159, opened 2025-05-24
WIP: Try fixing Mooncake originator

ChrisRackauckas:
This doesn't trigger either, so it seems to be a bug in Mooncake https://github.com/chalk-lab/Mooncake.jl/issues/587

### DIFF
--- a/ext/DiffEqBaseMooncakeExt.jl
+++ b/ext/DiffEqBaseMooncakeExt.jl
@@ -2,7 +2,7 @@ module DiffEqBaseMooncakeExt
 
 using DiffEqBase, Mooncake
 using DiffEqBase: SciMLBase
-using SciMLBase: ADOriginator, MooncakeOriginator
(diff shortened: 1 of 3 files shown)


ext/DiffEqBaseMooncakeExt.jl
```diff
@@ -17,6 +17,6 @@ Mooncake.@from_rrule(
     )
 
 Mooncake.@zero_adjoint Mooncake.MinimalCtx Tuple{typeof(DiffEqBase.numargs), Any}
-Mooncake.@mooncake_overlay DiffEqBase.set_mooncakeoriginator_if_mooncake(x::ADOriginator) = MooncakeOriginator
+Mooncake.@mooncake_overlay DiffEqBase.set_mooncakeoriginator_if_mooncake(x::ChainRulesOriginator) = MooncakeOriginator()
```
yebai, line 20: AstitvaAggarwal, this needs to be replaced with a handwritten Mooncake rule.

PR #1162, opened 2025-05-30
Ensure internal ITP bounds are in proper order

dcourteville:
This is a fix for https://github.com/SciML/DifferentialEquations.jl/issues/1087.

## Checklist

- [ ] Appropriate tests were added: **This can be tested with the Minimal Reproducible Example in the original issue, but the test can't be implemented in DiffEqBase since it needs to actually solve the ODE**
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated: **Nothing to do**
- [x] The new code follows the
- [x] Any new documentation only uses public API: **Nothing to do**
  
## Additional context

We just ensure that `left < right` even if `prob.tspan` bounds are inverted when integrating backward. I got the change from the [original ITP implementation in NonlinearSolve.jl](https://github.com/SciML/NonlinearSolve.jl/blob/master/lib/BracketingNonlinearSolve/src/itp.jl)

### DIFF
--- a/src/internal_itp.jl
+++ b/src/internal_itp.jl
@@ -30,7 +30,7 @@ function SciMLBase.solve(prob::IntervalNonlinearProblem{IP, Tuple{T, T}}, alg::I
         args...;
         maxiters = 1000, kwargs...) where {IP, T}
     f = Base.Fix2(prob.f, prob.p)
-    left, right = prob.tspan # a and b
+    left, right = minmax(prob.tspan...) # a and b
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Looks right to me. But we should add a test so this doesn't regress.

dcourteville:
I managed to write a simple test case that reproduce the error, it should be good now.

oscardssmith:
thanks for fixing this!

PR #1188, opened 2025-08-08
do not replace LazyBufferCache when promoting f

thomvet:
Addresses https://github.com/SciML/OrdinaryDiffEq.jl/issues/2719
Needs: https://github.com/SciML/PreallocationTools.jl/pull/131 and https://github.com/SciML/SciMLBase.jl/pull/1097

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -1422,7 +1422,7 @@ hasdualpromote(u0, t) = true
 
 function promote_f(f::SplitFunction, ::Val{specialize}, u0, p, t) where {specialize}
     (typeof(f._func_cache) === typeof(u0) || 
-    typeof(f._func_cache) == PreallocationTools.LazyBufferCache) && isinplace(f) ? f :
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Why would an LBF be promoted?

thomvet:
It shouldn't be, but when I am running the example in https://github.com/SciML/OrdinaryDiffEq.jl/issues/2719, promotion is attempted and this was a way of safe-guarding against this. I can look for a more elegant solution.

thomvet:
Yes, with the newest version of https://github.com/SciML/SciMLBase.jl/pull/1097 that should be the case.

We just need the no-op (`promote_f(f::SplitFunction, ::Val{specialize}, u0, p, t) where {specialize} = f`) here to keep the `DiffCache` in place, otherwise it gets converted into a `zero(u0)`, which just re-creates the original problem of not having the right cache.

thomvet:
This is still needed to make https://github.com/SciML/SciMLBase.jl/pull/1106 work (see test failures - without the change here the DiffCache just gets removed when promote_f is called).

PR #1195, opened 2025-08-17
Remove unused Statistics.jl dependency

ChrisRackauckas-Claude:
## Summary
- Removes Statistics.jl as an unused dependency from DiffEqBase.jl
- Reduces the dependency footprint without any functionality loss

## Changes
- Removed `using Statistics` from src/DiffEqBase.jl (line 24)
- Removed Statistics from Project.toml dependencies, compat, extras, and test targets

## Testing
- Package compiles and loads successfully without Statistics
- No actual usage of Statistics functions was found in the codebase

Statistics.jl was imported but never actually used anywhere in DiffEqBase. This is part of a broader effort to reduce dependencies across the SciML stack.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,7 +27,6 @@ SciMLStructures = "53ae85a6-f571-4167-b2af-e1d143709226"
 Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 Static = "aedffcd0-7271-4cad-89d0-dc628f76c6d3"
 StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
-Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
## Fix Applied

Found that `test/downstream/ensemble.jl` was using `Statistics.mean()` and `Statistics.var()` which caused test failures after removing the Statistics dependency.

### Changes Made
- Removed `using Statistics` from test/downstream/ensemble.jl
- Replaced `mean(u)` with `sum(u) / length(u)` 
- Replaced `var(u)` with manual variance calculation: `sum((x - μ)^2 for x in u) / length(u)`

These were the only usages of Statistics in the test suite. The dependency is now fully removed from DiffEqBase.

@ChrisRackauckas This should fix the CI failures. The downstream tests should now pass.

PR #1218, opened 2025-11-09
Fix Enzyme solve_up rule signature to support DuplicatedNoNeed

ChrisRackauckas-Claude:
## Summary

Fixes https://github.com/SciML/SciMLSensitivity.jl/issues/1225

The Enzyme rules for `solve_up` were failing when `sensealg` was passed via the `ODEProblem` constructor instead of `solve()`. The issue was that the return type annotation was restricted to `Duplicated{RT}`, but Enzyme can also use `DuplicatedNoNeed{RT}`.

## Changes

- Updated `augmented_primal` signature to accept both `Duplicated` and `DuplicatedNoNeed`
- Updated `reverse` signature to accept both `Duplicated` and `DuplicatedNoNeed`  
- Fixed type handling to extract the inner result type instead of using the annotation type directly
- Removed incorrect type assertion `dres::RT` in reverse function

## Details

The error was:
```
MethodError: no method matching reverse(::EnzymeCore.EnzymeRules.RevConfigWidth{1, ...}, ::Const{typeof(DiffEqBase.solve_up)}, ::Type{DuplicatedNoNeed{Any}}, ...)
```

The closest candidate was:
```julia
reverse(..., ::Type{Duplicated{RT}}, ...)  # Only accepted Duplicated, not DuplicatedNoNeed
```

The fix makes the signature accept both types:
```julia
reverse(..., ::Type{RT}, ...) where {RT <: Union{Duplicated, DuplicatedNoNeed}}
```

## Test plan

- [ ] Test with sensealg passed to ODEProblem constructor
- [ ] Test with sensealg passed to solve()
- [ ] Verify both Duplicated and DuplicatedNoNeed return types work

### DIFF
--- a/ext/DiffEqBaseEnzymeExt.jl
+++ b/ext/DiffEqBaseEnzymeExt.jl
@@ -9,10 +9,10 @@ module DiffEqBaseEnzymeExt
 
     function Enzyme.EnzymeRules.augmented_primal(
             config::Enzyme.EnzymeRules.RevConfigWidth{1},
-            func::Const{typeof(DiffEqBase.solve_up)}, ::Type{Duplicated{RT}}, prob,
+            func::Const{typeof(DiffEqBase.solve_up)}, ::Type{RT}, prob,
             sensealg::Union{
                 Const{Nothing}, Const{<:DiffEqBase.AbstractSensitivityAlgorithm}},
-            u0, p, args...; kwargs...) where {RT}
+            u0, p, args...; kwargs...) where {RT <: Union{Duplicated, DuplicatedNoNeed}}
         @inline function copy_or_reuse(val, idx)
             if Enzyme.EnzymeRules.overwritten(config)[idx] && ismutable(val)
                 return deepcopy(val)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
## Update

I've corrected the fix. The key changes:

**Previous (incorrect) approach:**
```julia
::Type{RT} where {RT <: Union{Duplicated, DuplicatedNoNeed}}
```
This made `RT` the annotation type itself, causing type mismatches.

**Corrected approach:**
```julia
::Type{<:Union{Duplicated{RT}, DuplicatedNoNeed{RT}}} where {RT}
```
This correctly extracts `RT` as the inner type while accepting both annotation types.

## Current Status

✅ **MethodError fixed** - The function signature now properly matches both `Duplicated` and `DuplicatedNoNeed` annotations

⚠️ **Segfault revealed** - With the MethodError resolved, testing reveals a segfault in Enzyme during AD compilation. This appears to be a deeper Enzyme issue, not a DiffEqBase problem.

The signature fix is correct and necessary - it resolves the immediate MethodError reported in issue #1225. The segfault is a separate Enzyme issue that would need to be reported upstream.

PR #1222, opened 2025-11-10
Correct enzymerules

wsmoses:


wsmoses:
the core1 failure is unrelated. Incidentally was a test clearly not added to enzyme integration tests that merit being added as an integration test [cc @ChrisRackauckas ].

That test ought be fixed in https://github.com/EnzymeAD/Enzyme.jl/pull/2762

PR #1234, opened 2025-12-15
Fix left/right rootfinding for backward propagation

dcourteville:
## Additional context
Fix for SciML/OrdinaryDiffEq.jl#3217


dcourteville:
Oh right I can load BracketingNonlinearSolve and it should work. I will try to do it.

oscardssmith:
I think this now looks good! Thanks for the perseverance on this!

PR #1238, opened 2025-12-28
Fix null solution path to check for callbacks

ChrisRackauckas-Claude:
## Summary

- Add `has_callbacks(kwargs)` helper function to check if callbacks are present in kwargs
- Modify `init_call` and `solve_call` to check for callbacks before taking the null solution/integrator fast path
- When callbacks are present on a null-state problem (`u0 === nothing`), the problem now flows through the normal solve path instead of being short-circuited

This fixes the issue reported in https://github.com/SciML/ModelingToolkit.jl/issues/4078 where `SymbolicDiscreteCallback` events were not being triggered for ModelingToolkit systems with no differential equations (which result in `u0 === nothing` after `mtkcompile`).

## Test plan

- [x] DiffEqBase tests pass
- [x] Manual testing with MWE that includes a dummy variable (callbacks now trigger correctly)

## Notes

The MWE from the issue that uses only `SymbolicDiscreteCallback` without any state variables still errors with `MethodError: no method matching oneunit(::Type{Any})` because OrdinaryDiffEq doesn't currently support solving problems with empty/null state vectors. This PR is a step in the right direction - it prevents the silent callback skipping, but full support for callbacks on truly stateless systems will require additional changes in OrdinaryDiffEq.

### DIFF
--- a/test/downstream/null_de.jl
+++ b/test/downstream/null_de.jl
@@ -102,3 +102,38 @@ end
     @test_nowarn step!(integ)
     @test integ.t ≈ 1.0
 end
+
+# https://github.com/SciML/ModelingToolkit.jl/issues/4078
+@testset "Null problem with callbacks" begin
+    using DiffEqCallbacks
+
+    # Test 1: has_callbacks detection - null problem WITHOUT callbacks takes fast path
+    prob_no_cb = ODEProblem(Returns(nothing), nothing, (0.0, 1.0))
+    sol_no_cb = solve(prob_no_cb, Tsit5())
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Added downstream tests in `test/downstream/null_de.jl`:

1. **Test null problem without callbacks** - verifies the fast path still works
2. **Test null problem WITH callbacks** - verifies we don't take the fast path (throws error proving callbacks are not silently skipped)
3. **Test ODE with state + DiscreteCallback** - verifies callbacks trigger correctly on normal problems
4. **Test init() with null problem + callback** - verifies `init_call` also respects the callback check

PR #1260, opened 2026-01-19
Cleanup event finding and change nudging condition

dcourteville:
While implementing the change in nudging condition for https://github.com/SciML/DiffEqBase.jl/issues/1245, I somehow ended up cleaning up all the event finding code to remove the redundant computations and useless branches, and factorize some code between ContinuousCallback and VectorContinuousCallback. If I didn't break anything it should make maintaining the code easier.
I know I shouldn't make two changes in one PR, I can split into two PR if its an issue.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -305,7 +305,7 @@ end
 end
 
 """
-Return a nudged (if neccessary) value of `integrator.tprev` to avoid repeat event detaction
+Return a nudged (if neccessary) value of `integrator.tprev` to avoid repeat event detection
 - `integrator`
 - `callback`: Last occuring callback
 - `condition_tprev`: Condition of last occuring callback evaluated at `integrator.tprev`
@@ -330,7 +330,7 @@ function check_event_occurence(integrator, callback, bottom_sign)
         check_event_occurence_upto(integrator, callback, bottom_sign, top_t)
 
     if callback.interp_points != 0 && !isdiscrete(integrator.alg) &&
-       sum(event_idx) != length(event_idx)
+        any(iszero, event_idx)
         # Use the interpolants for safety checking
         ts = range(integrator.tprev, stop=integrator.t, length=callback.interp_points)
         for i in 2:length(ts)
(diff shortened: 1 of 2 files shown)


dcourteville:
I'll update the doc for `abstol` in SciMLBase

ChrisRackauckas:
```
Null problem with callbacks: Test Failed at /home/runner/work/DiffEqBase.jl/DiffEqBase.jl/test/downstream/null_de.jl:165
  Expression: solve(prob_with_cb, Tsit5(); callback = cb)
    Expected: Exception
  No exception thrown

Stacktrace:
 [1] top-level scope
   @ ~/work/DiffEqBase.jl/DiffEqBase.jl/test/downstream/null_de.jl:151
 [2] macro expansion
   @ /opt/hostedtoolcache/julia/1.12.4/x64/share/julia/stdlib/v1.12/Test/src/Test.jl:1776 [inlined]
 [3] macro expansion
   @ ~/work/DiffEqBase.jl/DiffEqBase.jl/test/downstream/null_de.jl:165 [inlined]
Null problem with callbacks: Test Failed at /home/runner/work/DiffEqBase.jl/DiffEqBase.jl/test/downstream/null_de.jl:181
  Expression: init(prob_with_cb, Tsit5(); callback = cb)
    Expected: Exception
  No exception thrown
```

update those, downstream now handles that so it just needs to be set to passing. we will need to see downstream tests because they have a lot of callback stuff

ChrisRackauckas:
This seemed to have some uncaught downstream consequences:

* https://github.com/SciML/DiffEqBase.jl/issues/1268
* https://github.com/SciML/DiffEqBase.jl/issues/1267

I can handle the GPU things, that's from using non-public API in DiffEqGPU. But the MWE @AayushSabharwal found seems more central as a repeat callback detection failure. I think because of how much stuff looks breaking right now, it might be good to revert, figure out the repeat callback, and then re-merge, and I'll get the GPU update setup and ready for that. @dcourteville can you look into that repeat callback issue?

PR #1262, opened 2026-02-02
Remove unused EnzymeCore strong dependency

ChrisRackauckas-Claude:
## Summary
- Remove `EnzymeCore` from `[deps]` — it was never imported or used in any source file
- Remove the now-dangling `[compat]` entry for `EnzymeCore`
- Bump version to 6.200.0

## Motivation
EnzymeCore as a strong dependency forces it to be installed for all DiffEqBase users. This causes OrdinaryDiffEqCore's `EnzymeCoreExt` extension to trigger unconditionally. Combined with the always-available `SparseArrays` stdlib triggering `SparseArraysExt`, Julia 1.10's precompilation system detects a circular dependency between the two extensions, preventing precompilation and causing errors when loading with EnzymeCore.

See SciML/OrdinaryDiffEq.jl#3025

Fixes SciML/OrdinaryDiffEq.jl#3025

## Test plan
- [x] `Pkg.test()` passes locally
- [x] `using DiffEqBase` loads without errors

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,14 +1,13 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.199.0"
+version = "6.200.0"
 
 [deps]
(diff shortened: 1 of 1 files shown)


oscardssmith:
Do we know why this was added in the first place?

ChrisRackauckas:
rules

PR #1287, opened 2026-02-22
Thread ForwardDiff chunk size through FunctionWrapper compilation

ChrisRackauckas-Claude:
## Summary
- Add `_forwarddiff_chunksize` helper to read the new `SciMLBase.forwarddiff_chunksize` trait
- Thread `Val(CS)` through `get_concrete_problem` → `promote_f` → `wrapfun_iip` for both ODE and DAE problems
- Add parameterized `dualgen(T, Val(CS))` in the ForwardDiff extension
- Add 3-arg `wrapfun_iip(ff, inputs, Val(CS))` that compiles FunctionWrapper variants with chunk=CS for u-related duals (Jacobian) and chunk=1 for t-related duals (time derivative)
- Add 3-arg fallback `wrapfun_iip(ff, inputs, ::Val)` in norecompile.jl for when ForwardDiff extension isn't loaded

## Context
DiffEqBase PR #1284 enabled FunctionWrapper wrapping for mass matrix ODEProblems. The FunctionWrapper variants were compiled with `Dual{..., 1}` (chunk=1 hardcoded in `dualgen`). When algorithms use a different chunk size (e.g. `AutoForwardDiff(chunksize=3)`), the solver creates `Dual{..., 3}` values but the wrapper only has `Dual{..., 1}` variants, causing `NoFunctionWrapperFoundError`.

The key insight is that u-related duals (Jacobian computation) use the algorithm's chunk size, while t-related duals (time derivative) always use chunk=1 since it's scalar differentiation. The 3-arg `wrapfun_iip` handles this by creating variants with different chunk sizes for the u-derivative vs t-derivative paths.

Depends on: SciML/SciMLBase.jl#1244

## Test plan
- [x] Precompilation succeeds with the changes
- [x] `Val(1)` default preserves existing behavior when trait returns 0
- [x] Tested end-to-end with OrdinaryDiffEqRosenbrock DAE AD tests (8/8 pass)
- [x] Rosenbrock convergence tests still pass (853/853)
- [ ] CI

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -676,7 +676,7 @@ function get_concrete_problem(prob, isadapt; alg = nothing, kwargs...)
     f_promote = promote_f(
         prob.f, Val(SciMLBase.specialization(prob.f)), u0_promote, p,
         tspan_promote[1], Val(_uses_forwarddiff(alg)),
-        Val(_forwarddiff_chunksize(alg))
+        _forwarddiff_chunksize(alg)
     )
     if isconcreteu0(prob, tspan[1], kwargs) && prob.u0 === u0 &&
             typeof(u0_promote) === typeof(prob.u0) &&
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
locally handled.

